PR #12, opened 2018-02-22
Refactor Search widget and introduce SearchableViewMixin

### LINKED ISSUE #5: Refactor Search widget
<a href="https://github.com/abalakh"></img></a> **Issue by [abalakh](https://github.com/abalakh)**
_Monday Feb 19, 2018 at 12:34 GMT_
_Originally opened as https://github.com/abalakh/airgun/issues/23_

----

Current implementation doesn't respect layers isolation - `read()` returns not value of searchbar, but value of results table, which is outside of widget and should be handled on view level. Such implementation will also cause issues when automating some uncommon entities where multiple fields from results table should be taken into consideration.
Ideally widget's `fill` and `read` should only work with searchbar, and view should contain separate `search()` method which will return results from table. 
We can also think of some `SearchableMixin` for views not to have to implement search for every view, since most of views will have very similar implementation and only locator will differ in most cases.

abalakh:
Fixes #5

### DIFF
--- a/airgun/views/common.py
+++ b/airgun/views/common.py
@@ -1,7 +1,7 @@
-from selenium.common.exceptions import NoSuchElementException
-from widgetastic.widget import View
+import six
+from widgetastic.widget import NoSuchElementException, View, WidgetMetaclass
 
-from airgun.widgets import HorizontalNavigation, ContextSelector, Search
+from airgun.widgets import ContextSelector, HorizontalNavigation, Search
 
 
 class BaseLoggedInView(View):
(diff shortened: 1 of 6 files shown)


airgun/widgets.py
```diff
@@ -73,12 +73,10 @@ class Search(Widget):
     def fill(self, value):
         return self.search_field.fill(value)
 
-    def read(self, value, result_locator=None):
-        if result_locator is None:
-            result_locator = self.default_result_locator
```
oshtaier: remove ` default_result_locator = Text("//a[contains(., '%s')]")` too

airgun/views/common.py
```diff
@@ -1,8 +1,48 @@
...
+    """Mixin which adds :class:`airgun.widgets.Search` widget and
+    :meth:`search` to your view. It's useful for _most_ entities list views
+    where searchbox is present.
+
+    Note that you can override expected result locator for the element which is
+    returned by :meth:`search` by specifying custom ``search_result_locator``
+    string variable in your view class.
+    """
+    searchbox = Search()
+    search_result_locator = "//a[contains(., '%s')]"
```
oshtaier: It is overridable, right?

abalakh: Take a closer look at docstring ;)
Yes, it's overridable, that's one of main points of such mixin.

oshtaier: right

airgun/widgets.py
abalakh: addressed, thanks.

PR #33, opened 2018-03-02
Add documentation spellcheck into CI process

jhutar:
What about enabling spellcheck on documentation as well?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 language: python
 python:
   - "3.6"
+before_install:
+  - "sudo apt-get install python-enchant"
 install:
   - "pip install -r requirements.txt"
   - "pip install -r requirements-optional.txt"
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,3 +7,4 @@ install:
 script:
   - "flake8 ."
   - "make docs-html"
+  - "make docs-spelling"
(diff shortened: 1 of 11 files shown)


docs/spelling_wordlist.txt
```diff
@@ -0,0 +1,16 @@
+AirGun
+Robottelo
+saucelabs
+jQuery
+Patternfly
+prequisite
```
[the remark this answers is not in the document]
jhutar: Yep, I know, but it comes from widgetastic. I have created PR for that, but I do not know when that fix gets released, so added this as a workaround.

jhutar:
Yes, makes sense. Lets not have this in the Travis

.travis.yml
```diff
@@ -1,6 +1,8 @@
 language: python
 python:
   - "3.6"
+before_install:
+  - "sudo apt-get install python-enchant"
```
abalakh: is this still required even if we don't run spellcheck on travis?

jhutar: Yes, because we have `sphinxcontrib-spelling` in `requirements-optional.txt` and that requires this lib. I agree it is not optimal. Another way I see here is to alter Makefile so it asks for that that dependency if you attmpt to run the spell check.

abalakh: Ok, not a blocker, let's decide later.

PR #34, opened 2018-03-02
Add new widgets for parameters and fix edit functionality

oshtaier:


airgun/views/os.py
```diff
@@ -35,6 +37,19 @@ def read(self):
             self.browser.click(self.view_tab)
             return self.ptables.read()
 
+    @View.nested
+    class parameters(View):
```
abalakh: Please take a look at widgetastic_patternfly `Tab()` view, we should probably use it or make our equivalent instead. Or submit an issue if you think it's out of scope :)

airgun/widgets.py
```diff
@@ -212,3 +212,53 @@ def fill(self, value):
...
+                <input placeholder="Name" type="text" ... id="..._name">
+                <textarea id="new_os_parameter_value" placeholder="Value" ...>
+
+    Locator example::
+
+        //input[@placeholder='Name']
+        //textarea[@placeholder='Value']
+
+    """
+    add_new_value = Text("//a[contains(text(),'+ Add Parameter')]")
```
abalakh: Can we use relative paths, aren't all the objects located at least inside the same view?

airgun/widgets.py
```diff
@@ -212,3 +212,53 @@ def fill(self, value):
...
+        locator="//textarea[@placeholder='Value' and not(text())]")
+    NAMES = "//tr[contains(@id, 'os_parameter')]/td/input[@placeholder='Name']"
+    VALUE = (
+        "//table[contains(@id, 'parameters')]//tr"
+        "/td[input[contains(@id, 'name')][contains(@value, '{}')]]"
+        "/following-sibling::td//textarea"
+    )
+
+    def read(self):
+        """Return a list of dictionaries. Each dictionary contains from name
```
abalakh: `contains from name`? Please rephrase. Maybe you meant `consists of`?

airgun/views/os.py
oshtaier: https://github.com/SatelliteQE/airgun/issues/35

PR #39, opened 2018-03-06
add few widgets + expand ak functionality

oshtaier:


airgun/widgets.py
```diff
@@ -267,3 +274,136 @@ def fill(self, values):
...
+            <div class="modal-footer ng-scope">
+                <button class="btn btn-danger" ng-click="ok()">
+                <button class="btn ..." ng-click="cancel()"...>
+
+    Locator example::
+
+        //div[@class='modal-content']
+
+    """
+    ROOT = "//div[@class='modal-content']"
```
abalakh: ".//div[...]"

abalakh: Btw, why `ConfirmationDialog` is a widget, it sounds like a view to me?

airgun/widgets.py
```diff
@@ -267,3 +274,136 @@ def fill(self, values):
...
+            <li class="path-list-item ng-scope"...>
+                <label class="path-list-item-label...>
+                    <input type="checkbox"...>
+
+    Locator example::
+
+        //ul[@class='path-list']
+
+    """
+    ROOT = "//ul[@class='path-list']"
```
abalakh: ".//ul[...]"

airgun/widgets.py
```diff
@@ -267,3 +274,136 @@ def fill(self, values):
...
+        """Identify whether specific checkbox is selected or not"""
+        return 'ng-not-empty' in self.browser.get_attribute('class', locator)
+
+    def select(self, locator, value):
+        """Select or deselect checkbox depends on the value passed"""
+        value = bool(value)
+        current_value = self.cb_selected(locator)
+        if value == current_value:
+            return False
+        else:
```
abalakh: Pls drop this else, it's not required after `return` in previous condition

airgun/widgets.py
```diff
@@ -267,3 +274,136 @@ def fill(self, values):
...
+        //ul[@class='path-list']
+
+    """
+    ROOT = "//ul[@class='path-list']"
+    LABELS = "./li/label[contains(@class, path-list-item-label)]"
+    CHECKBOX = (
+        ".//input[@ng-model='item.selected'][parent::label[contains(., '{}')]]"
+    )
+
+    def cb_selected(self, locator):
```
abalakh: s/cb_selected/checkbox_selected/

airgun/widgets.py
```diff
@@ -267,3 +274,136 @@ def fill(self, values):
...
+        return checkboxes
+
+    def fill(self, values):
+        """Assign value for specific checkbox from group
+
+        :param values: dictionary that consist of checkbox name and value
+            that should be assigned to that checkbox
+        """
+        checkbox_locator = self.CHECKBOX.format(values.get('entity_name'))
+        return self.select(checkbox_locator, values.get('select'))
```
abalakh: Helpers like `select` and `cb_selected` accept locators and create web elements multiple time. Are you sure that's the optimal way? I'd rather form web element for checkbox here and passed that single element to all the helpers.

airgun/widgets.py
```diff
@@ -267,3 +274,136 @@ def fill(self, values):
...
+            checkboxes.append({'name': name, 'value': value})
+        return checkboxes
+
+    def fill(self, values):
+        """Assign value for specific checkbox from group
+
+        :param values: dictionary that consist of checkbox name and value
+            that should be assigned to that checkbox
+        """
+        checkbox_locator = self.CHECKBOX.format(values.get('entity_name'))
```
abalakh: as per discussion, let's drop "entity_name" and "value" keys and just pass name as a key, e.g. 'foo': True

airgun/views/activationkey.py
```diff
@@ -14,13 +19,28 @@ def is_displayed(self):
...
     @property
     def is_displayed(self):
         return self.browser.wait_for_element(
             self.name, exception=False) is not None
+
+
+class ActivationKeyEditView(BaseLoggedInView):
+    return_to_all = Text("//a[text()='Activation Keys']")
+    action_list = SelectActionList()
+    dialog = ConfirmationDialog()
```
abalakh: Have you checked cfme code? Separate dialog window as a widget doesn't sound right to me, i'd rather expect it to be a view, which appears during this view's `fill`. Does it make sense?

airgun/widgets.py
oshtaier: You can't do that as `fill()` and `read()` has no connection with each other

airgun/widgets.py
oshtaier: as per discussion - we need to use:
`mydictionary.keys()[0]` and `mydictionary[mydictionary.keys()[0]]`
that is not the good idea.
Any alternatives without loop?

airgun/views/activationkey.py
oshtaier: Yes, I am sure. It is simple control with two buttons

Issue #21, opened 2018-02-22
Update README.rst


abalakh:
README.rst should include proper description, entities/views/widgets/session explanation, install guide, configuring section (both by `configure()` call and `settings.ini`) and some example of usage.

jhutar:
I consider this done by https://github.com/SatelliteQE/airgun/pull/45

Issue #23, opened 2018-02-26
Create "landing page" for new contributors


omaciel:
As a potential new contributor (upstream or downstream), I want to know how to get started contributing with this project and what available tasks are ready for me

jhutar:
I consider this done by https://github.com/SatelliteQE/airgun/pull/45

PR #45, opened 2018-03-15
Improving documentation - README, contributing, explaining views and entities

jhutar:
Attempt to address #21 and #23.


jhutar:
Split to separate pages? I like it this way, but do not have strong preference. If it is OK, I will merge "how to build docs locally" into index as well.

docs/build-docs.rst
```diff
@@ -7,3 +7,10 @@ To build documentation locally, use:
 
     pip install -r requirements-optional.txt
     make docs-html
+
+You might also try to run spell check:
+
+.. code-block:: bash
+
+    pip install sphinxcontrib-spelling
```
abalakh: it's already present in `requirements-optional.txt`

docs/index.rst
```diff
@@ -46,5 +72,42 @@ reuse already existing properties file since no changes were introduced there)
...
+    source venv/bin/activate
+
+If you are about to start contributing, please follow `GitHub flow`_.
+
+For a new-comer, we have issues tagged with "`good first issue`_" so that
+might be place where you can start?
+
+When testing your AirGun code using Robottelo tests, I have used following
+ugly trick to have mine current development version of AirGun available in
+Robottelo virtual environment:
```
abalakh: i'd rather use PYTHONPATH than such symlink - `export PYTHONPATH="$PYTHONPATH:/path/to/airgun/"` and would probably specify that in .bashrc/.zshrc
Shouldn't be an issue when we publish airgun to pypi.
And please don't use words like `ugly trick` in official documentation 🙂

docs/index.rst
```diff
@@ -46,5 +72,42 @@ reuse already existing properties file since no changes were introduced there)
...
+
+.. code-block:: bash
+
+    mv venv/lib/python3.6/site-packages/airgun{,ORIG}
+    ln -s ../airgun/airgun/ venv/lib/python3.6/site-packages/airgun   # depends where you have airgun checked out
+
+When you are running your tests locally, you (currently) need Chrome browser
+installed and `chromedriver`_ binary available somewhere in your PATH (in
+your robottelo checkout): download it, unzip it and put e.g. into (if you
+are using virtual environments) `venv/bin/`.
```
abalakh: Not necessary, there's `webdriver_binary=` option in robottelo.properties where you can specify path to chromedriver. Only if it's not specified then robottelo will look for it in PATH.

docs/index.rst
```diff
@@ -4,6 +4,10 @@ AirGun
 AirGun is a Python library which is build over `Widgetastic`_ and `navmazing`_
 to make Satellite 6 UI testing easier.
 
+AirGun is written in Python 3 and do not plan to support Python 2. AirGun
+supports Satellite 6.4 and above and Foreman 1.17 and above (with vertical
+navigation).
```
abalakh: Btw it's currently possible to use py2 - backward compatibility is preserved atm, but yeah, it can be dropped any time, so i'm not sure if that's worth specifying, just fyi.

oshtaier: I am also against specifying any static information like python or application versions. Usually, documentation is not updated in next years and such details become outdated

docs/index.rst
```diff
@@ -17,6 +21,32 @@ More in-depth coverage is provided in other sections.
...
+**Entity** is a functional side of application (Satellite or Foreman in our
+case) for some object in it. It defines actions you can do with the object.
+This is the API that end tests uses.
+
+AirGun adds bunch of **widgets** (which are specific to Satellite / Foreman
+web UI) to these defined by Widgetastic (like generic `Text`). Widgets allows
+you to interact with various UI elements and are used in views.
+
+And finally tests. Tests are not part of AirGun repository. You can
+contribute tests to the Robottelo (currently only in abalakh's fork, `branch
```
oshtaier: We need to not forget update that information when we move code to proper repo. Do we have anything like #todo tag?

docs/index.rst
```diff
@@ -17,6 +21,32 @@ More in-depth coverage is provided in other sections.
...
 
+
+Concepts
+--------
+
+Besides code which implements Widgetastic and navmazing functionality, two
+basic concepts used to abstract Satellite/Foreman UI functionality are Views
+and Entities.
+
+**View** is part of page containing widgets, it describes appearance of the
```
oshtaier: +**View** is part of page containing widgets, it describes appearance of the
+page. Here all the nasty XPaths goes, but please create Foreman or Katello
+pull requests for `id=...` additions where reasonable, as generally XPaths
+are too fragile.

**View** is fundamental basic concept that describes appearance of the page. Here we store all information about controls (widgets) on the page, specifically locators to them or when control is basic one that has common locator we just define widget without any further details. Of course, we try to address locators using id or name as first priority, but if we can't - we go with Xpath. If you see situation like this - create Foreman or Katello pull request to assign unique id attribute for necessary element where it is reasonable, as Xpaths are more fragile and has worse performance to interact with

docs/index.rst
```diff
@@ -17,6 +21,32 @@ More in-depth coverage is provided in other sections.
...
+page. Here all the nasty XPaths goes, but please create Foreman or Katello
+pull requests for `id=...` additions where reasonable, as generally XPaths
+are too fragile.
+
+**Entity** is a functional side of application (Satellite or Foreman in our
```
oshtaier: **Entity** is fundamental basic concept that is responsible for functional side of application (Satellite in our case). It defines actions you can do with objects like CRUD (create, update, delete) for example. That is only API that is visible for end user on test side. Also, we put all details about navigation here, so framework know how to get to necessary page when you need to create specific object in the application

docs/index.rst
abalakh: Oh, and maybe it's worth mentioning explicitly, that the only supported webdriver for now is chromedriver.

docs/index.rst
jhutar: I wanted to use this to somehow set expectations - just to express "do not waste your time with python 2" and "do not attempt to run on 6.3". But I agree this tends to become outdated. Any idea?

Ad proper repo: IMO we can fix first moment when somebody complains about it.

jhutar:
Thank you very much for comments. Will incorporate these.

docs/index.rst
jhutar: IMO this Python version have same validity as info that only chromedriver is supported - it helps whoever is working on the project to set expectations. Same for Sat/Foreman versions: it only says minimal version which were set by management so we should let everybody know not to spend time trying to support e.g. Sat 6.3. But I have absolutely no problem with removing it from docs. Let me know.

jhutar:
If this gets accepted, I plan to close #21 and #23 with it.

PR #46, opened 2018-03-20
Add radio group widget

oshtaier:


airgun/entities/subnet.py
```diff
@@ -36,3 +40,15 @@ def step(self, *args, **kwargs):
         self.view.browser.wait_for_element(
             self.parent.new, ensure_page_safe=True)
         self.parent.browser.click(self.parent.new)
+
+
+@navigator.register(SubnetEntity, 'Edit')
+class EditExistingSubnet(NavigateStep):
```
abalakh: Why `EditExistingSubnet` and not just `EditSubnet`?
I can see some other entities are following the same naming pattern, so probably not a blocker since we'll need some extra effort to fix it elsewhere.

airgun/widgets.py
```diff
@@ -12,6 +14,88 @@
 from widgetastic_patternfly import VerticalNavigation
 
 
+class RadioGroup(Widget):
```
abalakh: Why not inheriting from `GenericLocatorWidget`? That way you wouldn't have to specify `def __init__()`

airgun/widgets.py
```diff
@@ -12,6 +14,88 @@
...
+        """Return all radio group labels"""
+        return [
+            self.browser.text(btn)
+            for btn
+            in self.browser.elements(self.LABELS)
+        ]
+
+    def _get_parent_label(self, name):
+        """Get radio group label for specific button"""
+        br = self.browser
```
abalakh: This shortcut is not needed, you can fit your iterator in 2 lines:
```python
    def _get_parent_label(self, name):
        """Get radio group label for specific button"""
        try:
            return next(
                btn for btn in self.browser.elements(self.LABELS)
                if self.browser.text(btn) == name
            )
        except StopIteration:
            raise NoSuchElementException(
                "RadioButton {name} is absent on page".format(name=name))
```

airgun/widgets.py
```diff
@@ -12,6 +14,88 @@
...
+                btn for btn in br.elements(self.LABELS) if br.text(btn) == name
+            )
+        except StopIteration:
+            raise NoSuchElementException(
+                "RadioButton {name} is absent on page".format(name=name))
+
+    @property
+    def selected(self):
+        """Return name of a button that is currently selected in the group"""
+        names = self.button_names
```
abalakh: redundant var?

PR #55, opened 2018-03-29
Make code more generic + add support for tab with dropdown

oshtaier:


airgun/entities/contentview.py
```diff
@@ -33,6 +33,10 @@ def update(self, entity_name, values):
         view = self.navigate_to(self, 'Edit', entity_name=entity_name)
         return view.fill(values)
 
+    def add_yum_repo(self, entity_name, repo_name):
+        view = self.navigate_to(self, 'Edit', entity_name=entity_name)
```
abalakh: Hm... Shouldn't we have navigate step for navigating 'Yum content' tab? Won't we have difficulties with understanding where we are when we introduce corresponding logic later?

airgun/views/common.py
```diff
@@ -84,12 +99,13 @@ class resources(AddRemoveResourcesView):
...
-        no_rows = Text(
-            ".//table//span[@data-block='no-rows-message' or "
-            "@data-block='no-search-results-message']"
+        remove_button = Text(
+            './/div[@data-block="list-actions"]'
+            '//button[contains(@ng-click, "remove")]'
         )
+        table = Table(locator=".//table")
+        columns_exists = Text(
+            ".//table//*[contains(name(), 'body')]/tr[1]/td[2]")
```
abalakh: Hm... When we previously discussed this, you've rejected proposal to use indexes and described it's a bad idea here. Why changed your mind?

airgun/views/contentview.py
```diff
@@ -60,3 +62,12 @@ class Details(SatTab):
...
         composite = ReadOnlyEntry(name='Composite?')
         force_puppet = EditableEntryCheckbox(name='Force Puppet')
+
+    @View.nested
+    class yumrepo(SatTabWithDropdown):
+        TAB_NAME = 'Yum Content'
+        SUB_ITEM = 'Repositories'
+
+        @View.nested
+        class repos(AddRemoveResourcesView):
```
abalakh: tip: turns out there's nicer way to do add nested view when no modifications needed:
```python
repos = View.nested(AddRemoveResourcesView)
```

airgun/views/common.py
omaciel: Unless I'm missing something, this is screaming "give me an ID I can use"...

oshtaier: I changed my mind, because previous approach stopped to work in other places, this one is only one that will ever work

And no, we do not need id here as we use that xpath to calculate amount of columns in the table, not to address the element

airgun/views/contentview.py
oshtaier: thanks, will definitely try

airgun/views/common.py
abalakh: If table is empty - it contains span with `@data-block='no-rows-message'` or `@data-block='no-search-results-message'`, that's basically good enough "id" for us, no updates needed. And if table is not empty - we probably don't need such id (where to place it? first cell of the table will contain id like 'im_not_empty_table'? doesn't seem right). My question was rather - if table is not empty, we could check that span with `data-block='no-search-results-message'` is absent, instead of trying to access specific cell to see whether there's content.

airgun/entities/contentview.py
oshtaier: can you provide more details to that question? I am using the same idea that we used for associate product in activation key. Do we have changes in approach?

airgun/views/common.py
abalakh: >because previous approach stopped to work in other places, this one is only one that will ever work

@oshtaier i think that's what @omaciel was pointing out - if there're places where no ids are present for "empty" message, maybe we should update them?

airgun/entities/contentview.py
abalakh: Yes, i know we haven't thought about that before, it's not a blocker for current PR. I'm just wondering whether navigator should be aware of all the tabs. I can see cfme does have tabs registered as nav steps, at least some of them, which is probably allowing to figure out where user currently is and whether navigation is needed. On the other hand, all the tabs have a logic like "if user is not here - click the tab first", so maybe that's redundant. @mfalesni any details?

airgun/views/common.py
oshtaier: problem is completely in other area. Element is exist, but not visible, so `is_displayed` will not work in that case. I would say that is a bug for webUI from UX perspective, but it is so minor that it will take years before it will be fixed

PR #61, opened 2018-04-05
add datetime collection widget and syncplan entity

oshtaier:


airgun/views/syncplan.py
```diff
@@ -0,0 +1,70 @@
...
+    submit = Text("//button[contains(@ng-click, 'handleSave')]")
+
+    @property
+    def is_displayed(self):
+        return self.browser.wait_for_element(
+            self.name, exception=False) is not None
+
+
+class SyncPlanEditView(BaseLoggedInView):
+    return_to_all = Text("//a[text()='Sync Plans']")
```
abalakh: Please leave some kind of `fixme` or `todo` comment - that's not separate link, that's patternfly's `Breadcrumb` widget which we should implement and remove this `Text` widget afterwards

airgun/widgets.py
```diff
@@ -90,6 +90,57 @@ def fill(self, name):
         return self.select(name)
 
 
+class DateTime(Widget):
+    """Classical collection of paternfly data picker and two inputs for
```
abalakh: /s/data/date
And probably remove 'classical' and 'paternfly' words - i've checked patternfly, their widget looks differently, it's probably satellite's own implementation. Or am i missing something?

airgun/widgets.py
```diff
@@ -90,6 +90,57 @@ def fill(self, name):
...
+                        <table...>
+        ...
+
+    Locator example::
+
+        We don't need to pass locator here as widget seems has one structure
+        across all applications that use paternfly pattern
+    """
+    start_date = TextInput(id='startDate')
+    hours = TextInput(locator="//input[@ng-model='hours']")
```
abalakh: relative xpath if possible, pls

airgun/views/syncplan.py
```diff
@@ -0,0 +1,70 @@
...
+    action_list = SelectActionList()
+    dialog = ConfirmationDialog()
+
+    @property
+    def is_displayed(self):
+        return self.browser.wait_for_element(
+            self.return_to_all, exception=False) is not None
+
+    @View.nested
+    class Details(SatTab):
```
abalakh: Oh, and please use lowercase for tab names, so you won't have to bother with case in your tests (like `syncplan_values['Details']['interval']`). Tab names are capitalized automatically so lowercase class name will still work without need to specify `TAB_NAME` var.

PR #64, opened 2018-04-11
Add LCE entity

oshtaier:
Added lce entity

### DIFF
--- a/airgun/entities/lifecycleenvironment.py
+++ b/airgun/entities/lifecycleenvironment.py
@@ -1,0 +1,86 @@
+from navmazing import NavigateToSibling
+
+from airgun.entities.base import BaseEntity
+from airgun.navigation import NavigateStep, navigator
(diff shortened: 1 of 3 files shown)


airgun/entities/lifecycleenvironment.py
```diff
@@ -0,0 +1,85 @@
...
+        view.fill(values)
+        view.submit.click()
+
+    def create_environment(self, entity_name, values):
+        view = self.navigate_to(
+            self, 'New Environment', entity_name=entity_name)
+        view.fill(values)
+        view.submit.click()
+
+    def create(self, prior_entity_name=None, values=None):
```
abalakh: Should be `def create(self, values, prior_entity_name=None)` - values are required and should be specified in any case. Also such order will allow you skipping explicit specification of `.create(value={...})` when `prior_entity_name` is not needed.

PR #65, opened 2018-04-12
Added satellite's flash messages; updated search widget; added ak_add_hc helper

abalakh:


airgun/widgets.py
```diff
@@ -267,6 +272,63 @@ class SatVerticalNavigation(VerticalNavigation):
...
+                result.append(
+                    SatFlashMessage(self, flash_div, logger=self.logger))
+        except NoSuchElementException:
+            pass
+        return result
+
+
+class SatFlashMessage(FlashMessage):
+    """Satellite version of Patternfly alert. It doesn't contain ``<strong>``
+    tag and all the text is inside 2 ``<span>``.
```
abalakh: Submitted [BZ1566565](https://bugzilla.redhat.com/show_bug.cgi?id=1566565)

airgun/widgets.py
```diff
@@ -267,6 +272,63 @@ class SatVerticalNavigation(VerticalNavigation):
...
+        </div>
+
+    Locator example::
+
+        //div[@class="toast-notifications-list-pf"]
+
+    """
+
+    @property
+    def messages(self):
```
oshtaier: Okay, everything is inherited from `Widget` which doesn't have this nice `all` method which is good to use here
I am thinking about that checking visibility should be solely purpose of `satflashmessage` class.
I would more like to see something similar to `if `SatFlashMessage`.exists?=>append`. Is it possible to make more optimal that algorithm?

abalakh: I haven't touched original algorithm, it's completely the same as in `FlashMessages`, just returns our `SatFlashMessage` instead of `FlashMessage`. Imo, ideally such improvements should be considered in `widgetsastic_patternfly` repo, satellite's implementation should be dropped as soon as BZ1566565 is resolved.

PR #67, opened 2018-04-17
Add template entity and necessary widgets

oshtaier:


airgun/views/template.py
```diff
@@ -0,0 +1,60 @@
...
+            id='provisioning_template_template_kind')
+
+    @View.nested
+    class association(SatTab):
+        applicable_os = MultiSelect(
+            id='ms-provisioning_template_operatingsystem_ids')
+
+    @View.nested
+    class locations(SatTab):
+        resources = MultiSelect(
```
abalakh: Why `resources` for loc/org but `applicable_os` for association? I think we used `resources` only for `AddRemoveResourcesView`

airgun/widgets.py
```diff
@@ -473,6 +516,58 @@ def fill(self, values):
         self.new_parameter_value.fill(values['value'])
 
 
+class TableActionList(Widget):
```
abalakh: What's the difference with `SelectActionList` and widgetastic_patternfly's `Dropdown`? Seems like we're starting to introduce very similar separate widgets for the same purpose

airgun/views/template.py
oshtaier: because `applicable_os ` is a name of table and `resources`, because don't want to have `location.location`, so if you don't mind we will avoid serious discussion about that =)

airgun/widgets.py
oshtaier: 1) You probably meant `TableActionList`
2) My widget operate with separate button + dropdown, their with dropdown
3) I substituted 150 lines of redundant code with
```
self.open_dropdown.click()
self.browser.click(
    self.browser.element(self.dropdown_action.format(value)))
``` 
In future we can use my approach as I don't want do 100 actions and waste time on unnecessary waiters, closing dropdowns and etc

PR #68, opened 2018-04-18
Convert BaseEntity and Location

mshriver:
Work in progress!

Suggestions for entity attribute definition, view classes, and navigation via the entity object linked by the navigation class.

Attr use in classes to define attributes instead of passing during navigation
Properties for things that depend on browser in BaseEntity

Use navmazing.NavigateStep obj property to navigate, don't pass context
implement view with is_displayed instead of am_i_here

This will require test changes I haven't worked out yet.

### DIFF
--- a/airgun/entities/base.py
+++ b/airgun/entities/base.py
@@ -1,8 +1,14 @@
+import attr
 
 
+@attr.s
 class BaseEntity(object):
+    browser = attr.ib()
 
-    def __init__(self, browser):
-        self.browser = browser
-        self.session = browser.extra_objects['session']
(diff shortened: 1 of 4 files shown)


mshriver:
Travis build failed on flake, removed newline from end of `entities/login.py`

mshriver:
Given how these entity instances are cached by the session, this type of entity attribute implementation would not work. It would require implementation of true entity collections for the session, and instantiation of specific entity objects for things in satellite.

Served its purpose as a talking point and soapbox for object-oriented over functional designs :)

PR #69, opened 2018-04-18
Added AK Content Hosts tab and remove_host_collection helper

abalakh:
See SatelliteQE/robottelo#5933 for test results

### DIFF
--- a/airgun/widgets.py
+++ b/airgun/widgets.py
@@ -719,7 +719,11 @@ class EditableEntry(GenericLocatorWidget):
         :param value: string with value that should be used for field update
             procedure
         """
-        self.edit_button.click()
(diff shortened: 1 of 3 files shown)


airgun/widgets.py
```diff
@@ -719,7 +719,11 @@ def fill(self, value):
         :param value: string with value that should be used for field update
             procedure
         """
-        self.edit_button.click()
+        # in some cases editing fields automatically triggers editing others,
+        # so the field may be opened for editing and clicking "edit" button is
+        # not required for it
+        if self.edit_button.is_displayed:
```
oshtaier: not 100% happy with that approach as `edit_button is not displayed` doesn't equal `edit_field is displayed`, but let's resolve issue if that ever happen

Issue #24, opened 2018-02-26
Add compute profile entity


oshtaier:

tstrych:
I am working on it.

jhutar:
@tstrych please push what you have so far and ask for a review. It is better not to wait for a final version in case you there are some errors or style issues which can be fixed now so you are not building on these. Thank you!

abalakh:
Closed via #71

PR #71, opened 2018-04-19
First attend to implement computeprofile view and entity

tstrych:
I start with this, I will continue with delete option. 

Please, can you give me the feedback on computeprofile view and entity ?
Thanks

Test examples:

```
from nailgun import entities

from robottelo.datafactory import gen_string


def test_positive_create(session):
    name = gen_string('alpha')
    with session:
        session.computeprofile.create({
            'name': name,
            })
        assert session.computeprofile.search(name) == name


def test_positive_rename(session):
    name = gen_string('alpha')
    entities.ComputeProfile(name=name).create()
    new_name = gen_string('alpha')
    with session:
        session.computeprofile.rename(name, {'name': new_name, })
        assert session.computeprofile.search(new_name) == new_name
```


airgun/views/computeprofile.py
```diff
@@ -0,0 +1,33 @@
...
+    name = TextInput(locator=".//input[@id='compute_profile_name']")
+    submit = Text('//input[@name="commit"]')
+
+    @property
+    def is_displayed(self):
+        return self.browser.wait_for_element(
+            self.name, exception=False) is not None
+
+
+class ComputeProfileEditView(BaseLoggedInView):
```
tstrych: I'm sorry, I forgot to delete this class. Right now it's unreachable code. It will be fixed with delete option.

airgun/session.py
```diff
@@ -235,3 +236,8 @@ def organization(self):
     def subnet(self):
         """Instance of Subnet entity."""
         return SubnetEntity(self.browser)
+
+    @cached_property
```
tstrych: same mistake as schlupov, with next commit i will change it.

airgun/entities/computeprofile.py
```diff
@@ -0,0 +1,63 @@
...
+    def create(self, values):
+        view = self.navigate_to(self, 'New')
+        view.fill(values)
+        view.submit.click()
+
+    def search(self, value):
+        view = self.navigate_to(self, 'All')
+        return view.search(value)
+
+    def rename(self, value, new_value):
```
oshtaier: That code should be handled by navigation part, not the entity one:
```
+        view.searchbox.search(value)
+        view.rename.click()
```
So it is more like this:
```
+    def rename(self, old_name, new_name):
+        view = self.navigate_to(self, 'Rename', entity_name=old_name)
+        view.fill(new_name)
+        view.submit.click()

+@navigator.register(ComputeProfileEntity, 'Rename')
+class RenameComputeProfile(NavigateStep):
+    VIEW = ComputeProfileDetailsView
+
+
          self.parent.search(kwargs.get('entity_name')
+        self.parent.browser.click(self.parent.rename)
```

oshtaier:
make sure to run `flake8 .` before each commit as we are following pep8 code conventions

tstrych:
Please send me a feedback, ty

airgun/views/computeprofile.py
```diff
@@ -0,0 +1,26 @@
+from widgetastic.widget import Text, TextInput
+from airgun.views.common import BaseLoggedInView, SearchableViewMixin
+from airgun.widgets import ActionsDropdown
+
+
+class ComputeProfileView(BaseLoggedInView, SearchableViewMixin):
+    title = Text("//h1[text()='Compute Profiles']")
+    new = Text("//a[contains(@href, '/compute_profiles/new')]")
+    rename = Text("//a[contains(@href, '/compute_profiles/') and "
```
oshtaier: rename button is a part of `ActionsDropdown` widget, so please implement correspondingly

airgun/views/computeprofile.py
```diff
@@ -0,0 +1,26 @@
...
+                  "contains(@href, 'edit')]")
+    actions = ActionsDropdown("//td//div[contains(@class, 'btn-group')]")
+
+    @property
+    def is_displayed(self):
+        return self.browser.wait_for_element(
+            self.title, exception=False) is not None
+
+
+class ComputeProfileDetailsView(BaseLoggedInView):
```
oshtaier: `ComputeProfileCreateView` is more appropriate name, as edit functionality has completely another appearance.
BTW, maybe you like to implement `edit` view as well, but it can be a tough task

airgun/views/computeprofile.py
```diff
@@ -0,0 +1,26 @@
+from widgetastic.widget import Text, TextInput
```
oshtaier: put imports alphabetically

airgun/entities/computeprofile.py
```diff
@@ -0,0 +1,74 @@
+from navmazing import NavigateToSibling
+
+from airgun.entities.base import BaseEntity
+from airgun.navigation import NavigateStep, navigator
+from airgun.views.computeprofile import (
+    ComputeProfileView,
```
oshtaier: put imports alphabetically

airgun/views/computeprofile.py
oshtaier: however, it is probably, not edit actually, but like actually operating and applying correct parameters. Anyway, I am talking about that view:
image

airgun/session.py
```diff
@@ -204,6 +205,11 @@ def architecture(self):
         """Instance of Architecture entity."""
         return ArchitectureEntity(self.browser)
 
+    @cached_property
+    def computeprofile(self):
+        """Instance of ComputeProfileEntity entity."""
```
oshtaier: `Compute Profile`

airgun/entities/computeprofile.py
```diff
@@ -0,0 +1,74 @@
...
+    def search(self, value):
+        view = self.navigate_to(self, 'All')
+        return view.search(value)
+
+    def rename(self, old_name, new_name):
+        view = self.navigate_to(self, 'Rename', entity_name=old_name)
+        view.fill(new_name)
+        view.submit.click()
+
+    def delete(self, entity_name):
```
oshtaier: Delete navigation step looks redundant. You can use something like this:
```
    def delete(self, entity_name):
        view.searchbox.search(entity_name)
        view.actions.fill('Delete')
        self.browser.handle_alert()
```

oshtaier: and method for rename is valid as it is actually get you to another view

airgun/views/computeprofile.py
tstrych: you always put widgetastic.widget at the top, so I just forget the empty line, if I am right.
@oshtaier 
And what is the reason these lines and navmazing in entities are always on the top and also are entered from others ? It's like library we used but we don't implent them and modules which we implement after that ?

airgun/views/computeprofile.py
tstrych: I think It should be easier to implement this after the compute resources will be finally added. 
If I am understand properly,  you should implement test which create some resource and than apply the profile on some resource and change the parameters.

tstrych:
Please take a look at last commit

tstrych:
@oshtaier please take a look at latest commit

airgun/views/computeprofile.py
oshtaier: probably reason for that is a common one:
first goes simple `imports` without `from`
then third party tools like widgetastic or nailgun
then our internal library modules

airgun/entities/computeprofile.py
```diff
@@ -0,0 +1,64 @@
...
+@navigator.register(ComputeProfileEntity, 'Rename')
+class RenameComputeProfile(NavigateStep):
+    VIEW = ComputeProfileCreateView
+
+    def prerequisite(self, *args, **kwargs):
+        return self.navigate_to(self.obj, 'All')
+
+    def step(self, *args, **kwargs):
+        self.parent.search(kwargs.get('entity_name'))
+        self.parent.browser.click(self.parent.actions)
```
oshtaier: it doesn't seem as valid change, because `actions` is `ActionsDropdown` widget which should be operated by `fill` method and it will click or choose item depends on its own logic

tstrych:
@oshtaier look at latest commit, is this better solution I use method fill of ActionDropdown but now in views I have the same actions at View and CreateView, one for delete and one for rename is there a way how to remove this redundancy ?

tstrych:
@oshtaier look at latest commit, is this better solution I use method fill of ActionDropdown but now in views I have the same actions at View and CreateView, one for delete and one for rename is there a way how to remove this redundancy ?

oshtaier:
That is obviously ot right as you don't have dropdown in CreateView as far as I know, so you need to say something like view.parent.actions

abalakh:
If i got your conversation right, you should try smth like:
```python
        self.parent.search(kwargs.get('entity_name'))
        self.parent.actions.fill('Rename')
```
And remove actions widget from createview

tstrych:
sorry i forgot on typo in session

abalakh:
@tstrych please also rebase as your branch has conflicts

tstrych:
@oshtaier @abalakh it should be ok now, should I take another good first issue ?

abalakh:
@tstrych appreciated.

tstrych:
#24   please we forgot to change the state of this issue @oshtaier  @abalakh

PR #79, opened 2018-05-03
Increase timeout for checking if the action has completed. It could timeout in some cases, for example the 'Load Datacenters' button for loading data from external RHEV.

lhellebr:


airgun/browser.py
```diff
@@ -477,6 +477,9 @@ class AirgunBrowserPlugin(DefaultPlugin):
         }
         '''
 
+    def ensure_page_safe(self, timeout='30s'):
+        super().ensure_page_safe(timeout)
```
abalakh: Pls use following format:
```python
return super(AirgunBrowserPlugin, self).ensure_page_safe(timeout)
```
Currently it's still technically possible to execute airgun & tests on py2, and this py2-uncompatible format doesn't bring much value to drop backwards compatibility.
Also please add small 1-linear docstring.

omaciel: @abalakh to be honest I don't think that we should support python 2 at all. One of the goals I set for AirGun was to be python 3 only but if we have folks using it with python 2, let me know so I can gently discourage them from doing so.

omaciel: @abalakh I disagree that we should write Python 2-compliant code just because a Jenkins job does not handle Python 3 yet. Let's get Kedar to update that process but let us not compromise on supporting Python 3 only, please?

omaciel: 👍

lhellebr:
Docstring added. Is everything else OK now?

PR #82, opened 2018-05-08
Add Products entity(part_1)

oshtaier:
Going to add basic things in that PR and add more views in the next ones

### DIFF
--- a/airgun/views/product.py
+++ b/airgun/views/product.py
@@ -1,0 +1,72 @@
+from widgetastic.widget import Select, Text, TextInput, View
+
+from airgun.views.common import (
+    BaseLoggedInView,
+    SatTab,
+    SearchableViewMixin,
+)
+from airgun.widgets import (
+    ActionsDropdown,
+    ConfirmationDialog,
+    EditableEntry,
+    EditableEntrySelect,
+    ReadOnlyEntry,
+)
+
+
+class ProductsTableView(BaseLoggedInView, SearchableViewMixin):
+    title = Text("//h2[contains(., 'Products')]")
+    new = Text("//button[contains(@href, '/products/new')]")
+    edit = Text(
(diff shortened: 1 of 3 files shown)


airgun/views/product.py
```diff
@@ -0,0 +1,72 @@
...
+
+class ProductsTableView(BaseLoggedInView, SearchableViewMixin):
+    title = Text("//h2[contains(., 'Products')]")
+    new = Text("//button[contains(@href, '/products/new')]")
+    edit = Text(
+        "//td/a[contains(@ui-sref, 'product.repositories') and "
+        "contains(@href, 'products')]"
+    )
+    repo_discovery = Text("//button[contains(.,'Repo Discovery')]")
+    actions = ActionsDropdown("//td//div[contains(@class, 'btn-group')]")
```
abalakh: Could you please point me where's that actions list inside table cell? I'm a bit confused, but that's how table view looks for me (and this locator doesn't work there for obvious reasons):
image

airgun/views/product.py
```diff
@@ -0,0 +1,72 @@
...
+    actions = ActionsDropdown("//div[contains(@class, 'btn-group')]")
+    dialog = ConfirmationDialog()
+
+    @property
+    def is_displayed(self):
+        return self.browser.wait_for_element(
+            self.return_to_all, exception=False) is not None
+
+    @View.nested
+    class Details(SatTab):
```
abalakh: nitpick: lowercase

airgun/views/product.py
oshtaier: will rename class work for you?

oshtaier: because I was talking about 'select action' dropdown obviously

abalakh: You misunderstood me - key words were "inside table cell". Your locator points to some `//td`, but there's no table surrounding that actions dropdown. Wrong locator then?

PR #85, opened 2018-05-14
partition table entity implementation

tstrych:
Please give me a feedback, @abalakh  @oshtaier  

And here are questions, some things I didn't quite understand:
1.ACEEditor using fill method with pure string for create, works fine, but then when I clone I use read|fill method again to change or clone for ACEEditor and it doesn't work, the template is empty.  Do I fill ACEEditor properly , what I should change?
2.audit commit cannot be asserted with read because it's equal to "" why ?
3.How to test lock and unlock ?
4.How I can check if the help is not empty?
5.Is in organizations always Default Organization ?
6.Is necessary to write negative tests ?
7.And of course I didn't know how to solve this conditional widget so I improvise
I have to still work on 
`tier2 tests`

### DIFF
--- a/airgun/entities/partitiontable.py
+++ b/airgun/entities/partitiontable.py
@@ -23,6 +23,11 @@ class PartitionTableEntity(BaseEntity):
         view = self.navigate_to(self, 'All')
         return view.search(value)
 
+    def update(self, values, entity_name):
+        view = self.navigate_to(self, 'Edit', entity_name=entity_name)
+        view.fill(values)
+        view.submit.click()
+
     def clone(self, values, entity_name):
         view = self.navigate_to(self, 'Clone', entity_name=entity_name)
         view.fill(values)
--- a/airgun/entities/partitiontable.py
+++ b/airgun/entities/partitiontable.py
@@ -4,7 +4,7 @@ from airgun.entities.base import BaseEntity
 from airgun.navigation import NavigateStep, navigator
 from airgun.views.partitiontable import (
     PartitionTableEditView,
-    PartitionTableView,
+    PartitionTablesView,
 )
 
 
@@ -30,25 +30,25 @@ class PartitionTableEntity(BaseEntity):
 
     def lock(self, entity_name):
         view = self.navigate_to(self, 'All')
-        view.searchbox.search(entity_name)
-        view.actions.fill('Lock')
+        view.search(entity_name)
+        view.table.row(name=entity_name)['Actions'].widget.fill('Lock')
 
     def unlock(self, entity_name):
         view = self.navigate_to(self, 'All')
-        view.searchbox.search(entity_name)
-        view.actions.fill('Unlock')
+        view.search(entity_name)
+        view.table.row(name=entity_name)['Actions'].widget.fill('Unlock')
         self.browser.handle_alert()
 
     def delete(self, entity_name):
         view = self.navigate_to(self, 'All')
-        view.searchbox.search(entity_name)
-        view.actions.fill('Delete')
+        view.search(entity_name)
+        view.table.row(name=entity_name)['Actions'].widget.fill('Delete')
         self.browser.handle_alert()
 
 
 @navigator.register(PartitionTableEntity, 'All')
 class ShowAllPartitionTables(NavigateStep):
-    VIEW = PartitionTableView
+    VIEW = PartitionTablesView
 
     def step(self, *args, **kwargs):
         # TODO: No prereq yet
@@ -62,7 +62,7 @@ class AddNewPartitionTable(NavigateStep):
     prerequisite = NavigateToSibling('All')
 
     def step(self, *args, **kwargs):
-        self.parent.browser.click(self.parent.new)
+        self.parent.new.click()
 
 
 @navigator.register(PartitionTableEntity, 'Edit')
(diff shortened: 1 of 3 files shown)


airgun/entities/partitiontable.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def delete(self, entity_name):
+        view = self.navigate_to(self, 'All')
+        view.searchbox.search(entity_name)
+        view.actions.fill('Delete')
+        self.browser.handle_alert()
+
+
+@navigator.register(PartitionTableEntity, 'All')
+class ShowAllComputeProfile(NavigateStep):
```
abalakh: class ShowAllPartitionTables

airgun/entities/partitiontable.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+@navigator.register(PartitionTableEntity, 'New')
+class AddNewPartitionTable(NavigateStep):
+    VIEW = PartitionTableEditView
+
+    prerequisite = NavigateToSibling('All')
+
+    def step(self, *args, **kwargs):
+        self.view.browser.wait_for_element(
```
abalakh: redundant line, `browser.click()` will trigger `ensure_page_safe` anyways

airgun/views/partitiontable.py
```diff
@@ -0,0 +1,72 @@
...
+    default = Checkbox(id='ptable_default')
+    snippet = Checkbox(id='ptable_snippet')
+    os_family = FilteredDropdown(id='s2id_ptable_os_family')
+    template = ACEEditor()
+    audit_comment = TextInput(id="ptable_audit_comment")
+    submit = Text('//input[@name="commit"]')
+
+    @View.nested
+    class locations(SatTab):
+        TAB_NAME = 'Locations'
```
abalakh: It's redundant - if no TAB_NAME specified, it's set to capitalized class name ('Locations' in this case). Explicitly setting TAB_NAME is only required for tabs with name different from class name (e.g. when tab name contains space).
Same for organizations and help tabs.

oshtaier:
2. "The Audit Comment field is saved with the template auditing to document the template changes". So it seems it is saved in the documents which will be use that template later. I have no idea where that place is, but we can spend time to investigate

oshtaier:
3. Answer is easy, when document locked - you cannot delete it, so you can for example on test layer just to try to delete the ptable, get an exception and use corresponding context manager to handle it (expect that specific error)

oshtaier:
4. Probably help should be omitted at that moment. Doesn't seem as functionality that we have to test

oshtaier:
7. Here is one working example:
https://github.com/SatelliteQE/airgun/blob/master/airgun/views/product.py#L149
Let's me know if you need more help on that

airgun/views/partitiontable.py
tstrych: in os and activation key are used TAB_NAME 's, but names are the same with the class name should I fix it ?

tstrych:
@oshtaier, yes I need your help with this.
This was what I created and in comments is written how I think about that, but it's wrong, the entity after that doesn't know to where to fill values because I move paths to conditional switchable view.
```
#  here will be PartitionTableEditView without snippet and os_family

    #  So I will create Select on my checkbox (because the checkbox do the  conditionally changes 
    # depends on check and unchecked
    snippet_type = SatSelect(locator="//input[@id='ptable_snippet']")
    os_family_selection = ConditionalSwitchableView(reference='snippet_type')

    # and now creating two possibilities, for each conditional view with their xpaths
    @os_family_selection.register('Snippet')
    class SnippetOption(View):
        snippet = Checkbox(id='ptable_snippet')

    @os_family_selection.register('OS Family')
    class OSFamilyOption(View):
        os_family = FilteredDropdown(id='s2id_ptable_os_family')
```

tstrych:
2. I found the information in Foreman It's in Monitor -> Audits -> name of created Partition Table -> Details tab (at the bottom, where the information icon is)
I think I don't have to solve this problem, I will just not assert on audit_commits.

oshtaier:
Yeah only add the widget for audit comments, but neither fill nor read them in tests at that moment

tstrych:
@oshtaier @abalakh, review please.

airgun/views/partitiontable.py
```diff
@@ -0,0 +1,76 @@
...
+    class SnippetOption(View):
+        pass
+
+    @os_family_selection.register(False)
+    class OSFamilyOption(View):
+        os_family = FilteredDropdown(id='s2id_ptable_os_family')
+
+    @View.nested
+    class locations(SatTab):
+        locations = MultiSelect(id='ms-ptable_location_ids')
```
oshtaier: maybe rename to `resources` to avoid `locations.locations.` in tests

airgun/views/partitiontable.py
```diff
@@ -0,0 +1,76 @@
...
+    FilteredDropdown,
+    MultiSelect,
+)
+
+
+class PartitionTableView(BaseLoggedInView, SearchableViewMixin):
+
+    title = Text("//h1[text()='Partition Tables']")
+    new = Text("//a[contains(@href, '/ptables/new')]")
+    actions = ActionsDropdown("//td//div[contains(@class, 'btn-group')]")
```
abalakh: Please take a brief look at #89 and its changes to entities.
We're now finally working with entities table and widgets inside its columns, instead of picking the first button/link/actionsdropdown available on page.
Basically you'll have to update:
* PartitionTableView to have table widget with column widgets defined, instead of `actions` and `edit`
* EditPartitionTable and ClonePartitionTable step() to use column widgets
* lock(), unlock() and delete() entity methods to use column widgets

PR #89 contains a lot of examples for every single entity available, so it should be pretty straightforward, but feel free to ping me in case of any questions.

tstrych:
@abalakh please review

tstrych:
@abalakh tests with results are added, and update was added into entity.

PR #87, opened 2018-05-15
Add domain views/entities and tweak MultiSelect/CustomParameter widgets

bsquizz:
* Add views, entity, navigation for Infrastructure -> Domains
* Add `set_assigned` method to `MultiSelect` widget
* Change `CustomParameter` to inherit from `Table` widget so that its easier to repeatedly add values to the param list and read from each row. Update `OperatingSystemDetailsView` to pass in `id` when invoking `CustomParameter`
* Add `wait_for` to requirements, since its now being used in the `DomainEntity` CRUD methods

### DIFF
--- a/airgun/entities/domain.py
+++ b/airgun/entities/domain.py
@@ -44,9 +44,8 @@ class DomainEntity(BaseEntity):
         view = self.navigate_to(self, 'New')
         view.fill(values)
         view.submit_button.click()
-        if assert_no_errors:
-            view.flash.assert_no_error()
-            assert_no_errors_in_view(view)
+        view.flash.assert_no_error()
+        assert_no_errors_in_view(view)
 
     def search(self, value):
         """Search for 'value' and return domain names that match.
@@ -61,7 +60,7 @@ class DomainEntity(BaseEntity):
         view = self.navigate_to(self, 'Edit', entity_name=entity_name)
         return view.read()
 
-    def update(self, entity_name, values, assert_no_errors=True):
+    def update(self, entity_name, values):
         """
         Update an existing domain.
         """
@@ -69,37 +68,32 @@ class DomainEntity(BaseEntity):
 
         view.fill(values)
         view.submit_button.click()
-        if assert_no_errors:
-            view.flash.assert_no_error()
-            assert_no_errors_in_view(view)
+        view.flash.assert_no_error()
+        assert_no_errors_in_view(view)
 
-    def add_parameter(self, entity_name, param_name, param_value,
-                      assert_no_errors=True):
+    def add_parameter(self, entity_name, param_name, param_value):
         view = self.navigate_to(self, 'Edit', entity_name=entity_name)
         view.parameters.params.add({'name': param_name, 'value': param_value})
         view.submit_button.click()
-        if assert_no_errors:
-            view.flash.assert_no_error()
-            assert_no_errors_in_view(view)
+        view.flash.assert_no_error()
+        assert_no_errors_in_view(view)
 
-    def remove_parameter(self, entity_name, param_name, assert_no_errors=True):
+    def remove_parameter(self, entity_name, param_name):
         view = self.navigate_to(self, 'Edit', entity_name=entity_name)
         view.parameters.params.remove(param_name)
         view.submit_button.click()
-        if assert_no_errors:
-            view.flash.assert_no_error()
-            assert_no_errors_in_view(view)
+        view.flash.assert_no_error()
+        assert_no_errors_in_view(view)
 
-    def delete(self, name, assert_no_errors=True):
+    def delete(self, name):
         view = self.navigate_to(self, 'All')
         self.search(name)
         if not view.table.row_count:
             raise ValueError("Unable to find name '{}'".format(name))
         view.table[0]['Actions'].widget.click()
         self.browser.handle_alert()
-        if assert_no_errors:
-            view.flash.assert_no_error()
-            assert_no_errors_in_view(view)
+        view.flash.assert_no_error()
+        assert_no_errors_in_view(view)
 
 
 @navigator.register(DomainEntity, 'All')
--- a/airgun/entities/domain.py
+++ b/airgun/entities/domain.py
@@ -20,14 +20,13 @@ class DomainEntity(BaseEntity):
         view.submit_button.click()
         view.flash.assert_no_error()
 
-    def search(self, value=None):
+    def search(self, value):
         """Search for 'value' and return domain names that match.
 
         :param value: text to filter (default: no filter)
         """
         view = self.navigate_to(self, 'All')
-        view.search(value or '')
-        return [row['Description'].text for row in view.table.rows()]
+        return view.search(value)
 
     def read(self, entity_name):
         """Return dict with properties of domain."""
--- a/airgun/entities/domain.py
+++ b/airgun/entities/domain.py
@@ -9,16 +9,44 @@ from airgun.views.domain import (
 )
 
 
+def assert_no_errors_in_view(view):
+    """
+    Check that there's no elements of class "has-error" in the view
+
+    There can be elements that have an error, and there can also be
+    error message blocks. If errors are found on the page, then
+    we will search for any displayed error messages and print
+    their text in the AssertionError
+
+    TODO: Move this somewhere for general use by other entities?
+    """
+    error_elements = view.browser.elements(
+        ".//*[contains(@class,'has-error') "
+        "and not(contains(@style,'display:none'))]"
+    )
+    if error_elements:
+        error_msgs = view.browser.elements(
+            ".//*[(contains(@class,'error-msg-block') "
+            "or contains(@class,'error-message')) "
+            "and not(contains(@style,'display:none'))]"
+        )
+        error_msgs = [view.browser.text(error_msg) for error_msg in error_msgs]
+        raise AssertionError(
+            "errors present on page, displayed messages: {}".format(error_msgs)
+        )
+
+
 class DomainEntity(BaseEntity):
-    def create(self, values, assert_no_error=True):
+    def create(self, values, assert_no_errors=True):
         """
         Create a new domain.
         """
         view = self.navigate_to(self, 'New')
         view.fill(values)
         view.submit_button.click()
-        if assert_no_error:
+        if assert_no_errors:
             view.flash.assert_no_error()
+            assert_no_errors_in_view(view)
 
     def search(self, value):
         """Search for 'value' and return domain names that match.
@@ -33,7 +61,7 @@ class DomainEntity(BaseEntity):
         view = self.navigate_to(self, 'Edit', entity_name=entity_name)
         return view.read()
 
-    def update(self, entity_name, values, assert_no_error=True):
+    def update(self, entity_name, values, assert_no_errors=True):
         """
         Update an existing domain.
         """
@@ -41,33 +69,37 @@ class DomainEntity(BaseEntity):
 
         view.fill(values)
         view.submit_button.click()
-        if assert_no_error:
+        if assert_no_errors:
             view.flash.assert_no_error()
+            assert_no_errors_in_view(view)
 
     def add_parameter(self, entity_name, param_name, param_value,
-                      assert_no_error=True):
+                      assert_no_errors=True):
         view = self.navigate_to(self, 'Edit', entity_name=entity_name)
         view.parameters.params.add({'name': param_name, 'value': param_value})
         view.submit_button.click()
-        if assert_no_error:
+        if assert_no_errors:
             view.flash.assert_no_error()
+            assert_no_errors_in_view(view)
 
-    def remove_parameter(self, entity_name, param_name, assert_no_error=True):
+    def remove_parameter(self, entity_name, param_name, assert_no_errors=True):
         view = self.navigate_to(self, 'Edit', entity_name=entity_name)
         view.parameters.params.remove(param_name)
         view.submit_button.click()
-        if assert_no_error:
+        if assert_no_errors:
             view.flash.assert_no_error()
+            assert_no_errors_in_view(view)
 
-    def delete(self, name, assert_no_error=True):
+    def delete(self, name, assert_no_errors=True):
         view = self.navigate_to(self, 'All')
         self.search(name)
         if not view.table.row_count:
             raise ValueError("Unable to find name '{}'".format(name))
         view.table[0]['Actions'].widget.click()
         self.browser.handle_alert()
-        if assert_no_error:
+        if assert_no_errors:
             view.flash.assert_no_error()
+            assert_no_errors_in_view(view)
 
 
 @navigator.register(DomainEntity, 'All')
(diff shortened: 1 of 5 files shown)


airgun/entities/domain.py
```diff
@@ -0,0 +1,192 @@
+from navmazing import NavigateToSibling
+from wait_for import wait_for
+
+from airgun.entities.base import BaseEntity
+from airgun.navigation import NavigateStep, navigator
+from airgun.views.domain import (
+    DomainCreateView,
```
oshtaier: put alphabetically

airgun/entities/domain.py
```diff
@@ -0,0 +1,192 @@
...
+    DomainListView,
+    DomainEditView
+)
+
+
+class DomainEntity(BaseEntity):
+    _DEFAULT_ORG = 'Default Organization'
```
oshtaier: These constants already defined in robottelo and basically all such data should be checked on a test level. Framework should only pass data and return it back to the tests

airgun/entities/domain.py
```diff
@@ -0,0 +1,192 @@
...
+    _DEFAULT_LOC = 'Default Location'
+    _CREATE_SUCCESSFUL = 'Successfully created {}.'
+    _UPDATE_SUCCESSFUL = 'Successfully updated {}.'
+    _DELETE_SUCCESSFUL = 'Successfully deleted {}.'
+
+    @staticmethod
+    def _submit_domain_form(view, **kwargs):
```
oshtaier: that whole method is redundant as test layer handle what is passed to entity and on entity level you only should say `view.fill()`, so necessary fields/widgets will be filled
refer to other entities implementation to get the idea

airgun/entities/domain.py
```diff
@@ -0,0 +1,192 @@
...
+            Default: ['Default Location']
+        :param organizations: list of org strings to be assigned.
+            All other orgs will be unassigned.
+            Default: ['Default Organization']
+        """
+        view = self.navigate_to(self, 'New')
+
+        kwargs['dns_domain'] = dns_domain
+
+        if not kwargs.get('locations'):
```
oshtaier: all that data is also redundant as you should specify nothing in case it was not required on a test level or if such assignment is mandatory then test still should handle that on its own. Yes, that will take more efforts and lines of code for tests, but you will avoid situation when framework do things that it should not do

airgun/entities/domain.py
```diff
@@ -0,0 +1,192 @@
...
+        if not kwargs.get('organizations'):
+            kwargs['organizations'] = [self._DEFAULT_ORG]
+
+        self._submit_domain_form(view, **kwargs)
+
+        wait_for(
+            lambda: view.flash.messages, fail_condition=[], timeout=10,
+            delay=2, message='wait for flash success'
+        )
+        view.flash.assert_success_message(self._CREATE_SUCCESSFUL.format(dns_domain))
```
oshtaier: so you always expect that you procedure should be finished successfully which is also wrong as we need to perform a lot of negative testing, so that validation is also needs to be performed on test layer or at least we should be able to say what we want to expect in the end

airgun/entities/domain.py
```diff
@@ -0,0 +1,192 @@
...
+
+    def search(self, value=None):
```
oshtaier: we already have `SearchableViewMixin` that should handle all the logic that we need, so usually we just use:

```
        return view.search(value, expected_result)
```

airgun/entities/domain.py
```diff
@@ -0,0 +1,192 @@
...
+        :param value: text to filter (default: no filter)
+        """
+        view = self.navigate_to(self, 'All')
+        view.search(value or '')
+        return [row['Description'].text for row in view.table.rows()]
+
+    def read(self, domain_name):
+        """Return dict with properties of domain."""
+        view = self.navigate_to(self, 'Edit', entity_name=domain_name)
+        properties = dict(
```
oshtaier: that data is also seem redundant as entity layer should know nothing about structure of the view, so we usually write:
```
        return view.read()
```

airgun/entities/domain.py
```diff
@@ -0,0 +1,192 @@
...
+            full_name=view.domain.full_name.read(),
+            dns_capsule=view.domain.dns_capsule.read(),
+            parameters=view.parameters.params.read(),
+            locations=view.locations.multiselect.read()['assigned'],
+            organizations=view.organizations.multiselect.read()['assigned']
+        )
+        # Return only non-null entries
+        return {key: val for key, val in properties.items() if val}
+
+    def edit(self, domain_name, **kwargs):
```
oshtaier: `def update(self, domain_name, **kwargs):` as it is CRUD

airgun/entities/domain.py
```diff
@@ -0,0 +1,192 @@
...
+
+
+@navigator.register(DomainEntity, 'New')
+class AddNewDomain(NavigateStep):
+    VIEW = DomainCreateView
+
+    prerequisite = NavigateToSibling('All')
+
+    def step(self, *args, **kwargs):
+        self.view.browser.wait_for_element(
```
oshtaier: that is redundant waiter as click() do that anyway

airgun/session.py
```diff
@@ -27,6 +27,7 @@
 from airgun.entities.subnet import SubnetEntity
 from airgun.entities.syncplan import SyncPlanEntity
 from airgun.entities.user import UserEntity
+from airgun.entities.domain import DomainEntity
```
oshtaier: put alphabetically

airgun/widgets.py
```diff
@@ -279,6 +279,27 @@ def fill(self, values):
                 self.assigned.fill(value)
         return True
 
+    def set_assigned(self, desired_values):
```
oshtaier: not sure why that situation should happen. Are tests not isolated or you do things in one organization?

airgun/entities/domain.py
```diff
@@ -0,0 +1,192 @@
...
+
+        # If domain is renamed, flash message contains the new name...
+        updated_name = kwargs.get('dns_domain') or domain_name
+        view.flash.assert_success_message(
+            self._UPDATE_SUCCESSFUL.format(updated_name))
+
+    def delete(self, name):
+        view = self.navigate_to(self, 'All')
+        self.search(name)
+        row = view.table.row(('Description', name))
```
oshtaier: probably you should define that your table contains button action as third column and access it like that, so it will be on view layer, not on entity one that basically should not care about that at all

airgun/widgets.py
```diff
@@ -489,39 +513,58 @@ class CustomParameter(Widget):
...
-    NAMES = (
-        ".//tr[contains(@id, 'os_parameter')]/td/input[@placeholder='Name']")
-    VALUE = (
-        ".//table[contains(@id, 'parameters')]//tr"
-        "/td[input[contains(@id, 'name')][contains(@value, '{}')]]"
-        "/following-sibling::td//textarea"
-    )
+    add_new_value = Text("..//a[contains(text(),'+ Add Parameter')]")
+
+    def __init__(self, *args, **kwargs):
```
oshtaier:
```
def __init__(self, parent, locator=None, id=None, logger=None):
        """Supports initialization via ``locator=`` or ``id=``"""
        if locator and id or not locator and not id:
            raise TypeError('Please specify either locator or id')
```
approach looks more readable and better for understanding. And you will not need these pops

airgun/widgets.py
```diff
@@ -489,39 +513,58 @@ class CustomParameter(Widget):
...
 
     def read(self):
         """Return a list of dictionaries. Each dictionary consists of name and
         value parameters
         """
         parameters = []
-        for item in self.browser.elements(self.NAMES):
-            name = self.browser.get_attribute('value', item)
-            value = self.browser.text(self.VALUE.format(name))
+        for row in self.rows():
```
oshtaier: if you actually transforming that widget to table, so see no reason for extra logic. It should be simple `table.read()`

setup.py
```diff
@@ -18,6 +18,7 @@
         'navmazing==1.1.4',
         'pytest==3.4.0',
         'six==1.11.0',
+        'wait_for',
```
oshtaier: that is probably already added in other PRs, so just make rebase for current one

airgun/entities/domain.py
abalakh: it's already defined, @bsquizz is just using a bit different style of identifying needed row (which is fine as it's supported & documented way)

airgun/widgets.py
```diff
@@ -489,39 +513,58 @@ class CustomParameter(Widget):
...
         """
         self.add_new_value.click()
-        self.new_parameter_name.fill(values['name'])
-        self.new_parameter_value.fill(values['value'])
+        new_row = self.__getitem__(-1)
+        new_row['Name'].widget.fill(value['name'])
+        new_row['Value'].widget.fill(value['value'])
+
+    def fill(self, values):
+        """Fill parameter entries. Existing values will be overwritten.
```
abalakh: This change may effect existing tests, we need to make sure they're still working as expected before merging.

airgun/entities/domain.py
bsquizz: dns_domain is the only required arg, I made it a required arg for creating a domain since it is required in the UI as well. I could make it a non-required kwarg in case people want to test creating a domain without properly filling in dns_domain. The other args cab all be 'None' and nothing in the form will be filled.

I can also change things to not fill locations and organizations with defaults if those are 'None'

airgun/entities/domain.py
bsquizz: My impression of this is the test layer does handle what's passed. This is a helper method to fill the fields in the view. My intention was for a tester to pass in the args they want to fill with something like this:

```python
session.domain.create(dns_domain='testdomain100', parameters=[{'name': 'myname', 'value': 'myvalue'}, {'name': 'myname2', 'value': 'myvalue2'}], locations=['Default Location'], organizations=['myorg1', 'myorg2'])
```

airgun/widgets.py
bsquizz: This is a helper method so that I can pass in 1 list of values I want in the 'assigned' box and it takes care of un-assigning anything else that is in there

airgun/entities/domain.py
bsquizz: Right, I noticed a few flaws with that Mixin so I extended this method slightly. For example, if you tried to do a `view.search('a')` you'd get back the first text on the page that has the character `a` on it. This would be `Default Organization`. I do realize you can change the `search_result_locator`, but what if you have multiple entries with 'a' in their name? So I decided to have it iterate through the rows on the search table and return a list of all matches that search has returned.

If every page that we use `SearchableViewMixin` has a table on it I could move this logic into there... but we'd have to be careful if tests are expecting to get 1 string back instead of a list.

airgun/entities/domain.py
bsquizz: Ok, so I can move the collections of the properties here into a read() method under the Edit view if that works? Then here we'd just `return view.read()` as you suggest.

airgun/widgets.py
bsquizz: I doing it this way so that I don't alter the return format (a list of dictionaries with name/value keys). Do we want to change the returned content?

airgun/entities/domain.py
bsquizz: Correct, if after doing `search` the row with that "Description" is not found, it'll raise a RowNotFound error. So this ensures the row we're looking for after the `search()` is in there.

airgun/entities/domain.py
abalakh: Personally i'd left flash notification type check here (assert_no_errors) - that one will still allow us to do negative testing and won't require any text constants in airgun.
Negative test will just ensure exception is risen after creation:
```python
with raises(AssertionError):
    session.domain.create('bad_inputs')
```
Take a look at https://github.com/SatelliteQE/airgun/blob/master/airgun/entities/activationkey.py#L45 , i'm checking whether notification appeared and whether it's a good one.

Regarding messages testing - that's a different story, you've suggested one of options but there probably are others. Let's discuss that in separate issue when we actually hit the test which requires such message check. AFAIK they're not numerous. In any case - i agree with @oshtaier - we want to isolate layers and not to introduce exact expected messages (which may change quite often) in airgun, they should lie in robottelo.

airgun/entities/domain.py
abalakh: That's what i'm currently working on - updating that mixing to return table values, not just string by locator. Pls postpone this comment for now and resolve others - i hope i'll submit PR soon so you'll just update yours to use updated mixin.

airgun/entities/domain.py
bsquizz: OK I see, so your tests just pass in the raw view dict data

bsquizz:
Most requests addressed, I have 2 questions above, see the review comments which are still remaining.

airgun/widgets.py
oshtaier: probably, you are right. For that specific widget we only interested in these two fields values

airgun/widgets.py
oshtaier: Yeah, such situation can be required in some unique situation, when we had (at least I hope that we don't have them anymore) entities that were not organization specific or we cannot filter them in some way and you cannot predict what entities will be in the lists. But then, we still can assert that some entities are `in` not `equal`
Please, remove that method and we will add it in the future in case we face something described above

oshtaier:
Pending 1 comment from my side, fixing travis, final review from @abalakh and proper tests for all these code

airgun/entities/domain.py
```diff
@@ -0,0 +1,81 @@
...
+        view.flash.assert_no_error()
+
+    def search(self, value=None):
+        """Search for 'value' and return domain names that match.
+
+        :param value: text to filter (default: no filter)
+        """
+        view = self.navigate_to(self, 'All')
+        view.search(value or '')
+        return [row['Description'].text for row in view.table.rows()]
```
abalakh: pls rebase your PR and update to:
```python
        return view.search(value)
```
as #89 is already merged

bsquizz:
@oshtaier @abalakh all requests addressed now, rebased, and lint issues in travis runs are fixed now as well. Let me know if you want me to squash.

oshtaier:
Only reviewers can squash and merge PRs
I think that I am fine with code. @abalakh please check it once more in details
And now, you need to port tier2 tests from https://github.com/SatelliteQE/robottelo/blob/master/tests/foreman/ui/test_domain.py to make sure that your changes here are actually working
Check our other ported modules to avoid extra efforts. We need to have tests that are actually make sense, so try to unite test cases or remove them if they do almost same things or redundant things and etc.

bsquizz:
@oshtaier @abalakh take a look at the last 3 commits. I came up with a temp solution to check for error message blocks within the Domain views. In future we can move this to a more general spot that all views/entities could use, but we shouldn't have to change the tests as long as the text of the AssertionError remains similar (the exception will display the text of the error msgs displayed on the page).

I also am now dismissing flash messages on the Search widget, since large flash messages can hide the search button.

All the linked tests in the linked robottelo PR are passing.

bsquizz:
Also forgot to mention I had to update the fill method on the `CustomParameter` widget, we now check to see if a name in the desired values is already being used on the page, if so, we re-use it and simply update that name's value. This is because if a name is removed, it is only hidden, and adding a new line which uses the same name with throw a 'name already in use' error. I'll open a bug for this.

PR #89, opened 2018-05-17
Refacored entities list views and SearchableViewMixin, both are using table widget and rows

abalakh:
Summary of changes:
* Entity list views are now correctly using Table widget instead of search+relative locators
* SearchableViewMixin now relies on `table` class attr and returns `table.read()` as search result
* Updated all entities helpers to use table column widgets instead of direct links/buttons
* Added few content view helpers and navigate steps for upcoming test coverage
* Added few notifications asserts and dismisses where they could overlap widgets and cause failures
* Fixed SatTab's `is_displayed` property, as for some reason it was still 'visible' to selenium when actually hidden on UI
* Made LCESelector widget use relative xpaths when selecting some lce, so now it will select 'Library' from specific LCESelector, not the first one found on page
* Updated SatTable and SatSecondaryTab to support newly found table/tab with different style
* Other small fixes

### DIFF
--- a/airgun/views/common.py
+++ b/airgun/views/common.py
@@ -84,8 +84,7 @@ class SatSecondaryTab(Tab):
             TAB_NAME = 'List/Remove'
     """
     ROOT = ParametrizedLocator(
-        './/nav[@class="ng-scope" or not(@class) and not(@data-block)]'
-        '/following-sibling::div')
+        './/nav[@class="ng-scope" or not(@*)]/following-sibling::div')
 
 
 class LCESelectorGroup(ParametrizedView):
@@ -309,8 +308,8 @@ class SearchableViewMixin(WidgetMixin):
     searchbox = Search()
 
     def search(self, query):
-        """Perform search using searchbox on the page and return element text
-        if found.
+        """Perform search using searchbox on the page and return table
+        contents.
 
         :param str query: search query to type into search field. E.g. ``foo``
             or ``name = "bar"``.


airgun/entities/contentview.py
```diff
@@ -102,8 +134,46 @@ def prerequisite(self, *args, **kwargs):
...
     def step(self, *args, **kwargs):
-        self.parent.search(kwargs.get('entity_name'))
-        self.parent.edit.click()
+        entity_name = kwargs.get('entity_name')
+        self.parent.search(entity_name)
+        self.parent.table.row(name=entity_name)['Name'].widget.click()
+
+
+@navigator.register(ContentViewEntity, 'AddPuppetModule')
+class AddNewPuppetModule(NavigateStep):
```
oshtaier: remind me to discuss necessity of that class tomorrow

airgun/views/common.py
```diff
@@ -81,7 +84,8 @@ class listremove(SatSecondaryTab):
             TAB_NAME = 'List/Remove'
     """
     ROOT = ParametrizedLocator(
-        './/nav[@class="ng-scope"]/following-sibling::div')
+        './/nav[@class="ng-scope" or not(@class) and not(@data-block)]'
```
oshtaier: anything simpler?

abalakh: i found secondary tab which has no attrs at all - just `<nav>`

airgun/entities/contentview.py
oshtaier: skip on this

airgun/views/common.py
oshtaier: ok, spend 20 mins more on that per our discussion

abalakh:
@oshtaier addressed.
Test results:
```python
pytest -v tests/foreman/ui_airgun/test_contentview.py -k test_positive_create_composite
platform darwin -- Python 3.6.4, pytest-3.4.0, py-1.5.3, pluggy-0.6.0 -- /Users/andrii/workspace/py3a/bin/python3.6
cachedir: .pytest_cache
plugins: wait-for-1.0.9, xdist-1.22.0, services-1.2.1, mock-1.6.3, forked-0.2, env-0.6.2
collected 10 items
2018-05-18 16:57:22 - conftest - DEBUG - BZ deselect is disabled in settings


tests/foreman/ui_airgun/test_contentview.py::test_positive_create_composite PASSED [100%]

============================== 9 tests deselected ==============================
=================== 1 passed, 9 deselected in 319.15 seconds ===================
```

PR #96, opened 2018-05-30
Drop six & py2 backwards compatibility

abalakh:
Per @omaciel request here: https://github.com/SatelliteQE/airgun/pull/79#discussion_r188379682

### DIFF
--- a/airgun/browser.py
+++ b/airgun/browser.py
@@ -4,8 +4,6 @@ tests.
 import logging
 import time
 
-import six
-
 from datetime import datetime
 from fauxfactory import gen_string
 from selenium import webdriver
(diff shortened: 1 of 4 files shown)


oshtaier:
Can we prove in some way that we really remove everything we need?
Do we need to run any tests?

abalakh:
@oshtaier latest PR with 10 CV tests was executed with these changes applied. Regarding proof - well, six is great compatibility lib, basically dropping it alone with all the usages should be good enough. These are all the places i'm aware of, if you find smth else - please let me know.

abalakh:
@ekohl rather the opposite - robottelo dropped py2 support some time ago, that's why there was no point in keeping airgun backwards-compatible. Not all py2 workarounds were removed from robottelo but there's definitely some py2-incompatible code merged already.
But that only applies to master branch (6.4), previous versions should be still py2-compatible afaik.

PR #102, opened 2018-06-06
Add repos functionality

oshtaier:


airgun/views/product.py
```diff
@@ -122,6 +122,10 @@ class details(SatTab):
         tasks_count = ReadOnlyEntry(name='Active Tasks')
         sync_plan = EditableEntrySelect(name='Sync Plan')
 
+    @View.nested
+    class repositories(SatTab):
```
abalakh: This tab shouldn't be empty, but rather include `RepositoriesView`, as otherwise you're unable to fetch product's repositories. Next time with separate issue for that?

oshtaier: Right, as that require adding some validation test cases for product entity
I will do that as separate effort

PR #105, opened 2018-06-13
adding more functionality to location

tstrych:
It's not done yet. You can find some TODO's there.
But please give me a feedback @oshtaier or @abalakh  on what is already written.

### DIFF
--- a/airgun/views/location.py
+++ b/airgun/views/location.py
@@ -1,12 +1,16 @@
-from widgetastic.widget import Text, TextInput
+from widgetastic.widget import Checkbox, Text, TextInput, View
 from widgetastic_patternfly import BreadCrumb
 
 from airgun.views.common import (
     BaseLoggedInView,
+    SatVerticalTab,
     SearchableViewMixin,
+
 )
 from airgun.widgets import (
     ActionsDropdown,
+    CustomParameter,
+    MultiSelect,
     SatTable,
 )
 
@@ -28,6 +32,7 @@ class LocationsView(BaseLoggedInView, SearchableViewMixin):
             self.title, exception=False) is not None
 
 
+# TODO add filtered dropdown to create and Edit
 class LocationCreateView(BaseLoggedInView):
     breadcrumb = BreadCrumb()
     name = TextInput(id='location_name')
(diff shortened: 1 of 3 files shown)


airgun/views/location.py
```diff
@@ -43,3 +48,94 @@ def is_displayed(self):
...
+        all_hostgroups = Checkbox(id='location_ignore_types_hostgroup')
+        resources = MultiSelect(id='ms-location_hostgroup_ids')
+
+    @View.nested
+    class organizations(SatVerticalTab):
+        resources = MultiSelect(id='ms-location_organization_ids')
+
+    @View.nested
+    class parameters(SatVerticalTab):
+        resources = CustomParameter()
```
oshtaier: https://github.com/SatelliteQE/airgun/pull/87 is submitted, so:
`resources = CustomParameter(id='global_parameters_table')`
and fix for organization file too

jhutar:
Tomas is working on this.

tstrych:
@abalakh or @oshtaier please give me a feedback

airgun/views/location.py
```diff
@@ -43,3 +49,95 @@ def is_displayed(self):
...
+            self.breadcrumb, exception=False)
+        return (
+                breadcrumb_loaded
+                and self.breadcrumb.locations[0] == 'Locations'
+                and self.breadcrumb.read().starts_with('Edit ')
+        )
+
+    @View.nested
+    class primary(SatVerticalTab):
+        parent_id = FilteredDropdown(id='location_parent_id')
```
abalakh: why `parent_id` and not just `parent`?
image

airgun/views/location.py
```diff
@@ -30,6 +35,7 @@ def is_displayed(self):
 
 class LocationCreateView(BaseLoggedInView):
     breadcrumb = BreadCrumb()
+    parent_id = FilteredDropdown(id='location_parent_id')
```
abalakh: same here

tstrych:
@abalakh  and @oshtaier  it looks good now ? ACK ? maybe you should look before that at the tests in robotello 
[https://github.com/SatelliteQE/robottelo/pull/6064](url)

airgun/views/location.py
tstrych: @abalakh With just parent it doesn't work, it looks like I rewrite parent somewhere else and then airgun browser want to fill parent but he doesn't have method fill and it will cause in widget not implemented error. Maybe you can explain to me which parent method I rewrote to better understanding.

abalakh: Oh, right. Every view or widget have `.parent` attr which basically points to either browser or another parent widget.
Maybe `parent_location` then? I'd never guess `parent` dropdown is called `parent_id` :)

PR #109, opened 2018-06-14
Unpin pytest and let pinning be controlled by robottelo

lpramuk:
We pin pytest in robottelo already, no need to pin in airgun aswell
So that we avoid pinning conflicts or maintaining pinning to be the same in various projects

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,7 +17,7 @@ setup(
         'cached_property',
         'fauxfactory',
         'navmazing==1.1.4',
-        'pytest==3.4.0',
+        'pytest',
         'wait_for',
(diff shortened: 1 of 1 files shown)


lpramuk:
Oops
https://pyup.io/changelogs/pytest/#3.5.0 
```
- Defining ``pytest_plugins`` is now deprecated in non-top-level conftest.py
files, because they "leak" to the entire directory tree. (`3084
<https://github.com/pytest-dev/pytest/issues/3084>`_)
```
vs.

https://github.com/SatelliteQE/airgun/blob/master/tests/conftest.py#L4
```
pytest_plugins = ["airgun.fixtures"]
```

Issue #119, opened 2018-07-04
Ensure every create/update/etc helper validates notification messages


abalakh:
Every single entity helper which triggers operation with notification like `Entity was successfully created` should validate that notification as a part of that helper by using just
```python
view.flash.assert_no_error()
view.flash.dismiss()
```
Example - https://github.com/SatelliteQE/airgun/blob/master/airgun/entities/contentview.py#L53

We already have few failures which couldn't create some entity, didn't check the message, then proceeded with test steps and failed later with completely different exceptions, making it harder to understand what's the root cause of failure from the first glance.
E.g. test_positive_v3_wui_can_add_resource failed with the following message:
```
NoSuchElementException('Could not find an element {}'.format(repr(locator)))
selenium.common.exceptions.NoSuchElementException: Message: Could not find an element "//input[@id='search' or @ng-model='table.searchTerm' or contains(@ng-model, 'Filter')]"
```
but the real root cause was error during compute resource creation, which would've been caught by `view.flash.assert_no_error()` and descriptive exception message would've been shown.

PR #155, opened 2018-07-17
Add puppet class and smart variable functionality

oshtaier:
Not all functionality parts can be implemented at that moment due widgetastic restrictions, but all code present here were tested at least in debug mode. Maybe, it will require some polishing, but it will be much easier to do for whoever start to work with tests in that area


airgun/views/puppet_class.py
```diff
@@ -0,0 +1,58 @@
...
+        puppet_environment = TextInput(id='puppetclass_environments')
+        host_group = MultiSelect(id='ms-puppetclass_hostgroup_ids')
+
+    @View.nested
+    class smart_class_parameter(SatTab):
+        TAB_NAME = 'Smart Class Parameter'
+        pass
+
+    @View.nested
+    class smart_class_variables(SatTab):
```
abalakh: ` class smart_variables`, according to tab name

airgun/views/smart_variable.py
```diff
@@ -0,0 +1,149 @@
...
+        expander = Text(
+            ".//h2[contains(@data-target, '#optional_input_validators_')]")
+        validator_type = Select(
+            locator=".//select[contains(@name, '[validator_type]')]")
+        validator_rule = TextInput(
+            locator=".//input[contains(@name, '[validator_rule]')]")
+
+        def __init__(self, parent, logger=None):
+            View.__init__(self, parent, logger=logger)
+            if 'collapsed' in self.browser.element(
```
abalakh: why not `self.browser.classes`?

airgun/views/smart_variable.py
```diff
@@ -0,0 +1,149 @@
...
+                'Attribute type': MatcherAttribute(),
+                'Value': TextInput(
+                    locator=".//textarea[contains(@id, 'value')]"),
+            }
+        )
+        add_new_matcher = Text(
+            ".//a[contains(@data-original-title, 'add a new matcher')]")
+
+        def fill(self, values):
+            """Example
```
abalakh: nitpick: most likely such format will cause issues with building and/or displaying the docs. Could you please format this docstring properly?

airgun/entities/puppet_class.py
```diff
@@ -0,0 +1,58 @@
...
+    def read(self, entity_name):
+        """Read puppet class entity values"""
+        view = self.navigate_to(self, 'Edit', entity_name=entity_name)
+        return view.read()
+
+    def update(self, entity_name, values):
+        """Update puppet class values"""
+        view = self.navigate_to(self, 'Edit', entity_name=entity_name)
+        view.fill(values)
+        view.submit.click()
```
abalakh: Let's not increase a scope of #119 and make a rule to add all necessary flash notifications checks for all new entities, what do you think? :)

airgun/entities/host.py
```diff
@@ -62,6 +62,14 @@ def change_host_environment(self, entities_list, values):
         view.flash.assert_no_error()
         view.flash.dismiss()
 
+    def read_yaml_output(self, entity_name):
+        """Get puppet external nodes YAML dump for specific host"""
+        view = self.navigate_to(self, 'Details', entity_name=entity_name)
+        view.yaml_dump.click()
+        output = view.browser.element(view.yaml_output).text
```
abalakh: why not `view.yaml_output.text`?

oshtaier: okay, logically they are the same, but in fact they did different actions for me. I am not sure that I want to debug that as it is not big deal
`view.browser.element(view.yaml_output).text` returned string with all `\n` symbols, but `view.browser.text(view.yaml_output)` returned without

oshtaier: nvm, let's me try to test `view.yaml_output.text`

airgun/views/smart_variable.py
oshtaier: hmm, as far as I tried and according to documentation - it will return set of classes, but I need simple string

airgun/entities/host.py
oshtaier: tried `output = view.yaml_output.text`. Also doesn't work for me and crash entire run

airgun/views/smart_variable.py
abalakh: and what's wrong with set (or list)? you'll have the same check - `if 'collapsed' in self.browser.classes`

oshtaier: ok, let's me test

PR #159, opened 2018-07-24
Fixed failures in case when the expected VM was on page > 1 of the VM table

lhellebr:


airgun/views/computeresource.py
```diff
@@ -146,3 +146,5 @@ class virtual_machines(SatTab):
                     'Power': Text('.//span[contains(@class,"label")]'),
                 }
         )
+        vm_search = TextInput(
```
oshtaier: Would `class virtual_machines(SatTab, SearchableViewMixin):` do the trick instead?

airgun/entities/computeresource.py
```diff
@@ -37,6 +37,7 @@ def delete(self, value):
     def list_vms(self, rhev_name, expected_vm_name=None):
         """Returns all the VMs on the CR or VM with specified name"""
         view = self.navigate_to(self, 'Detail', rhev_name=rhev_name)
+        view.virtual_machines.fill({'vm_search': expected_vm_name})
```
oshtaier: I have feeling even in old version two different functionalities mixed in one method: search and list_all. After that change it become even more obvious that we should distinguish these methods

airgun/views/computeresource.py
lhellebr: I tried it but it didn't work.

abalakh: I'd try to extend our search field's locator then, since it's yet-another-search-field-with-completely-the-same-functionality-but-different-unique-attr 🙄 
I mean smth like:
```python
class Search(Widget):
    search_field = TextInput(locator=(
        ".//input[@id='search' or @type='search' or "
        "@ng-model='table.searchTerm' or contains(@ng-model, 'Filter')]"))
```
Could you please try it out?

oshtaier:
seems as not actual anymore, at least in current implementation. Please, look and update it

oshtaier:
@lhellebr ping

lhellebr:
Pong. I'll look into it when I have time. Yes, it's entirely possible that after such a long time, the fix is not actual anymore.

oshtaier:
@abalakh @lhellebr ping.

lhellebr:
Rebased and fixed, please review

airgun/widgets.py
```diff
@@ -333,8 +333,8 @@ def read(self):
 
 class Search(Widget):
     search_field = TextInput(locator=(
-        ".//input[@id='search' or @ng-model='table.searchTerm' or "
-        "contains(@ng-model, 'Filter')]"))
+        ".//input[@id='search' or @type='search' or "
```
abalakh: Unfortunately, just found out `@type='search'` is not unique (at least on content view details page there're few such elements). Could you please try `@placeholder='Filter...'` instead?

lhellebr:
Fixed

airgun/entities/computeresource.py
```diff
@@ -52,6 +52,7 @@ def delete(self, value):
     def list_vms(self, rhev_name, expected_vm_name=None):
         """Returns all the VMs on the CR or VM with specified name"""
         view = self.navigate_to(self, 'Detail', rhev_name=rhev_name)
+        view.virtual_machines.search(expected_vm_name)
```
abalakh: `expected_vm_name` is optional, so forcing this search will lead to searching for 'None' each time `expected_vm_name` is not specified. And btw `return` can be simplified as well:
```python
if expected_vm_name:
    view.virtual_machines.search(expected_vm_name)
    return view.virtual_machines.table.row(name=expected_vm_name)
return view.virtual_machines.table.rows()
```

PR #161, opened 2018-07-26
Implemented Errata entity

### LINKED ISSUE #133: Views: Errata entity
Location:
Content -> Errata
(repo with errata is required to see some errata)

Views:
* [ ] Erratum list
* [ ] Errata Details - Details tab
* [ ] Errata Details - Content Hosts tab
* [ ] Errata Details - Repositories tab
* [ ] Errata - Apply errata

abalakh:
Closes #133

### DIFF
--- a/airgun/entities/errata.py
+++ b/airgun/entities/errata.py
@@ -1,0 +1,157 @@
+import re
+
+from airgun.entities.base import BaseEntity
+from airgun.navigation import NavigateStep, navigator
+from airgun.views.errata import (
+    ErratumView,
+    ErrataDetailsView,
+    ErrataTaskDetailsView,
+)
+
+
+class ErrataEntity(BaseEntity):
+
+    def search(self, value, applicable=True, installable=False, repo=None):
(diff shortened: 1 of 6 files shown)


airgun/entities/errata.py
```diff
@@ -0,0 +1,157 @@
...
+        applicable = kwargs.get('applicable')
+        installable = kwargs.get('installable')
+        repo = kwargs.get('repo')
+        self.parent.search(
+            entity_name,
+            applicable=applicable,
+            installable=installable,
+            repo=repo
+        )
+        row_filter = {'title': entity_name}
```
oshtaier: I feel some kind of redundancy here as your search already knew the criteria it searched with

abalakh: Yeah, but here i select different column depending on that criteria... Any better ideas? Theoretically we could try picking just first row available with no filtering, since searching was done and there shouldn't be lots of results, but i'm not sure that's a good idea.

PR #163, opened 2018-07-30
[WIP] Red Hat Subscriptions

mirzal:
I started working on Content -> Subscriptions, but it turned out to be more complex than I initially thought. Airgun side that covers adding and removing manifest is in this push request, but I would like to raise some questions before I move forward:

1. When deleting manifest, at one point there are three buttons labeled "Delete" on one page - one above the main table, one in Manage Manifest popup and one in final confirmation popup. It seems that `DeleteManifestConfirmation.delete_button` refers to some other button than the one it should, which is probably caused by my locators not being specific enough.
How can I ensure that correct button is selected? These popups don't have ids and all look about the same, with only difference being in title. My Xpath expressions start with dot, which is supposed to mean "current element" - but how it is selected? How can I debug how "current element" changes during execution of test and what it points to at any given time?

2. When adding or removing manifest, info flash popup and progress bar appear. My current approach  for waiting until operation finishes will fail, as info popup will be found and script will proceed. Can I somehow wait until element disappear? Or can I wait until popup message of given type / content appears?

3. Should I somehow handle "Entitlements" column in main table? It seems that old robottelo did not, so I can probably skip it. If I should cover it, is there anything in Airgun that I can reuse? That column is kind of special, as each cell only displays value, but changes itself into input on click.

4. When uploading manifest, I need to dismiss flash popups, as they cover UI elements that user might want to access (e.g. to remove manifest). How can I do it from robottelo test code? It seems that `view` is not exposed outside of Airgun. Or is dismissing them from Airgun OK?

5. I would welcome some suggestions what widgets I may use on Subscription details page (URL: `$HOST/subscriptions/$ID/`). This is static page with content that can only be read.

### DIFF
--- a/airgun/entities/subscription.py
+++ b/airgun/entities/subscription.py
@@ -1,16 +1,14 @@
-from selenium.common.exceptions import NoSuchElementException
-
 from navmazing import NavigateToSibling
 from wait_for import wait_for
 
 from airgun.entities.base import BaseEntity
 from airgun.navigation import NavigateStep, navigator
 from airgun.views.subscription import (
-    SubscriptionListView,
-    ManageManifestView,
-    DeleteManifestConfirmationView,
     AddSubscriptionView,
+    DeleteManifestConfirmationView,
+    ManageManifestView,
     SubscriptionDetailsView,
+    SubscriptionListView,
 )
 
 
@@ -42,18 +40,8 @@ class SubscriptionEntity(BaseEntity):
         })
         self._wait_for_process_to_finish(has_manifest=True)
 
-    def delete_manifest(self, really=True):
+    def delete_manifest(self):
         view = self.navigate_to(self, 'Delete Manifest Confirmation')
-        if not really:
-            view.cancel_button.click()
-            try:
-                while view.is_displayed:
-                    pass
-            except NoSuchElementException:
-                pass
-            self.navigate_to(self, 'Manage Manifest').close_button.click()
-            return
-
         view.delete_button.click()
         self._wait_for_process_to_finish(has_manifest=False)
 
@@ -73,22 +61,13 @@ class SubscriptionEntity(BaseEntity):
         return view.search(value)
 
     def provided_products(self, entity_name):
-        view = self.navigate_to(self, 'Details', subscription=entity_name)
-        return [elem.text
-                for elem
-                in view.details.provided_products.browser.elements("./li")]
+        view = self.navigate_to(self, 'Details', entity_name=entity_name)
+        return view.details.provided_products
 
     def enabled_products(self, entity_name):
-        locator = ("//div[contains(@class, 'list-group')]"
-                   "//div[contains(@class, 'list-group-item-heading')]"
-                   )
-        view = self.navigate_to(self, 'Details', subscription=entity_name)
+        view = self.navigate_to(self, 'Details', entity_name=entity_name)
         view.enabled_products.select()
-        try:
-            view.browser.wait_for_element(locator, visible=True)
-        except NoSuchElementException:
-            return []
-        return [elem.text for elem in view.browser.elements(locator)]
+        return view.enabled_products.enabled_products_list
 
     def update(self, entity_name, values):
         # This operation was never implemented in Robottelo (no test
@@ -134,9 +113,7 @@ class DeleteManifestConfirmation(NavigateStep):
 
     def step(self, *args, **kwargs):
         wait_for(
-                lambda: self.view.browser.get_attribute(
-                        "disabled",
-                        self.parent.manifest.delete_button) is None,
+                lambda: not self.parent.manifest.delete_button.disabled,
                 logger=self.view.logger,
                 timeout=10
         )
@@ -153,7 +130,8 @@ class AddSubscription(NavigateStep):
     prerequisite = NavigateToSibling('All')
 
     def am_i_here(self, *args, **kwargs):
-        return self.view.browser.selenium.current_url.endswith('/add')
+        return (self.view.is_displayed
+                and self.view.breadcrumb.locations[1] == 'Add Subscriptions')
 
     def step(self, *args, **kwargs):
         self.parent.add_button.click()
@@ -168,13 +146,9 @@ class SubscriptionDetails(NavigateStep):
 
     def am_i_here(self, *args, **kwargs):
         subscription_name = kwargs.get('subscription')
-        if not subscription_name:
-            raise ValueError('Invalid value of subscription parameter')
         return (self.view.is_displayed and
                 self.view.breadcrumb.read() == subscription_name)
 
     def step(self, *args, **kwargs):
-        subscription_name = kwargs.get('subscription')
-        if not subscription_name:
-            raise ValueError('Invalid value of subscription parameter')
+        subscription_name = kwargs.get('entity_name')
         self.parent.table.row(name=subscription_name)['Name'].widget.click()
--- a/airgun/entities/subscription.py
+++ b/airgun/entities/subscription.py
@@ -40,6 +40,7 @@ class SubscriptionEntity(BaseEntity):
 
     def add_manifest(self, manifest_file):
         view = self.navigate_to(self, 'Manage Manifest')
+        view.wait_animation_end()
         view.fill({
             'manifest.manifest_file': manifest_file,
         })
@@ -47,11 +48,13 @@ class SubscriptionEntity(BaseEntity):
 
     def delete_manifest(self):
         view = self.navigate_to(self, 'Delete Manifest Confirmation')
+        view.wait_animation_end()
         view.delete_button.click()
         self._wait_for_process_to_finish('Delete Manifest', has_manifest=False)
 
     def read_delete_manifest_message(self):
         view = self.navigate_to(self, 'Delete Manifest Confirmation')
+        view.wait_animation_end()
         return view.message.read()
 
     def add(self, entity_name, quantity=1):
@@ -108,10 +111,6 @@ class ManageManifest(NavigateStep):
     def step(self, *args, **kwargs):
         self.parent.manage_manifest_button.click()
 
-    def post_navigate(self, _tries, *args, **kwargs):
-        self.view.wait_animation_end()
-        return self.view.is_displayed
-
 
 @navigator.register(SubscriptionEntity, 'Delete Manifest Confirmation')
 class DeleteManifestConfirmation(NavigateStep):
@@ -125,10 +124,6 @@ class DeleteManifestConfirmation(NavigateStep):
                 handle_exception=True, logger=self.view.logger, timeout=10
         )
         self.parent.manifest.delete_button.click()
-
-    def post_navigate(self, _tries, *args, **kwargs):
-        self.view.wait_animation_end()
-        return self.view.is_displayed
 
 
 @navigator.register(SubscriptionEntity, 'Add')
--- a/airgun/entities/subscription.py
+++ b/airgun/entities/subscription.py
@@ -29,7 +29,7 @@ class SubscriptionEntity(BaseEntity):
         wait_for(
                 lambda: self.has_manifest == has_manifest,
                 handle_exception=True, timeout=10,
-                logger=view.progressbar.logger
+                logger=view.logger
         )
         view.flash.dismiss()
 
@@ -97,6 +97,13 @@ class SubscriptionEntity(BaseEntity):
 @navigator.register(SubscriptionEntity, 'All')
 class SubscriptionList(NavigateStep):
     VIEW = SubscriptionListView
+
+    def pre_navigate(self, _tries, *args, **kwargs):
+        super(SubscriptionList, self).pre_navigate(_tries, *args, **kwargs)
+        wait_for(
+                lambda: not self.view.fake_fade_widget.is_displayed,
+                handle_exception=True, logger=self.view.logger, timeout=10 * 60
+        )
 
     def step(self, *args, **kwargs):
         self.view.menu.select('Content', 'Subscriptions')
--- a/airgun/entities/subscription.py
+++ b/airgun/entities/subscription.py
@@ -72,7 +72,7 @@ class SubscriptionEntity(BaseEntity):
 
     def provided_products(self, entity_name):
         view = self.navigate_to(self, 'Details', entity_name=entity_name)
-        return view.details.provided_products
+        return view.details.provided_products.read()
 
     def enabled_products(self, entity_name):
         view = self.navigate_to(self, 'Details', entity_name=entity_name)
--- a/airgun/entities/subscription.py
+++ b/airgun/entities/subscription.py
@@ -1,3 +1,4 @@
+from selenium.common.exceptions import NoSuchElementException
 from navmazing import NavigateToSibling
 from wait_for import wait_for
 
@@ -62,8 +63,8 @@ class SubscriptionEntity(BaseEntity):
         for row in view.table.rows(subscription_name=entity_name):
             row['Quantity to Allocate'].fill(quantity)
         view.submit_button.click()
-        self._wait_for_process_to_finish('Bind entitlements to an allocation',
-                                         has_manifest=True)
+        self._wait_for_process_to_finish(
+                'Bind entitlements to an allocation', has_manifest=True)
 
     def search(self, value):
         view = self.navigate_to(self, 'All')
@@ -76,7 +77,10 @@ class SubscriptionEntity(BaseEntity):
     def enabled_products(self, entity_name):
         view = self.navigate_to(self, 'Details', entity_name=entity_name)
         view.enabled_products.select()
-        return view.enabled_products.enabled_products_list
+        try:
+            return view.enabled_products.enabled_products_list
+        except NoSuchElementException:
+            return []
 
     def update(self, entity_name, values):
         # This operation was never implemented in Robottelo (no test
@@ -90,8 +94,8 @@ class SubscriptionEntity(BaseEntity):
             row['Select all rows'].fill(True)
         view.delete_button.click()
         view.confirm_deletion.confirm()
-        self._wait_for_process_to_finish('Delete Upstream Subscription',
-                                         has_manifest=True)
+        self._wait_for_process_to_finish(
+                'Delete Upstream Subscription', has_manifest=True)
 
 
 @navigator.register(SubscriptionEntity, 'All')
--- a/airgun/entities/subscription.py
+++ b/airgun/entities/subscription.py
@@ -31,7 +31,7 @@ class SubscriptionEntity(BaseEntity):
     @property
     def has_manifest(self):
         view = self.navigate_to(self, 'All')
-        return "disabled" not in view.browser.classes(view.add_button)
+        return not view.add_button.disabled
 
     def add_manifest(self, manifest_file):
         view = self.navigate_to(self, 'Manage Manifest')
@@ -114,8 +114,7 @@ class DeleteManifestConfirmation(NavigateStep):
     def step(self, *args, **kwargs):
         wait_for(
                 lambda: not self.parent.manifest.delete_button.disabled,
-                logger=self.view.logger,
-                timeout=10
+                handle_exception=True, logger=self.view.logger, timeout=10
         )
         self.parent.manifest.delete_button.click()
 
--- a/airgun/entities/subscription.py
+++ b/airgun/entities/subscription.py
@@ -13,17 +13,22 @@ from airgun.views.subscription import (
 
 
 class SubscriptionEntity(BaseEntity):
-    def _wait_for_process_to_finish(self, has_manifest=False):
+    def _wait_for_process_to_finish(self, name, has_manifest=False):
         view = self.navigate_to(self, 'All')
-        view.progressbar.wait_displayed(timeout=20)
+        wait_for(
+                lambda: view.flash.assert_message(
+                        "Task {} completed".format(name), partial=True),
+                handle_exception=True, timeout=60*10,
+                logger=view.flash.logger
+        )
         wait_for(
                 lambda: not view.progressbar.is_displayed,
-                timeout=60*10,
+                handle_exception=True, timeout=60*10,
                 logger=view.progressbar.logger
         )
         wait_for(
                 lambda: self.has_manifest == has_manifest,
-                timeout=10,
+                handle_exception=True, timeout=10,
                 logger=view.progressbar.logger
         )
         view.flash.dismiss()
@@ -38,12 +43,12 @@ class SubscriptionEntity(BaseEntity):
         view.fill({
             'manifest.manifest_file': manifest_file,
         })
-        self._wait_for_process_to_finish(has_manifest=True)
+        self._wait_for_process_to_finish('Import Manifest', has_manifest=True)
 
     def delete_manifest(self):
         view = self.navigate_to(self, 'Delete Manifest Confirmation')
         view.delete_button.click()
-        self._wait_for_process_to_finish(has_manifest=False)
+        self._wait_for_process_to_finish('Delete Manifest', has_manifest=False)
 
     def read_delete_manifest_message(self):
         view = self.navigate_to(self, 'Delete Manifest Confirmation')
@@ -54,7 +59,8 @@ class SubscriptionEntity(BaseEntity):
         for row in view.table.rows(subscription_name=entity_name):
             row['Quantity to Allocate'].fill(quantity)
         view.submit_button.click()
-        self._wait_for_process_to_finish(has_manifest=True)
+        self._wait_for_process_to_finish('Bind entitlements to an allocation',
+                                         has_manifest=True)
 
     def search(self, value):
         view = self.navigate_to(self, 'All')
@@ -81,7 +87,8 @@ class SubscriptionEntity(BaseEntity):
             row['Select all rows'].fill(True)
         view.delete_button.click()
         view.confirm_deletion.confirm()
-        self._wait_for_process_to_finish(has_manifest=True)
+        self._wait_for_process_to_finish('Delete Upstream Subscription',
+                                         has_manifest=True)
 
 
 @navigator.register(SubscriptionEntity, 'All')
--- a/airgun/entities/subscription.py
+++ b/airgun/entities/subscription.py
@@ -1,0 +1,180 @@
+from selenium.common.exceptions import NoSuchElementException
+
+from navmazing import NavigateToSibling
+from wait_for import wait_for, TimedOutError
+
+from airgun.entities.base import BaseEntity
+from airgun.navigation import NavigateStep, navigator
+from airgun.views.subscription import (
+    SubscriptionListView,
+    ManageManifestView,
+    DeleteManifestConfirmationView,
+    AddSubscriptionView,
+    SubscriptionDetailsView,
+)
+
+
+class SubscriptionEntity(BaseEntity):
+    def _wait_for_process_to_finish(self, has_manifest=False):
+        view = self.navigate_to(self, 'All')
+        view.progressbar.wait_displayed(timeout=20)
+        wait_for(
+                lambda: not view.progressbar.is_displayed,
+                timeout=60*10,
+                logger=view.progressbar.logger
+        )
+        wait_for(
+                lambda: self.has_manifest == has_manifest,
+                timeout=10,
+                logger=view.progressbar.logger
+        )
+        view.flash.dismiss()
+
+    @property
+    def has_manifest(self):
+        view = self.navigate_to(self, 'All')
+        return "disabled" not in view.browser.classes(view.add_button)
+
+    def add_manifest(self, manifest_file):
+        view = self.navigate_to(self, 'Manage Manifest')
+        view.fill({
+            'manifest.manifest_file': manifest_file,
+        })
+        self._wait_for_process_to_finish(has_manifest=True)
+
+    def delete_manifest(self, really=True):
+        view = self.navigate_to(self, 'Delete Manifest Confirmation')
+        if not really:
+            view.cancel_button.click()
+            try:
+                while view.is_displayed:
+                    pass
+            except NoSuchElementException:
+                pass
+            self.navigate_to(self, 'Manage Manifest').close_button.click()
+            return
+
+        view.delete_button.click()
+        self._wait_for_process_to_finish(has_manifest=False)
+
+    def read_delete_manifest_message(self):
+        view = self.navigate_to(self, 'Delete Manifest Confirmation')
+        return view.message.read()
+
+    def add(self, entity_name, quantity=1):
+        view = self.navigate_to(self, 'Add')
+        for row in view.table.rows(subscription_name=entity_name):
+            row['Quantity to Allocate'].fill(quantity)
+        view.submit_button.click()
+        self._wait_for_process_to_finish(has_manifest=True)
+
+    def search(self, value):
+        view = self.navigate_to(self, 'All')
+        return view.search(value)
+
+    def provided_products(self, entity_name):
+        view = self.navigate_to(self, 'Details', subscription=entity_name)
+        return [elem.text
+                for elem
+                in view.details.provided_products.browser.elements("./li")]
+
+    def enabled_products(self, entity_name):
+        locator = ("//div[contains(@class, 'list-group')]"
+                   "//div[contains(@class, 'list-group-item-heading')]"
+                   )
+        view = self.navigate_to(self, 'Details', subscription=entity_name)
+        view.enabled_products.select()
+        try:
+            view.browser.wait_for_element(locator, visible=True)
+        except NoSuchElementException:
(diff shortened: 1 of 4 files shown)


oshtaier:
>>When uploading manifest, I need to dismiss flash popups, as they cover UI elements that user might want to access (e.g. to remove manifest). How can I do it from robottelo test code? It seems that view is not exposed outside of Airgun. Or is dismissing them from Airgun OK?

Right, view is not exposed outside airgun code. And correct, we usually dismiss `flash` messages on entity level

oshtaier:
>5) I would welcome some suggestions what widgets I may use on Subscription details page (URL: $HOST/subscriptions/$ID/). This is static page with content that can only be read.

`SatTableWithUnevenStructure(locator="//table[@class='table']")`
Already tested out:
```
{'Name': 'MdvPSbRYPDsyTxknNSNu', 'Description': '', 'Requires Virt-Who': 'false', 'Consumed': '0', 'Starts': '2018-07-25 19:48:14 +0300', 'Ends': '2048-07-17 18:48:14 +0200', 'Product ID': '423109244712', 'Contract Number': '', 'Account Number': '', 'Support Level': '', 'Support Type': '', 'Architecture(s)': 'ALL', 'Type': 'NORMAL', 'Multi-entitlement': '', 'Stacking ID': '', 'Limits': '', 'Instance-based': 'Yes'}
```
Same widget can also be used for `Associations` as it was created specifically for such tables

oshtaier:
3) Do you talk about that?
image

It does nothing for me on click

But in case you faced such behavior then you need to create your own widget and re-use it in same we do all other widgets:
```
table = SatTable(
    './/table',
    column_widgets={...'Entitlement': Your_widget('locator'),}
```

oshtaier:
2) It depends on specific situation, but on the idea that can be used:
```
wait_for(
            lambda: element.is_displayed is False,
            timeout=timeout,
            delay=delay,
            logger=self.logger
        )
```

oshtaier:
1) You are on right way. `.` is a proper way to distinguish current context. `.` in other words mean ROOT for current widget.
Briefly go through `ROOT` usage here: https://github.com/ManageIQ/integration_tests/blob/master/widgetastic_manageiq/__init__.py
That was a place for me to understand the idea

mirzal:
@oshtaier , @abalakh I'll have to focus on other tasks for next couple of days, but this PR has been updated with what I have so far, so we can start review process.

This mostly works, although there are few issues:

1. For some reason, `step` in `SubscriptionDetails` doesn't work - it seems that cell in table is recognized correctly, debug log shows that something was clicked, but Subscription Info page is never opened. Help in debugging this is very welcome!
2. `SubscriptionEntity.search()` will throw exception when searching for something that is not in table. The root cause is that Subscriptions page uses something that looks like our `SatTable`, but is actually not. E.g. SatTable displays only header when nothing matches search criteria, but that thing displays one row that spans all columns. Maybe these differences are also root cause for issue #1 above. Pointers on how to handle it are welcome!
3. In Robottelo there were three methods that are not ported - `get_content_products`, `get_guests_provided_products` and `get_guests_content_products`. It seems that content product was dropped in 6.4 - compare [Sat6.3](https://user-images.githubusercontent.com/12373754/43835216-67060fe2-9b11-11e8-8f9b-a6756823bf57.png) with [Sat6.4](https://user-images.githubusercontent.com/12373754/43835223-6abd63ba-9b11-11e8-9ee1-0d94f94a1a46.png). I think the same happened to guests products, but it might be that this data is hidden until there actually are some guests products and I don't know how to add them.

airgun/views/subscription.py
```diff
@@ -0,0 +1,102 @@
...
+from airgun.views.common import BaseLoggedInView, SearchableViewMixin, SatTab
+from airgun.widgets import DeleteSubscriptionConfirmationDialog, SatTable, ProgressBar
+
+
+class SubscriptionListView(BaseLoggedInView, SearchableViewMixin):
+    """List of all subscriptions."""
+    table = SatTable(
+        locator='//*[@id="subscriptions-table"]//table',
+        column_widgets={
+            'Select all rows': Checkbox(locator=".//input[@type='checkbox']"),
```
oshtaier: not sure that it work in that way:
`0: Checkbox(locator=".//input[@type='checkbox']"),`

airgun/views/subscription.py
```diff
@@ -0,0 +1,102 @@
...
+            'Name': Text(".//a"),
```
oshtaier: try:
`'Name': Text("./a")`

oshtaier:
2) Funny that you described exactly opposite situation to what SatTable is doing. It was designed specifically for `one row that spans all columns` situations. Read more accurately its code. The problem 
that generic case is:
```
no_rows_message = (
        ".//td/span[contains(@data-block, 'no-rows-message') or "
        "contains(@data-block, 'no-search-results-message')]")
```
but in your case:
```
<td colspan="9">No subscriptions match your search criteria.</td>
```
I don't think that we like to expand generic class with exceptions like this, so probably just create your own class for your entity:
```
class RHSubTable(SatTable)
no_rows_message  = ...
```

oshtaier:
3) I am not familiar with the logic behind that. Maybe you can ask someone from the team who works with 6.4 at that moment

airgun/entities/subscription.py
```diff
@@ -0,0 +1,180 @@
...
+    def add_manifest(self, manifest_file):
+        view = self.navigate_to(self, 'Manage Manifest')
+        view.fill({
+            'manifest.manifest_file': manifest_file,
+        })
+        self._wait_for_process_to_finish(has_manifest=True)
+
+    def delete_manifest(self, really=True):
+        view = self.navigate_to(self, 'Delete Manifest Confirmation')
+        if not really:
```
oshtaier: please remove:
```
        if not really:
            view.cancel_button.click()
            try:
                while view.is_displayed:
                    pass
            except NoSuchElementException:
                pass
            self.navigate_to(self, 'Manage Manifest').close_button.click()
            return
```
as we will never verify that functionality for tier2-tier4 tests

airgun/entities/subscription.py
```diff
@@ -0,0 +1,180 @@
...
+
+    def search(self, value):
+        view = self.navigate_to(self, 'All')
+        return view.search(value)
+
+    def provided_products(self, entity_name):
+        view = self.navigate_to(self, 'Details', subscription=entity_name)
+        return [elem.text
+                for elem
+                in view.details.provided_products.browser.elements("./li")]
```
oshtaier: please always make locators as class attributes

airgun/entities/subscription.py
```diff
@@ -0,0 +1,180 @@
...
+
+    def enabled_products(self, entity_name):
+        locator = ("//div[contains(@class, 'list-group')]"
+                   "//div[contains(@class, 'list-group-item-heading')]"
+                   )
+        view = self.navigate_to(self, 'Details', subscription=entity_name)
+        view.enabled_products.select()
+        try:
```
oshtaier: any specific reason to have additional exception handling here? `NoSuchElementException` sounds pretty good

airgun/entities/subscription.py
```diff
@@ -0,0 +1,180 @@
...
+
+
+@navigator.register(SubscriptionEntity, 'Add')
+class AddSubscription(NavigateStep):
+    VIEW = AddSubscriptionView
+
+    prerequisite = NavigateToSibling('All')
+
+    def am_i_here(self, *args, **kwargs):
+        return self.view.browser.selenium.current_url.endswith('/add')
```
oshtaier: that is a really not good criteria. Can you refer to elements on a page?

airgun/entities/subscription.py
```diff
@@ -0,0 +1,180 @@
...
+    VIEW = SubscriptionDetailsView
+
+    def prerequisite(self, *args, **kwargs):
+        return self.navigate_to(self.obj, 'All')
+
+    def am_i_here(self, *args, **kwargs):
+        subscription_name = kwargs.get('subscription')
+        if not subscription_name:
+            raise ValueError('Invalid value of subscription parameter')
+        return (self.view.is_displayed and
```
oshtaier: does it suffice your needs https://github.com/SatelliteQE/airgun/blob/master/airgun/navigation.py#L73?

airgun/entities/subscription.py
```diff
@@ -0,0 +1,180 @@
...
+        return view.search(value)
+
+    def provided_products(self, entity_name):
+        view = self.navigate_to(self, 'Details', subscription=entity_name)
+        return [elem.text
+                for elem
+                in view.details.provided_products.browser.elements("./li")]
+
+    def enabled_products(self, entity_name):
+        locator = ("//div[contains(@class, 'list-group')]"
```
oshtaier: we cannot work with locators on entity layer, please move to view class

airgun/entities/subscription.py
```diff
@@ -0,0 +1,180 @@
...
+
+
+@navigator.register(SubscriptionEntity, 'Delete Manifest Confirmation')
+class DeleteManifestConfirmation(NavigateStep):
+    VIEW = DeleteManifestConfirmationView
+
+    prerequisite = NavigateToSibling('Manage Manifest')
+
+    def step(self, *args, **kwargs):
+        wait_for(
```
oshtaier: so you need to find out whether element is disabled? if yes - then probably we will need implement wrapper for selenium method: element.is_enabled()

airgun/entities/subscription.py
```diff
@@ -0,0 +1,180 @@
...
+                lambda: self.view.browser.get_attribute(
+                        "disabled",
+                        self.parent.manifest.delete_button) is None,
+                logger=self.view.logger,
+                timeout=10
+        )
+        self.parent.manifest.delete_button.click()
+
+    def post_navigate(self, _tries, *args, **kwargs):
+        return self.view.is_displayed
```
oshtaier: why do you need to do that in post_navigation? That step is supposed to be performed before navigation, not after.
just remove it as it is wrong for any possible situation, if 3 attempts not enough for you - think about other options

airgun/entities/subscription.py
```diff
@@ -0,0 +1,180 @@
...
+@navigator.register(SubscriptionEntity, 'Details')
+class SubscriptionDetails(NavigateStep):
+    VIEW = SubscriptionDetailsView
+
+    def prerequisite(self, *args, **kwargs):
+        return self.navigate_to(self.obj, 'All')
+
+    def am_i_here(self, *args, **kwargs):
+        subscription_name = kwargs.get('subscription')
+        if not subscription_name:
```
oshtaier: that validation is redundant as it is just doesn't make sense (probably created for dumb situation like `provided_products( entity_name=None)`)

airgun/entities/subscription.py
```diff
@@ -0,0 +1,180 @@
...
+            raise ValueError('Invalid value of subscription parameter')
+        return (self.view.is_displayed and
+                self.view.breadcrumb.read() == subscription_name)
+
+    def step(self, *args, **kwargs):
+        subscription_name = kwargs.get('subscription')
```
oshtaier: same here

airgun/entities/subscription.py
```diff
@@ -0,0 +1,180 @@
...
+        if not subscription_name:
```
oshtaier: same here

airgun/widgets.py
```diff
@@ -334,10 +334,12 @@ def read(self):
 class Search(Widget):
     search_field = TextInput(locator=(
         ".//input[@id='search' or @ng-model='table.searchTerm' or "
-        "contains(@ng-model, 'Filter')]"))
+        "contains(@ng-model, 'Filter') or "
+        "starts-with(@id, 'downshift-')]"))
```
oshtaier: that is not good. If you have specific locator, so override that class for your view. Never expand most fundamental classes in framework like this

airgun/widgets.py
```diff
@@ -334,10 +334,12 @@ def read(self):
...
     search_button = Text(
         ".//button[contains(@type,'submit') or "
-        "@ng-click='table.search(table.searchTerm)']"
+        "@ng-click='table.search(table.searchTerm)' or "
```
oshtaier: same here

airgun/widgets.py
```diff
@@ -826,6 +828,11 @@ def read(self):
         do_not_read_this_widget()
 
 
+class DeleteSubscriptionConfirmationDialog(ConfirmationDialog):
```
oshtaier: put that class in your view module.

airgun/widgets.py
```diff
@@ -1240,6 +1247,18 @@ def read(self):
         return read_rows
 
 
+class SatSubscriptionsViewTable(SatTable):
```
oshtaier: sounds as one that can be moved to your view module too

airgun/widgets.py
```diff
@@ -1240,6 +1247,18 @@ def read(self):
...
+    """Table used on Red Hat Subscriptions page. It's mostly the same as
+    normal table, but when search returns no results, it does display single
+    row with message.
+    Not to be confused with SatSubscriptionsTable, which is not used on
+    that page
+    """
+    @property
+    def has_rows(self):
+        return self.tbody_row.read() != "No subscriptions match your search criteria."  # noqa: E501
```
oshtaier:
```
return (
    self.tbody_row.read()
    != "No subscriptions match your search criteria."
)

File: airgun/entities/subscription.py [hunk 19]
```diff
@@ -0,0 +1,180 @@
+from selenium.common.exceptions import NoSuchElementException
+
+from navmazing import NavigateToSibling
+from wait_for import wait_for
+
+from airgun.entities.base import BaseEntity
+from airgun.navigation import NavigateStep, navigator
+from airgun.views.subscription import (
+    SubscriptionListView,
```
oshtaier: sort alphabetically

airgun/entities/subscription.py
mirzal: Yes, I want Airgun method to always return list - even if it is empty list. Methods like `search` already return empty list when no elements are found and I wanted to be consistent.

airgun/entities/subscription.py
mirzal: I tried to use it, but unfortunately it does not cut in. In relevant Robottelo test ([source](https://github.com/SatelliteQE/robottelo/pull/6265/files#diff-9bef302b55c1860e808fb7b1a9bf8094R136)) I first create product. The way Satellite display breadcrumbs there will fool this method into thinking that detailed product page is Subscription detail page (because both happen to have the same string as second breadcrumb item).

airgun/entities/subscription.py
mirzal: I started using Button from widgetastic_patternfly, which already provides `disabled` property. Hopefully it's OK now.

airgun/entities/subscription.py
mirzal: I've seen it used already in location and organization views and thought that idea is not terrible, but have now removed it.

oshtaier:
Thanks for the information
I will think more about comments
Meanwhile, ping @abalakh hey bro, check last comment and PR in general

airgun/views/subscription.py
```diff
@@ -0,0 +1,117 @@
...
+    confirm_deletion = DeleteSubscriptionConfirmationDialog()
+
+    @property
+    def is_displayed(self):
+        return self.browser.wait_for_element(
+            'div#subscriptions-table', timeout=10, exception=False) is not None
+
+
+class ManageManifestView(BaseLoggedInView):
+    _locator = '//h4[@class="modal-title"][text()="Manage Manifest"]'
```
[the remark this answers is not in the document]
oshtaier:
```
ROOT = '//div[@class="modal-content"][div/h4[text()="Manage Manifest"]]'
```
sounds better

airgun/views/subscription.py
```diff
@@ -0,0 +1,117 @@
...
+        )
+
+    @property
+    def is_displayed(self):
+        return self.browser.wait_for_element(
+            self._locator, visible=True, exception=False) is not None
+
+
+class DeleteManifestConfirmationView(BaseLoggedInView):
+    _locator = '//h4[@class="modal-title"][text()="Confirm delete manifest"]'
```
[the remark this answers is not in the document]
oshtaier: ROOT = `//div[@class="modal-content"][div/h4[text()="Confirm delete manifest"]]`

mirzal:
> `self.view.is_displayed` doesn't wait for view to be displayed

Doesn't it?

`post_navigate` is used for `ManageManifestView` and `DeleteManifestConfirmation`. Both of them have `is_displayed` following template:

```
def is_displayed(self):
        self.delete_button, visible=True, exception=False) is not None
```

`wait_for_element` calls `wait_for` which - as I understand it - calls function repeatedly until it completes successfully, or some timeout is reached. In other words, `wait_for_element` will return falsey value after timeout (5 seconds) or True sometime earlier. The same is true for `is_displayed`.

If that helps, I can call `wait_for_element` or `wait_for` explicitly in `post_navigate`.

> In any case - that's not how we should handle ajax, if there's smth running in the background - we should write proper js and update ensure_page_safe

There are no requests here.

I observed this problem when going Content -> Subscriptions -> Manage Manifest -> Delete. Sometimes final confirmation popup would be displayed, but Selenium complained that it failed to click element. Confirmation popup appears with short animation and I figured that Selenium tries to click button to soon - while animation is still running. Since addition of `post_navigate` seems to solve the problem, I take it as confirmation of my assumptions.

Possibly it could make sense to verify in ENSURE_PAGE_SAFE that no animations are currently running on page, assuming there is unified animations solution in Satellite and no animation is legitimate part of page (i.e. any animation always means that page is not yet ready). But even then, it would require extensive regression testing. I don't really want to do that as part of this PR.

Another solution could be to model these popups as widgets on 'All' page instead of views representing separate pages. But I don't have good idea on how to approach creating them, especially the part when widget ensures it is visible before actually interacting with child widgets (so when I run `NEW_WIDGET.delete_button.click()`, NEW_WIDGET will hold clicking until it itself is visible).

> Another vulnerable place is your view.progressbar.wait_displayed(timeout=20) line - what if process finishes before we navigate to All page and check for progressbar presence?

Well, when we confirm change of subscription (add/delete manifest or subscription), Satellite is already at 'All' page, so the process would have to finish in the time we run `am_i_here` for view. Since right now it takes at least 40 seconds for Satellite to finish the task, I don't think this will be practical concern anytime soon, if ever.

Either way, I will try to use flash messages as indication for success. Thanks for suggestion.

abalakh:
@mirzal re `is_displayed` - my bad, you're right.

>Confirmation popup appears with short animation and I figured that Selenium tries to click button to soon - while animation is still running.

How about finding the element which represents that animation and add waiter for that animation to disappear?

abalakh:
>I can extend ENSURE_SAFE_PAGE to add check that would pass when there are no elements with fade class or when all elements with fade class also have in class. But is this really better solution than current post_navigate? Especially considering that it will affect virtually every page and button click and will require time-consuming performance and regression testing?

Of course it's not better solution, ENSURE_SAFE_PAGE should be used for scripts affecting all (or at least a lot of) pages, not one specific. I was talking about creating a waiter for that specific page - separate method for that view, which uses wait_for to ensure that `in` appears in div. Then you can use that method in you entity helpers or even in post_navigate if needed. My concern wasn't about `post_navigate` use, but about improper waiter inside of it - it's cool coincidence that waiting for view to be displayed works for you on your env, but we can't rely on such coincidence and should wait for animation to disappear.

oshtaier:
We went manually through the steps and now see your points more clearly

Let's introduce small fixes to the code and we will do final review.

```
Manage

    def post_navigate(self, _tries, *args, **kwargs):
        return self.view.is_displayed

    @property
    def is_displayed(self):
            self.close_button, visible=True, exception=False) is not None
***
	and .class contains 'in'
***

Delete

    def post_navigate(self, _tries, *args, **kwargs):
        return self.view.is_displayed


    @property
    def is_displayed(self):
            self.close_button, visible=True, exception=False) is not None
***
	and .class contains 'in'
***


All:

@navigator.register(SubscriptionEntity, 'All')
class SubscriptionList(NavigateStep):
    VIEW = SubscriptionListView

***
pre-navigate:
wait for element with `fade` is not visible
***

        self.view.menu.select('Content', 'Subscriptions')
```

oshtaier:
That's fine to not move message functionality on a higher level as it is out of scope for current PR.

Just create issue for airgun and work on it later. If you like, of course

mirzal:
@oshtaier @abalakh Suggestions included in newest PR version. In d0f6e62c71958b7e74891f53e916480b6eedf92e I call `wait_animation_end()` in entity methods instead of `post_navigate`.

> `class SubscriptionList(NavigateStep):`
> pre-navigate:
> wait for element with `fade` is not visible


Turns out it's not that easy.

Only after adding such change I noticed that progress bar is actually placed inside dialog box. As long as add/delete manifest/subscription task is running, there will be element with "fade" class on page. The same timeout is shared in two cases: when task is running (after closing dialog windows) and when we navigate to page from main menu. If I add big timeout (allow task to finish), we will waste time when trying to navigate from menu. If I add low timeout (quick navigation from menu), tests will fail - timeout will be reached before task is finished.

There are couple of ways we can try to work around that, but neither is particularly nice (low timeout and ignore exception; pass timeout parameter somehow, so timeout is longer when we know task should be running; change timeout depending on what page we currently are). 

I think that waiting for element with `fade` class to disappear is needed only for dialog windows, but as progress bar is inside dialog, I already cover that case by waiting for progress bar to disappear. Or am I missing something?

oshtaier:
You missed the case when you are on main page and any dialog is present on a page due any reason.
Navigation will check
```
    @property
    def is_displayed(self):
            'div#subscriptions-table', timeout=10, exception=False) is not None
```
yes, element is there, but not clickable, because the whole page is blocked by dialog box.
Not sure, what `If I add big timeout (allow task to finish), we will waste time when trying to navigate from menu.` means. Why do you waste time on navigation? Do you have situation when you need to have navigation before task is finished or dialog box is closed?

mirzal:
@oshtaier 

In a1ae1cb5edb866b039b849e2813649a9e3aa59c4 I added `pre_navigation` that you requested before. I thought that this will create unnecessary wait when navigating to Subscription page from main menu, but I was mistaken. Adding this seems to be free of negative side-effects.

I think that addresses all your concerns?

airgun/views/subscription.py
```diff
@@ -65,6 +65,9 @@ class SubscriptionListView(BaseLoggedInView, SubscriptionSearchableViewMixin):
     delete_button = Button('Delete')
     progressbar = ProgressBar('//div[contains(@class,"progress-bar-striped")]')
     confirm_deletion = DeleteSubscriptionConfirmationDialog()
+    # In pre_navigate we wait for element with class `fade` to be not
+    # visible; we need to first define it here
+    fake_fade_widget = Text(".//*[contains(@class, 'fade')]")
```
oshtaier: if that is div and I believe it is div, so use it. Avoid `*` usage if possible

airgun/entities/subscription.py
```diff
@@ -0,0 +1,164 @@
...
+        )
+        wait_for(
+                lambda: self.has_manifest == has_manifest,
+                handle_exception=True, timeout=10,
+                logger=view.logger
+        )
+        view.flash.dismiss()
+
+    @property
+    def has_manifest(self):
```
oshtaier: why do we need that property? Looks like redundant navigation as we can have `lambda: not view.add_button.disabled == has_manifest`

airgun/entities/subscription.py
```diff
@@ -0,0 +1,164 @@
...
+        # exception until it is actually needed
+        raise NotImplementedError("Subscriptions update is not implemented")
+
+    def delete(self, entity_name):
+        view = self.navigate_to(self, 'All')
+        for row in view.table.rows(name=entity_name):
+            row['Select all rows'].fill(True)
+        view.delete_button.click()
+        view.confirm_deletion.confirm()
+        self._wait_for_process_to_finish('Delete Upstream Subscription',
```
oshtaier:
```
    'Delete Upstream Subscription', has_manifest=True)
```

airgun/entities/subscription.py
```diff
@@ -0,0 +1,164 @@
...
+        view = self.navigate_to(self, 'Delete Manifest Confirmation')
+        view.wait_animation_end()
+        return view.message.read()
+
+    def add(self, entity_name, quantity=1):
+        view = self.navigate_to(self, 'Add')
+        for row in view.table.rows(subscription_name=entity_name):
+            row['Quantity to Allocate'].fill(quantity)
+        view.submit_button.click()
+        self._wait_for_process_to_finish('Bind entitlements to an allocation',
```
oshtaier:
```
    'Bind entitlements to an allocation', has_manifest=True)
```

airgun/views/subscription.py
```diff
@@ -0,0 +1,185 @@
...
+    def has_rows(self):
+        return (self.tbody_row.read()
+                != "No subscriptions match your search criteria."
+                )
+
+
+class SubscriptionListView(BaseLoggedInView, SubscriptionSearchableViewMixin):
+    """List of all subscriptions."""
+    table = SatSubscriptionsViewTable(
+        locator='//*[@id="subscriptions-table"]//table',
```
oshtaier: change `//*` on corresponding element

airgun/views/subscription.py
```diff
@@ -0,0 +1,185 @@
...
+    class enabled_products(SatTab):
+        TAB_NAME = "Enabled Products"
+
+        @property
+        def enabled_products_list(self):
+            locator = ("//div[contains(@class, 'list-group')]"
+                       "//div[contains(@class, 'list-group-item-heading')]"
+                       )
+            try:
+                self.browser.wait_for_element(locator, visible=True)
```
oshtaier: I feel something wrong on that logic. That level should not be ambiguous and raise exception in case element is not found. Otherwise you will wait here for 12 seconds for no reason

oshtaier:
pending few nitpicks and few question.
We almost done =) thanks for patience

oshtaier:
@abalakh please review once more whole PR

mirzal:
@oshtaier Most comments resolved in 4b284b509a530706359c91c962e800a395b28f1e

`has_manifest` property exists to allow Robottelo to probe into state of application. On test side we want to upload manifest and assert that it is available afterwards. Currently we always run this assertion after `_wait_for_process_to_finish`, which pretty much guarantees result, but it's mere coincidence. And even then, without this property, we won't have anything to run assertions against for upload / delete manifest scenarios.

Navigation is needed because property might be read on Robottelo side while Satellite is on completely different page. Think about scenario like: create organization, upload manifest, delete organization, create new organization with the same name, verify that it does not have manifest.

-----

As for `try` inside view:

I have moved try / except into entity. I still catch exception on Airgun side, because I want `enabled_products` to return list to Robottelo even if there are no products enabled.

However, I can't remove wait. The problem is that when you activate tab, message about no enabled products is displayed each and every time. Only then React looks into internal store (there is no web request involved) and re-constructs page with data. Time needed for update varies, but might take as much as 500ms-1s. In other words - when I activate tab and see that there are no enabled products, it might just mean that page hasn't been updated yet. I have to give browser some time to update content and only then return result.

This is yet another case of making things more complex due to React...

airgun/views/subscription.py
```diff
@@ -0,0 +1,181 @@
...
+    @View.nested
+    class details(SatTab):
+
+        @property
+        def provided_products(self):
+            return [elem.text
+                    for elem
+                    in self.browser.elements(
+                        ("//h2[text()='Provided Products']"
+                         "/following::ul/li"))]
```
abalakh: Let's not cheat with properties and use widgets instead, `ItemsList` from airgun's widgets should work here

mirzal:
@abalakh the whole point is that there is no AJAX request that you could track.

Added docstrings and switched to `ItemsListReadOnly` widget. I was looking for something like that before rolling out my own solution, but somehow missed this one. Thanks!

mirzal:
New widget `EnabledProductsItemsList` added in cf72f88aa6ffc1a05bcb646f3c5080844bf731f2. It's in view, as this is the only place where it is used.

Let me know if there is anything else to improve.

airgun/views/subscription.py
mirzal: Sorry, missed that one.
Unfortunately, it won't work. If I use numbered index in column_widgets definition, Table widget will ignore my column content definition and try to find widget automatically. First column in table contains checkbox and (hidden) label. This will fool table widget into thinking that column contains text. Trying to change checkbox state will result in error:

```
Traceback (most recent call last):
  File "/home/mzalewsk/.local/share/virtualenvs/robottelo-ps2A3-Bl/lib/python3.6/site-packages/widgetastic/log.py", line 117, in wrapped
    result = f(self, *args, **kwargs)
  File "/home/mzalewsk/.local/share/virtualenvs/robottelo-ps2A3-Bl/lib/python3.6/site-packages/widgetastic/widget.py", line 41, in wrapped
    return method(self, Fillable.coerce(value), *args, **kwargs)
  File "/home/mzalewsk/.local/share/virtualenvs/robottelo-ps2A3-Bl/lib/python3.6/site-packages/widgetastic/widget.py", line 1309, in fill
    self.column_name or self.position, value, self.text))
TypeError: Cannot fill column Select all rows, no widget and the value differs (wanted to fill True but there is 'Select row'
```

I know that `Select all rows` is awkward, but it needs to stay this way.

airgun/entities/subscription.py
mirzal: Addressed in comment below: https://github.com/SatelliteQE/airgun/pull/163#issuecomment-427853002

airgun/entities/subscription.py
```diff
@@ -0,0 +1,168 @@
...
+        for row in view.table.rows(name=entity_name):
+            row['Select all rows'].fill(True)
+        view.delete_button.click()
+        view.confirm_deletion.confirm()
+        self._wait_for_process_to_finish(
+                'Delete Upstream Subscription', has_manifest=True)
+
+
+@navigator.register(SubscriptionEntity, 'All')
+class SubscriptionList(NavigateStep):
```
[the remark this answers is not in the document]
mirzal: Docstrings added in e681427b4ff1b53eabeb2e7737cbcb2f2b095339

mirzal:
@oshtaier 
I somehow missed two comments back in August. One is now resolved by commit, to another one I replied with explanation (https://github.com/SatelliteQE/airgun/pull/163#pullrequestreview-148864018).
I think that's all? Or have I missed something else?

I'll re-run tests and post results after @abalakh review.

airgun/entities/subscription.py
```diff
@@ -0,0 +1,211 @@
...
+        return view.details.provided_products.read()
+
+    def enabled_products(self, entity_name):
+        """Read list of enabled products provided by subscription
+        Catches possible exception to always return known data structure
+        :param entity_name: Name of subscription
+        :return: List of strings with product names (may be empty)
+        """
+        view = self.navigate_to(self, 'Details', entity_name=entity_name)
+        view.enabled_products.select()
```
abalakh: nitpick: not needed anymore. Next line you access `enabled_products` tab widget, and when widget from specific tab is access, it gets selected automatically.

oshtaier:
please, unite all your test cases into one end-to-end scenario and run it.
appreciate, if you can do that today or at least till Monday morning. Want to have your code merged before going to 6.5 branching procedure

mirzal:
@oshtaier 

Let's merge it based on results of tests that we already have, OK? Posted below.

I'm not sure if merging all current tests into one makes sense from Robottelo perspective. Their goal was to show that Airgun side is working. We did tests like that earlier, but removed all of them as part of https://github.com/SatelliteQE/robottelo/pull/6336 . We didn't try to merge them into one big test then.

I think the correct approach is to merge this PR based on what we have, drop Robottelo PR #6265 (as it doesn't bring any value) and port tier2 / tier3 tests from Robottelo `tests/foreman/ui/test_subscription.py` in new PR.

If you are sure that merged tests are way to go, I can update Robottelo PR later today.

```
$ pytest -v tests/foreman/ui_airgun/test_subscription.py
test_subscription.py::test_positive_upload_and_delete PASSED         [ 14%]
test_subscription.py::test_positive_delete_confirmation PASSED       [ 28%]
test_subscription.py::test_read_provided_products PASSED             [ 42%]
test_subscription.py::test_read_enabled_products PASSED              [ 57%]
test_subscription.py::test_negative_read_enabled_products PASSED     [ 71%]
test_subscription.py::test_search_subscription PASSED                [ 85%]
test_subscription.py::test_negative_search_subscription PASSED       [ 100%]
```

abalakh:
@mirzal no objections regarding airgun part. But we can't merge robottelo part with tier1 tests anymore.

mirzal:
@abalakh I'd say let's close https://github.com/SatelliteQE/robottelo/pull/6265 without merging. There is nothing useful to merge there - only bunch of tier1 tests and tests without docstrings and markers. tier2 and tier3 - things that we want to have in Robottelo now - were never part of that PR.

Issue #137, opened 2018-07-06
Views: Packages entity


abalakh:
Location:
Content -> Packages
(make sure you have synced repo with some packages first)

Views:
* [ ] Packages list
* [ ] Package Details - Details tab
* [ ] Package Details - Files tab

abalakh:
Not required for GA (no t2/t3 tests):
* [ ] Package Details - Dependencies tab
* [ ] Package Details - Repositories tab

PR #169, opened 2018-08-06
Implement Package entity

ldjebran:
close https://github.com/SatelliteQE/airgun/issues/137

### DIFF
--- a/airgun/views/package.py
+++ b/airgun/views/package.py
@@ -1,0 +1,92 @@
+
+from widgetastic.widget import Checkbox, Select, Text, View
+from widgetastic_patternfly import BreadCrumb
+
+from airgun.views.common import (
+    BaseLoggedInView,
+    ReadOnlyEntry,
+    SatTab,
+    SatTable,
+    SearchableViewMixin,
+)
+from airgun.widgets import ItemsList
+
+
+class SearchablePackageViewMixin(SearchableViewMixin):
+    """Packages search widget"""
+    repository = Select(locator=".//select[@ng-model='repository']")
+    applicable = Checkbox(locator=".//input[@ng-model='showApplicable']")
+    upgradable = Checkbox(locator=".//input[@ng-model='showUpgradable']")
+
+    def search(self, query, repository='All Repositories', applicable=False,
+               upgradable=False):
+        """Perform search using search box on the page and return table
+        contents.
+
+        :param str query: search query to type into search field. E.g.
+            ``name = "bar"``.
+
+        :param str repository: repository name to select when searching for the
+            package.
+        :param bool applicable: To show only applicable packages
+        :param bool upgradable: To show only upgradable packages
+        :return: list of dicts representing table rows
+        :rtype: list
+        """
+
+        self.repository.fill(repository)
+        # set the upgradable first as if enabled applicable element will be
+        # disabled
+        self.upgradable.fill(upgradable)
+        if not upgradable:
+            self.applicable.fill(applicable)
+
+        return super(SearchablePackageViewMixin, self).search(query)
+
+
+class PackagesView(BaseLoggedInView, SearchablePackageViewMixin):
+    """Main Packages view"""
+    title = Text("//h2[contains(., 'Packages')]")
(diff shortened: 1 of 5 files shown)


ldjebran:
Packages page:
pages_page

Package details:
package_details

Package files:
 
package_files

airgun/views/package.py
```diff
@@ -0,0 +1,92 @@
...
+            self.breadcrumb, exception=False)
+
+        return (
+                breadcrumb_loaded
+                and self.breadcrumb.locations[0] == 'Packages'
+        )
+
+    @View.nested
+    class details(SatTab):
+        applicable_to = ReadOnlyEntry(name='Applicable To')
```
oshtaier: you don't need to sort them alphabetically, but more according to UI order as it easier to read, understand and actually more valid

airgun/views/package.py
```diff
@@ -0,0 +1,92 @@
...
+        license = ReadOnlyEntry(name='License')
+        size = ReadOnlyEntry(name='Size')
+        source_rpm = ReadOnlyEntry(name='Source RPM')
+        summary = ReadOnlyEntry(name='Summary')
+        upgradable_for = ReadOnlyEntry(name='Upgradable For')
+        url = ReadOnlyEntry(name='Url')
+
+    @View.nested
+    class files(SatTab):
+        package_files = ItemsList(locator=".//div[@data-block='content']//ul")
```
oshtaier: that widget expect list of clickable elements, not static text, so not sure whether it is correct to use it or not

airgun/views/package.py
```diff
@@ -0,0 +1,92 @@
...
+            ``name = "bar"``.
+
+        :param str repository: repository name to select when searching for the
+            package.
+        :param bool applicable: To show only applicable packages
+        :param bool upgradable: To show only upgradable packages
+        :return: list of dicts representing table rows
+        :rtype: list
+        """
+
```
oshtaier: redundant line

airgun/views/package.py
```diff
@@ -0,0 +1,92 @@
...
+    upgradable = Checkbox(locator=".//input[@ng-model='showUpgradable']")
+
+    def search(self, query, repository='All Repositories', applicable=False,
+               upgradable=False):
+        """Perform search using search box on the page and return table
+        contents.
+
+        :param str query: search query to type into search field. E.g.
+            ``name = "bar"``.
+
```
oshtaier: redundant line

airgun/views/package.py
```diff
@@ -0,0 +1,92 @@
...
+    BaseLoggedInView,
+    ReadOnlyEntry,
+    SatTab,
+    SatTable,
+    SearchableViewMixin,
+)
+from airgun.widgets import ItemsList
+
+
+class SearchablePackageViewMixin(SearchableViewMixin):
```
oshtaier: I would say that is a little bit redundant class, maybe you can use similar approach like here:
https://github.com/SatelliteQE/airgun/blob/master/airgun/views/errata.py#L33

airgun/views/package.py
ldjebran: @oshtaier They have the same purpose the same locators except that the list here is read only , perhaps add a class to widgets:

```python
class ItemsListReadOnly(ItemsList):

    def fill(self, value):
        raise NotImplementedError('The widget is read only')
``` 
what do you think?

airgun/views/package.py
ldjebran: No problem I can do that , but still think that this should be a separate widget.

airgun/views/package.py
oshtaier: it is not 'not implemented', but more as expected behavior. Hmm, I am not sure what type of exception to use, but idea looks good

airgun/views/package.py
oshtaier: they still are different (installable vs upgradable), otherwise we can think about uniting them into generic widget. But once more, it is valid for some generic situation, not separate View for single situation without any re-use. Just thoughts

ldjebran:
comments addressed test output https://github.com/SatelliteQE/robottelo/pull/6171#issuecomment-411071867

PR #186, opened 2018-08-20
Added RHAI Actions

quarckster:
Added RHAI Actions entity and the respectful views.

### DIFF
--- a/airgun/views/rhai.py
+++ b/airgun/views/rhai.py
@@ -1,4 +1,5 @@
 from widgetastic.widget import Table, Text, TextInput
+from widgetastic_patternfly import Button
 
 from airgun.views.common import BaseLoggedInView
 from airgun.widgets import ActionsDropdown, SatTable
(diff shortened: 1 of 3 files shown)


airgun/views/rhai.py
```diff
@@ -0,0 +1,22 @@
...
+from widgetastic.widget import Text
+
+from airgun.views.common import BaseLoggedInView
+
+
+class ActionsDetailsView(BaseLoggedInView):
+    title = Text(".//h1[normalize-space(.)='Actions']")
+    export_csv = Button("Export CSV")
+    stability_issues = Text(".//span[normalize-space(.)='Stability']"
+                            "/parent::span/following-sibling::span")
```
abalakh: Could you please share html snippet of that element? There should be simpler way to locate it

quarckster:
```html
<a ng-class="category.id" ui-sref="app.topic({id: category.id, product: getSelectedProduct()})" href="/redhat_access/insights/actions/stability?product=all" class="stability">
    <i ng-class="category.id + '-color'" aria-hidden="true" class="fa fa-square inline-right stability-color"></i>
    <span translate="translate" class="desc">
        <span>Stability</span>
    </span>
    <span class="count">&nbsp;(1)</span>
</a>
```

abalakh: Here you go, simpler and independent from current user's locale:
```
.//a[@class='stability']/span[@class='count']
```
If you feel like relying on exact text is a must in such case (which i strongly discourage), then you can try:
```
.//a[span[normalize-space(.)='Stability']]/span[@class='count']
```

quarckster:
@abalakh @oshtaier changes made, please review

PR #175, opened 2018-08-13
Fixed ENSURE_PAGE_SAFE script

quarckster:
- remote webdriver doesn't work without `desired_capabilities`;
- navigation to Insights page caused the error in  `ENSURE_PAGE_SAFE` script

PR #170, opened 2018-08-07
Add react spinner to ENSURE_PAGE_SAFE

mirzal:
Some pages (Subscriptions and Red Hat Repositories for certain, maybe some others as well) use new react-based spinner widget. Our ensure_page_safe function did not take it into account and tried to proceed with automation even if page was not ready yet.

Function looks for element with class `loading-state` inside div with id `reactRoot`. That element is on page only when content is being loaded, i.e. it is removed from source on finish (instead of merely making it invisible). `reactRoot` is not present on other pages at all, so function will always return true on them.

PR #187, opened 2018-08-20
Fixed angular pending requests in ENSURE_PAGE_SAFE script

quarckster:
It seems changes in #175 were not enough. Airgun doesn't wait for the page loading in RHAI pages. I tried the solution from https://stackoverflow.com/a/13402028 and now `ENSURE_PAGE_SAFE` works as expected on RHAI pages.

### DIFF
--- a/airgun/browser.py
+++ b/airgun/browser.py
@@ -508,10 +508,16 @@ class AirgunBrowserPlugin(DefaultPlugin):
             Ajax.activeRequestCount < 1
         }
         function angularNoRequests() {
-         return (typeof angular === "undefined" ||
-          typeof angular.element(document).injector() === "undefined") ? true :
-          angular.element(document).injector().get(
-           "$http").pendingRequests.length < 1
+         if (typeof angular === "undefined") {
+           return true
+         } else if (typeof angular.element(
+             document).injector() === "undefined") {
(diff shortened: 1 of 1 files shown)


mirzal:
@quarckster I tried running my Subscriptions tests on top of your changes. Only one test fails: Airgun tries to click "Add Subscription" button when it isn't loaded yet ([screenshot](https://user-images.githubusercontent.com/12373754/44343565-847e9680-a48e-11e8-8ac0-29cd501137bd.png)). Since other tests that click other buttons are not affected, I am pretty sure that this can be worked around in views or entities.

On the other hand, my change from PR #170 waits for buttons to appear every time without any additional. Would #170 solve your problems with RHAI pages?

quarckster:
@mirzal it doesn't work:
```js
document.querySelector("#reactRoot .loading-state")
null
```

mirzal:
Looking at Insights sub-pages, I don't see reactRoot element at all.

I'd say that we need both - this for Insights and PR #170 for Subscriptions / Repositories / Select Organization (where react is used to pull in additional data).

quarckster:
I think the Satellite developers could add some variable into the global js namespace that would be designate is data loading or page rendering.

PR #197, opened 2018-08-24
RHAI Inventory fix

quarckster:
- fixed session property;
- improved search method of `InventoryHostEntity`

### DIFF
--- a/airgun/entities/rhai/inventory.py
+++ b/airgun/entities/rhai/inventory.py
@@ -15,6 +15,7 @@ class InventoryHostEntity(BaseEntity):
         """Search a certain host."""
         view = self.navigate_to(self, "All")
         view.search.fill(host_name)
+        return view.table
 
 
 @navigator.register(InventoryHostEntity, "All")


quarckster:
@oshtaier @abalakh please review and merge if it's ok.

oshtaier:
any specific reason for that change? We always return table in any search that we have in our code and make actual assertion on a test level. That makes sense as sometimes you need to fetch other column values rather than name

quarckster:
@oshtaier `InventoryAllHosts` cannot use `SearchableViewMixin`, for that [test](https://github.com/SatelliteQE/robottelo/pull/6199/files#diff-9d8bd439118c370472292e633d7d4597R70) I need to check if a host in the table.

oshtaier:
still all search methods return table to test, so you can check on a test level whether host name is present in the table, so:
```
        view = self.navigate_to(self, "All")
        view.search.fill(host_name)
        return view.table
```
and on test level:
```
assert session.rhai_inventory.search(host_name)[0]['System Name'] == host_name
```
that approach let you check all other columns, e.g.:
```
assert session.rhai_inventory.search(host_name)[0]['Column_2_Value'] == expected_value
```

That is a common approach used all other out entities

quarckster:
@oshtaier fixed, thank you so much for all your hard work you've done and please sorry me for my stupidity. I hope I don't disturb you a lot.

oshtaier:
@quarckster Thank you for your patience =)

oshtaier:
please rebase,
https://github.com/SatelliteQE/airgun/blob/master/airgun/session.py#L285

PR #198, opened 2018-08-28
[POC] Taretto adoptation

quarckster:
This PR shows one of possible ways of [taretto](https://github.com/RedHatQE/taretto) adoptation for airgun. The main goal is getting clean, straightforward and more object oriented API for satellite tests.

# Directory structure

All taretto related code moved to `base` directory:
```
airgun/base
├── __init__.py
├── application
│   ├── __init__.py
│   └── implementations
│       ├── __init__.py
│       ├── rest_api.py
│       └── web_ui.py
├── browser_manager.py
└── modeling
    ├── __init__.py
    └── entities
        ├── __init__.py
        └── domains.py
```
# Objects hierarchy
`Application` - a main entry point of taretto. It holds some basic information about the application we are testing, such as its URL or schema.

`Implementations` - in the other words the interfaces of the testing application. It can be web ui, rest api, ssh, cli and so on. This PR implements `web_ui` and `rest_api`. Each implementation has its own low level library. For example the lowest level of the web ui implementation is `selenium.webdriver`, for rest api it could be `requests`

`Entity Objects` - the conception of the entities is already used in airgun, in this PR it's extended by adding `Collection Objects`. Collections required when you need to manipulate on several same type entities.

# Sentaku

Taretto includes [sentaku](https://github.com/RedHatQE/Sentaku) library. It allows to specify multiple methods on an object with the same name. 

# Example

This PR implements `DomainEntity` and `DomainsCollection`. A simple create tests could be look like this:

```python
import pytest

from airgun.base.application import Application
from airgun.base.application.implementations.rest_api import RESTAPI
from airgun.base.application.implementations.web_ui import WebUI


@pytest.fixture
def application():
    return Application()


@pytest.mark.parametrize(context=[WebUI, RESTAPI])
def test_domain_create(application, context):
    with application.context.use(context):
        domain = application.collections.domains.create(random_string())
    assert domain.exists
```
Method `create` is implemented as a contextual method for the domains collection object.


quarckster:
@omaciel

omaciel:
@quarckster great! Let me set some time to chat with you about this PR? I know exactly what it is doing because Pete and I talked about it many, many times about the idea :) The one important thing to discuss will be:

- Do we want to implement something like this now for Satellite? The answer will probably be `not right now`

- Can some of the ideas here be implemented right now or in the short term? That is something I want to learn more from you :)

omkarkhatavkar:

lhellebr:
I believe this is very outdated and will never be worked on. I suggest closing this. Any objections, @quarckster ?

quarckster:
> I believe this is very outdated and will never be worked on. I suggest closing this. Any objections, @quarckster ?

Sure, no objections.

Issue #122, opened 2018-07-05
Views: Content view entity


abalakh:
Location:
Content -> Content Views

Views:
* [x] Content views list
* [x] Create content view
* [x] CV Details - Details tab
* [x] CV Details - Versions tab
* [ ] CV Details - Versions - (version) - Details tab
* [x] CV Details - Publish new version
* [x] CV Details - Promote version
* [ ] CV Details - Delete version
* [x] CV Details - Content Views tab (for composite CV)
* [x] CV Details - Yum Content - Repositories tab
* [ ] CV Details - Yum Content - Filters tab
* [ ] CV Details - Yum Content - Yum Filters - New Yum Filter
* [ ] CV Details - Yum Content - Yum Filters - Details tab
* [ ] CV Details - Yum Content - Yum Filters - RPMs tab
* [ ] CV Details - Yum Content - Yum Filters - Affected Repositories tab
* [x] CV Details - Puppet Modules tab
* [x] CV Details - Puppet Modules - Add Puppet Module
* [x] CV Details - Puppet Modules - Add Puppet Module - Select Version
* [ ] CV Details - Tasks - (task)
* [ ] CV Details - Copy content view
* [ ] CV Details - Remove content view
* [ ] CV Details - Remove content view (with promoted versions)

abalakh:
Not required for GA:
* [ ] CV Details - Apt Repositories tab
* [ ] CV Details - File Repositories tab
* [ ] CV Details - Container Images - Repositories tab
* [ ] CV Details - Container Images - Filters tab
* [ ] CV Details - Container Images - Filters - New Container Image Tag Filter
* [ ] CV Details - Container Images - Filters - (filter) - Details tab
* [ ] CV Details - Container Images - Filters - (filter) - Tags tab
* [ ] CV Details - Container Images - Filters - (filter) - Affected Repositories tab
* [ ] CV Details - OSTree Content tab
* [ ] CV Details - History tab
* [ ] CV Details - Tasks tab

PR #203, opened 2018-09-11
Added some CV helpers and views

abalakh:
Part of #122

### DIFF
--- a/airgun/entities/contentview.py
+++ b/airgun/entities/contentview.py
@@ -29,11 +29,10 @@ class ContentViewEntity(BaseEntity):
     def delete(self, entity_name):
         """Delete existing content view"""
         view = self.navigate_to(self, 'Delete', entity_name=entity_name)
-        if view.conflicts_present:
-            raise AssertionError(
-                'Unable to delete content view. Following conflicts are'
-                'present: {}'.format(view.table.read())
-            )
+        assert not view.conflicts_present, (
+            'Unable to delete content view. '
+            'Following conflicts are present: {}'.format(view.table.read())
+        )
         view.remove.click()
         view.flash.assert_no_error()
         view.flash.dismiss()
@@ -87,6 +86,8 @@ class ContentViewEntity(BaseEntity):
         """Add puppet module to selected view either by its author name or by
         its version.
 
+        :param str entity_name: content view name
+        :param str module_name: puppet module name
         :param str optional filter_term: can be used to filter the module by
             'author' or by 'version'.
         """
@@ -158,6 +159,9 @@ class ContentViewEntity(BaseEntity):
             self, entity_name, version_name, query, repo=None):
         """Search for a package inside content view version
 
+        :param str entity_name: content view name
+        :param str version_name: content view version name
+        :param str query: search query for content view version's package
         :param str optional repo: repository name to filter by
         """
         view = self.navigate_to(
@@ -171,8 +175,10 @@ class ContentViewEntity(BaseEntity):
                        lces=None):
         """Remove content view version.
 
+        :param str entity_name: content view name
+        :param str version_name: content view version name
         :param bool completely: complete content view version removal if True
-            or removal from all lifecycle environments otherwise
+            or just disassociating from all lifecycle environments otherwise
         :param list optional lces: list of lifecycle environment names to
             select on content view version removal screen
         """
--- a/airgun/entities/contentview.py
+++ b/airgun/entities/contentview.py
@@ -4,11 +4,16 @@ from airgun.entities.base import BaseEntity
 from airgun.navigation import NavigateStep, navigator
 from airgun.views.contentview import (
     AddNewPuppetModuleView,
+    ContentViewCopyView,
     ContentViewCreateView,
     ContentViewEditView,
+    ContentViewRemoveView,
     ContentViewTableView,
+    ContentViewVersionDetailsView,
     ContentViewVersionPromoteView,
     ContentViewVersionPublishView,
+    ContentViewVersionRemoveConfirmationView,
+    ContentViewVersionRemoveView,
     SelectPuppetModuleVersionView,
 )
 
@@ -16,24 +21,43 @@ from airgun.views.contentview import (
 class ContentViewEntity(BaseEntity):
 
     def create(self, values):
+        """Create a new content view"""
         view = self.navigate_to(self, 'New')
         view.fill(values)
         view.submit.click()
 
     def delete(self, entity_name):
-        view = self.navigate_to(self, 'Edit', entity_name=entity_name)
-        view.actions.fill('Remove Content View')
-        view.dialog.confirm()
+        """Delete existing content view"""
+        view = self.navigate_to(self, 'Delete', entity_name=entity_name)
+        if view.conflicts_present:
+            raise AssertionError(
+                'Unable to delete content view. Following conflicts are'
+                'present: {}'.format(view.table.read())
+            )
+        view.remove.click()
+        view.flash.assert_no_error()
+        view.flash.dismiss()
+
+    def copy(self, entity_name, new_name):
+        """Make a copy of existing content view"""
+        view = self.navigate_to(self, 'Copy', entity_name=entity_name)
+        view.new_name.fill(new_name)
+        view.create.click()
+        view.flash.assert_no_error()
+        view.flash.dismiss()
 
     def search(self, value):
+        """Search for content view"""
         view = self.navigate_to(self, 'All')
         return view.search(value)
 
     def read(self, entity_name):
+        """Read content view values"""
         view = self.navigate_to(self, 'Edit', entity_name=entity_name)
         return view.read()
 
     def update(self, entity_name, values):
+        """Update existing content view"""
         view = self.navigate_to(self, 'Edit', entity_name=entity_name)
         filled_values = view.fill(values)
         view.flash.assert_no_error()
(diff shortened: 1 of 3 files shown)


airgun/entities/contentview.py
```diff
@@ -4,43 +4,68 @@
...
-        view = self.navigate_to(self, 'Edit', entity_name=entity_name)
-        view.actions.fill('Remove Content View')
-        view.dialog.confirm()
+        """Delete existing content view"""
+        view = self.navigate_to(self, 'Delete', entity_name=entity_name)
+        if view.conflicts_present:
+            raise AssertionError(
+                'Unable to delete content view. Following conflicts are'
+                'present: {}'.format(view.table.read())
+            )
```
ldjebran: is there a specific reason to raise explicitly assert error,
assert key word is not only for test frameworks 
```python
assert not view.conflicts_present, 'Unable to delete content view. Following conflicts are'
                'present: {}'.format(view.table.read())
```
what do you think?

airgun/entities/contentview.py
```diff
@@ -115,6 +140,65 @@ def promote(self, entity_name, version_name, lce_name):
...
+    def search_version(self, entity_name, query):
+        """Search for content view version"""
+        view = self.navigate_to(self, 'Edit', entity_name=entity_name)
+        return view.versions.search(query)
+
+    def search_version_package(
+            self, entity_name, version_name, query, repo=None):
+        """Search for a package inside content view version
+
+        :param str optional repo: repository name to filter by
```
ldjebran: Think if we make a description for one param, we have to make for all of them, or add that param to desc.
```
Search for a package inside content view version, optionally we can indicate 
a repository to search by.
```
what do you think?

airgun/views/contentview.py
```diff
@@ -192,6 +256,58 @@ def is_displayed(self):
...
+                and self.breadcrumb.locations[2] == 'Versions'
+        )
+
+    @View.nested
+    class yum_repositories(SatSecondaryTab, SearchableViewMixin):
+        TAB_NAME = 'Yum Repositories'
+        table = SatTable('.//table')
+
+    @View.nested
+    class rpm_packages(SatSecondaryTab):
```
ldjebran: why not inherit from SearchableViewMixin also?

airgun/views/contentview.py
```diff
@@ -210,3 +326,123 @@ def is_displayed(self):
...
+    class rpms(SatSecondaryTab, SearchableViewMixin):
+        TAB_NAME = 'RPMs'
+
+        exclude_no_errata = Checkbox(
+            locator=".//input[@type='checkbox']"
+                    "[@ng-model='filter.original_packages']"
+        )
+        add_rule = Text(".//button[@ng-click='addRule()']")
+        remove_rule = Text(".//button[@ng-click='removeRules(filter)']")
+
```
ldjebran: redundant line :)

airgun/entities/contentview.py
```diff
@@ -115,6 +140,65 @@ def promote(self, entity_name, version_name, lce_name):
...
+            entity_name=entity_name,
+            version_name=version_name,
+        )
+        return view.rpm_packages.search(query, repo=repo)
+
+    def remove_version(self, entity_name, version_name, completely=True,
+                       lces=None):
+        """Remove content view version.
+
+        :param bool completely: complete content view version removal if True
```
oshtaier: what is the difference? Just not sure that I understand what 'completely' word means

airgun/views/common.py
```diff
@@ -115,6 +115,11 @@ class listremove(SatSecondaryTab):
     ROOT = ParametrizedLocator(
         './/nav[@class="ng-scope" or not(@*)]/following-sibling::div')
 
+    TAB_LOCATOR = ParametrizedLocator(
```
oshtaier: what was a reason for this change?

airgun/views/contentview.py
```diff
@@ -58,6 +62,50 @@ def is_displayed(self):
...
+
+class ContentViewRemoveView(BaseLoggedInView):
+    breadcrumb = BreadCrumb()
+
+    conflicts = Text("//div[@ng-show='conflictingVersions.length > 0']")
+    table = SatTable('.//table')
+    remove = Text(".//button[@ng-click='delete()']")
+    cancel = Text(
+        ".//a[contains(@class, 'btn')][@ui-sref='content-view.versions']")
```
oshtaier: does it have `Cancel` text? it will be better then

airgun/views/contentview.py
```diff
@@ -192,6 +256,58 @@ def is_displayed(self):
...
+        table = SatTable('.//table')
+
+    @View.nested
+    class rpm_packages(SatSecondaryTab):
+        TAB_NAME = 'rpm Packages'
+        repo_filter = SatSelect(".//select[@ng-model='repository']")
+        searchbox = Search()
+        table = SatTable(".//table")
+
+        def search(self, query, repo=None):
```
oshtaier: seeing that type of search not for first time. Is it time consuming procedure to make it generic?

airgun/views/contentview.py
```diff
@@ -210,3 +326,123 @@ def is_displayed(self):
...
+    next = Text(".//button[@ng-click='processSelection()']")
+    cancel = Text(".//button[normalize-space(.)='Cancel']")
+
+    @property
+    def is_displayed(self):
+        breadcrumb_loaded = self.browser.wait_for_element(
+            self.breadcrumb, exception=False)
+        return (
+                breadcrumb_loaded
+                and self.breadcrumb.locations[0] == 'Content Views'
```
oshtaier: does it make sense to have so many conditions? Here and previously

airgun/views/contentview.py
```diff
@@ -210,3 +326,123 @@ def is_displayed(self):
...
+            column_widgets={
+                0: Checkbox(locator=".//input[@type='checkbox']"),
+                4: Text("./button[contains(@ng-click, 'rule.editMode')]"),
+            },
+        )
+
+    @View.nested
+    class affected_repositories(SatSecondaryTab):
+        TAB_NAME = 'Affected Repositories'
+        filter_toggle = RadioGroup(".//div[@class='col-sm-8']")
```
oshtaier: any other alternatives

airgun/views/contentview.py
```diff
@@ -210,3 +326,123 @@ def is_displayed(self):
...
+                    "[@ng-model='filter.original_packages']"
+        )
+        add_rule = Text(".//button[@ng-click='addRule()']")
+        remove_rule = Text(".//button[@ng-click='removeRules(filter)']")
+
+        table = SatTable(
+            locator='//table',
+            column_widgets={
+                0: Checkbox(locator=".//input[@type='checkbox']"),
+                4: Text("./button[contains(@ng-click, 'rule.editMode')]"),
```
oshtaier: going to add fix to application?

airgun/entities/contentview.py
abalakh: Should i rephrase? Non-complete content view version removal just 'marks' cvv as deleted and only deletes it from lces, when complete means complete removal from app.

airgun/views/common.py
abalakh: since it wasn't overriding TAB_LOCATOR before, both primary and secondary tabs could be caught by this locator. It's just we were lucky every single automated page had different labels for primary and secondary tabs; now i faced the page where we have both primary and secondary 'Details' tab, this way the issue was uncovered :)

airgun/entities/contentview.py
oshtaier: You can just put these words into docstring if possible

airgun/views/contentview.py
abalakh: really? why? i'd prefer unique attr over localized string

airgun/views/common.py
oshtaier: so it will not affect existing tests?

airgun/views/contentview.py
abalakh: if i override is_searchable and search, then what's the point of inheritance here if 100% of features will be overwritten? :)
btw is_searchable doesn't really make sense here as welcome message might be shown only on entities list page, not inside some entity details tab.

airgun/views/common.py
abalakh: The only secondary tabs we used till now are common 'Add' and 'List/Remove' tabs for resources lists. robottelo's PR does have test results for them

airgun/views/contentview.py
oshtaier: because none of these attributes point on `cancel` specifically, if we have`[@ui-sref='content-view.versions.cancel']` your approach would be better

airgun/views/contentview.py
abalakh: Submitted https://github.com/SatelliteQE/airgun/issues/205 . Imo it's not that big deal, but requires update in many places and lots of retesting.

airgun/views/contentview.py
abalakh: Yep, unfortunately that's special case when we have multiple different CV pages with the same breadcrumb location, so we can't just rely on breadcrumb only to figure out where we are.

airgun/views/contentview.py
abalakh: That page isn't that nice, and that's just a start :)
`<div class="col-sm-8">` is all i have for that `div`. I can find radio button and go 3 parents higher but i'm not sure such locator will be way more readable, smth like that:
`//div[div/label/input[@ng-model='showRepos']]` (can't use `//` there to shorten path as those div elements are inside other divs, so many results will be found)

airgun/views/contentview.py
oshtaier: just was my opinion, it is not better, but more logical

airgun/views/contentview.py
oshtaier: okay

airgun/views/contentview.py
oshtaier: okay

PR #204, opened 2018-09-11
Adding dashboard and task entities

oshtaier:
!!! Remove backward compatibility for previous widgetastic versions

### DIFF
--- a/airgun/entities/dashboard.py
+++ b/airgun/entities/dashboard.py
@@ -12,7 +12,14 @@ class DashboardEntity(BaseEntity):
         view = self.navigate_to(self, 'All')
         return view.search(value)
 
-    def read(self):
+    def read(self, widget_name):
+        """Read specific widget value"""
+        view = self.navigate_to(self, 'All')
+        if widget_name not in view.widget_names:
+            raise ValueError('Provide correct widget name to be read')
+        return getattr(view, widget_name).read()
+
+    def read_all(self):
         """Read all dashboard widgets values"""
         view = self.navigate_to(self, 'All')
         return view.read()
@@ -37,6 +44,4 @@ class OpenDashboard(NavigateStep):
         """Disable auto-refresh feature for dashboard entity each time
         navigation to the page is finished
         """
-        if self.view.browser.element(self.view.AUTO_REFRESH).get_attribute(
-                'data-original-title') == 'Auto refresh on':
-            self.view.browser.element(self.view.AUTO_REFRESH).click()
+        self.view.refresh.fill(False)
--- a/airgun/entities/dashboard.py
+++ b/airgun/entities/dashboard.py
@@ -1,0 +1,42 @@
+from airgun.entities.base import BaseEntity
+from airgun.navigation import NavigateStep, navigator
+from airgun.views.dashboard import DashboardView
+
+
+class DashboardEntity(BaseEntity):
+
+    def search(self, value):
+        """Initiate search procedure that applied on all dashboard widgets.
+        Return widgets values as a result
+        """
+        view = self.navigate_to(self, 'All')
+        return view.search(value)
+
+    def read(self):
+        """Read all dashboard widgets values"""
+        view = self.navigate_to(self, 'All')
+        return view.read()
+
+    def action(self, values):
+        """Perform action against specific widget. In most cases, re-direction
+        to another entity is happened
+        """
+        view = self.navigate_to(self, 'All')
+        view.fill(values)
+
+
+@navigator.register(DashboardEntity, 'All')
+class OpenDashboard(NavigateStep):
+    """Navigate to Dashboard page"""
+    VIEW = DashboardView
+
+    def step(self, *args, **kwargs):
+        self.view.menu.select('Monitor', 'Dashboard')
+
+    def post_navigate(self, _tries=0, *args, **kwargs):
+        """Disable auto-refresh feature for dashboard entity each time
+        navigation to the page is finished
+        """
+        if self.view.browser.element(self.view.AUTO_REFRESH).get_attribute(
+                'data-original-title') == 'Auto refresh on':
+            self.view.browser.element(self.view.AUTO_REFRESH).click()
--- a/airgun/views/dashboard.py
+++ b/airgun/views/dashboard.py
@@ -1,0 +1,162 @@
+from widgetastic.widget import (
+    Text,
+    View,
+    Widget,
+)
+
+from airgun.views.common import (
+    BaseLoggedInView,
+    SatTable,
+)
+from airgun.widgets import (
+    ActionsDropdown,
+    PieChart,
+    Search,
+)
+
+
+class ItemValueList(Widget):
+    """List of name-value pairs. Each name element from every pair is clickable
+
+     Example html representation::
+
+    <ul>
+        <li>
+            <a class="dashboard-links"...>Hosts with no reports</a>
+            <h4>5</h4>
+        </li>
+     """
+    LABELS = ".//li/a[@class='dashboard-links']"
+    LABEL = ".//li/a[@class='dashboard-links'][normalize-space(.)='%s']"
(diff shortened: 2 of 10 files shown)


airgun/widgets.py
```diff
@@ -1441,3 +1443,23 @@ def read(self):
...
+    chart_title_value = Text(
+        ".//*[name()='svg']//*[name()='tspan']"
+        "[contains(@class,'donut-title-big-pf')]"
+    )
+
+    def read(self):
+        """Return dictionary that contains chart title name as key and chart
+        value as its value
+        """
+        return {self.chart_title_text.read(): self.chart_title_value.read()}
```
ldjebran: Think chart_title_text and chart_title_value can target different meanings in the chart, and they should be returned separately,
```python
return dict(title=self.chart_title_text.read(), title_value=self.chart_title_value.read())
```  
or let it without this read function and we should have chart_title_text and chart_title_value returned by the default read.

airgun/views/dashboard.py
```diff
@@ -0,0 +1,162 @@
...
+
+    <ul>
+        <li>
+            <a class="dashboard-links"...>Hosts with no reports</a>
+            <h4>5</h4>
+        </li>
+     """
+    LABELS = ".//li/a[@class='dashboard-links']"
+    LABEL = ".//li/a[@class='dashboard-links'][normalize-space(.)='%s']"
+    VALUE = './/h4[preceding-sibling::a[contains(., "{}")]]'
```
ldjebran: Use one string formatting convention via format function or via %s

airgun/views/dashboard.py
```diff
@@ -0,0 +1,162 @@
...
+            values[name] = value
+        return values
+
+    def fill(self, value):
+        """Click on specific criteria from the widget list"""
+        self.browser.click(
+            self.browser.element(self.LABEL % value))
+
+
+class TotalCount(Widget):
```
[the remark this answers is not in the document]
oshtaier: because there are more dashboard widgets that can use it, I just didn't implement them. Same for Items class

airgun/views/dashboard.py
oshtaier: sounds good

airgun/widgets.py
oshtaier: actually on a chart it looks like 'Success: 100%' in the center of a chart, so previous statement target that idea, but not a problem, I can do it with `dict()` idea. Default read is not good in that case

airgun/views/dashboard.py
```diff
@@ -0,0 +1,162 @@
...
+    """Return total hosts count from Host Configuration Status type of
+    widgets
+    """
+    total_count = Text(".//h4[@class='total']")
+
+    def read(self):
+        """Return hosts count from widget. Usually it is a string like
+        'Total Hosts: 5'
+        """
+        _, count = self.total_count.read().split(': ')
```
ldjebran: we can omit the space 
```
 _, count = self.total_count.read().split(':')
```
but in this concrete usage prefere partition usage
```
_, _, count = self.total_count.read().partition(':')
```
"not a blocker"

airgun/views/dashboard.py
```diff
@@ -0,0 +1,162 @@
...
+
+    def read(self):
+        """Return a dictionary where keys are widget criteria names and
+        values are number of hosts that correspond to these criteria
+        """
+        values = {}
+        for item in self.browser.elements(self.LABELS):
+            name = self.browser.text(item)
+            value = self.browser.text(self.VALUE.format(name))
+            values[name] = value
```
ldjebran: if the total is an integer, perhaps also the values should be integers too

airgun/views/dashboard.py
```diff
@@ -0,0 +1,162 @@
...
+        'Total Hosts: 5'
+        """
+        _, count = self.total_count.read().split(': ')
+        return int(count)
+
+
+class DashboardView(BaseLoggedInView):
+    title = Text("//h1[text()='Overview']")
+    manage = ActionsDropdown("//div[@class='btn-group']")
+    AUTO_REFRESH = "//a[contains(@href, '/?auto_refresh')]"
```
ldjebran: why AUTO_REFRESH in capital letters as understood this is a simple toggle button element

oshtaier: because it is a constant, not a class variable

airgun/views/dashboard.py
```diff
@@ -0,0 +1,162 @@
...
+            column_widgets={'Content View': Text('./a')}
+        )
+
+    @View.nested
+    class syncoverview(View):
+        ROOT = ".//li[@data-name='Sync Overview']"
+        syncs = SatTable('.//table')
+
+    @View.nested
+    class HostSubscription(View):
```
ldjebran: why some nested views we can write as camelcase for others we have not to?

airgun/widgets.py
ldjebran: @oshtaier the 100% can have nothing to do with Success
for example it can be Failed : 100%   the 100% is related action applied to how many hosts.
the action was 100% so action was applied to all hosts
overall action status is Failed but for some hosts it can be Success

airgun/views/dashboard.py
oshtaier: oops, I will put views that are left in camel case too. I did that because it became unreadable in small case

airgun/views/dashboard.py
```diff
@@ -0,0 +1,162 @@
...
+    class TaskStatus(View):
+        ROOT = ".//li[@data-name='Task Status']"
+        states = SatTable(
+            './/table',
+            column_widgets={'No. of Tasks': Text('./a')},
+        )
+
+        def fill(self, values):
+            if 'state' not in values or 'result' not in values:
+                raise AttributeError(
```
abalakh: ValueError

airgun/views/dashboard.py
```diff
@@ -0,0 +1,162 @@
...
+    class HostSubscription(View):
+        ROOT = ".//li[@data-name='Host Subscription Status']"
+        subscriptions = SatTable(
+            './/table',
+            column_widgets={0: Text('./a')}
+        )
+
+        def fill(self, values):
+            if 'type' not in values:
+                raise AttributeError('You need provide subscription task type')
```
abalakh: ValueError

airgun/views/dashboard.py
```diff
@@ -0,0 +1,162 @@
...
+    class LatestFailedTasks(View):
+        ROOT = ".//li[@data-name='Latest Warning/Error Tasks']"
+        tasks = SatTable(
+            './/table',
+            column_widgets={'Name': Text('./a')}
+        )
+
+        def fill(self, values):
+            if 'name' not in values:
+                raise AttributeError('You need provide name of the task')
```
abalakh: ValueError

airgun/entities/dashboard.py
```diff
@@ -0,0 +1,42 @@
...
+class DashboardEntity(BaseEntity):
+
+    def search(self, value):
+        """Initiate search procedure that applied on all dashboard widgets.
+        Return widgets values as a result
+        """
+        view = self.navigate_to(self, 'All')
+        return view.search(value)
+
+    def read(self):
```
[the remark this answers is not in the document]
oshtaier: `read` sounds more generic as anyway we can't have any other alternative way to read something

it takes around 8-10 seconds. I thought about that idea, but not sure if that has real impact as it is not very generic case. It is like saving around 1-2 minutes for whole test suite in one thread, but add more logic that required from test level.
Let's me know your decision as it will require re-implementing all cases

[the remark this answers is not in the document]
oshtaier: okay, it takes more than 10 seconds :(

airgun/widgets.py
oshtaier: okay, it doesn't make sense to change that method at that moment, as it will just add necessity to add more assert statement

PR #207, opened 2018-09-19
Fix edge workaround applied on chrome

m-bucher:
Workaround for edge was also applied on chrome browser; resulting in error:
`selenium.common.exceptions.WebDriverException: Message: unknown error: unsupported protocol`


m-bucher:
Thanks for the quick reaction :+1:

PR #219, opened 2018-10-08
Implemented Bookmarks entity

### LINKED ISSUE #151: Views: Bookmarks entity
Location:
Administer -> Bookmarks
    Example of katello style form for list/creation:
    Content -> Activation Keys
    Example of foreman style form for list/creation:
    Hosts -> Architectures

Views:
* [ ] Bookmarks list
* [ ] Edit Bookmark
* [ ] Add Bookmark form for search widget, both foreman and katello style

abalakh:
Closes #151

### DIFF
--- a/airgun/entities/base.py
+++ b/airgun/entities/base.py
@@ -14,24 +14,24 @@ class BaseEntity(object):
         """Create a bookmark.
 
         :param dict values: dictionary with keys 'name', 'query', 'public'
-        :param str optional search_query: if something different from `query`
-            should be typed into searchbox
+        :param str optional search_query: a query to type into searchbox if
+            needed. Such query will be automatically populated as 'query' field
+            for bookmark
         """
         # not using separate navigator step here not to have to register
         # navigate step for every single entity
         view = self.navigate_to(self, 'All')
-        if not search_query:
-            search_query = values.get('query')
-        if not hasattr(view, 'search'):
+        if not hasattr(view, 'searchbox'):
             raise KeyError(
                 '{} does not have searchbox'.format(self.__class__.__name__))
-        view.search(search_query)
         if not view.searchbox.actions.is_displayed:
             raise NoSuchElementException(
                 'Unable to create a bookmark - {} has a searchbox with no'
                 'actions dropdown'
                 .format(self.__class__.__name__)
             )
+        if search_query:
+            view.searchbox.search_field.fill(search_query)
         view.searchbox.actions.fill('Bookmark this search')
         view = BookmarkCreateView(self.browser)
         view.fill(values)
--- a/airgun/entities/base.py
+++ b/airgun/entities/base.py
@@ -1,3 +1,6 @@
+from widgetastic.exceptions import NoSuchElementException
+
+from airgun.views.common import BookmarkCreateView
 
 
 class BaseEntity(object):
@@ -6,3 +9,32 @@ class BaseEntity(object):
         self.browser = browser
         self.session = browser.extra_objects['session']
         self.navigate_to = self.session.navigator.navigate
+
+    def create_bookmark(self, values, search_query=None):
+        """Create a bookmark.
+
+        :param dict values: dictionary with keys 'name', 'query', 'public'
+        :param str optional search_query: if something different from `query`
+            should be typed into searchbox
+        """
+        # not using separate navigator step here not to have to register
+        # navigate step for every single entity
+        view = self.navigate_to(self, 'All')
+        if not search_query:
+            search_query = values.get('query')
+        if not hasattr(view, 'search'):
+            raise KeyError(
+                '{} does not have searchbox'.format(self.__class__.__name__))
+        view.search(search_query)
+        if not view.searchbox.actions.is_displayed:
+            raise NoSuchElementException(
+                'Unable to create a bookmark - {} has a searchbox with no'
+                'actions dropdown'
+                .format(self.__class__.__name__)
+            )
+        view.searchbox.actions.fill('Bookmark this search')
+        view = BookmarkCreateView(self.browser)
+        view.fill(values)
+        view.submit.click()
+        view.flash.assert_no_error()
+        view.flash.dismiss()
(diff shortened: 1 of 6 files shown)


airgun/entities/base.py
```diff
@@ -6,3 +9,32 @@ def __init__(self, browser):
...
+        """Create a bookmark.
+
+        :param dict values: dictionary with keys 'name', 'query', 'public'
+        :param str optional search_query: if something different from `query`
+            should be typed into searchbox
+        """
+        # not using separate navigator step here not to have to register
+        # navigate step for every single entity
+        view = self.navigate_to(self, 'All')
+        if not search_query:
```
oshtaier: doing
```
        if not search_query:
            search_query = values.get('query')
        view.search(search_query)
```
you completely dismiss the whole point of `search_query` and `query` parameters difference.

You whether should permit next scenario:
1) I fill search query and then query is filled automatically
2) I fill nothing for search query and fill query manually

or not:
1) I can fill only search query. Then it doesn't make sense to have a parameter and we should `query` directly into `search_query`

airgun/entities/base.py
```diff
@@ -6,3 +9,32 @@ def __init__(self, browser):
...
+            search_query = values.get('query')
+        if not hasattr(view, 'search'):
+            raise KeyError(
+                '{} does not have searchbox'.format(self.__class__.__name__))
+        view.search(search_query)
```
oshtaier: Why do we need to do search? and not just input value?

airgun/entities/bookmark.py
```diff
@@ -0,0 +1,81 @@
...
+
+class BookmarkEntity(BaseEntity):
+
+    # Note: creation procedure takes place on specific entity page, generic
+    # helper is implemented inside :class:`BaseEntity`.
+
+    def delete(self, entity_name, controller=None):
+        """Delete existing bookmark"""
+        view = self.navigate_to(self, 'All')
+        row_query = {'name': entity_name}
```
oshtaier: That looks more readable for me, but it is just my opinion
```
query = entity_name
if controller:
    query = 'name = "{}" and controller = "{}"'.format(
        entity_name, controller)
view.search(query)
row_query = {'name': entity_name}
if controller:
    row_query['controller'] = controller
view.table.row(**row_query)['Actions'].widget.click(handle_alert=True)
```

airgun/entities/bookmark.py
```diff
@@ -0,0 +1,81 @@
...
+    VIEW = BookmarkEditView
+
+    def prerequisite(self, *args, **kwargs):
+        return self.navigate_to(self.obj, 'All')
+
+    def step(self, *args, **kwargs):
+        entity_name = kwargs.get('entity_name')
+        controller = kwargs.get('controller')
+        row_query = {'name': entity_name}
+        if not controller:
```
oshtaier: hmm, it looks too much duplicated, maybe we change our search method?

airgun/entities/bookmark.py
ldjebran: more over :)
```python
row_query = {'name': entity_name}
query = 'name = "{}"'.format(entity_name)
if controller:
    query = '{} and controller = "{}"'.format(query, controller)
    row_query['controller'] = controller

view.search(query)
view.table.row(**row_query)['Actions'].widget.click(handle_alert=True)
```

airgun/entities/bookmark.py
ldjebran: Think we should keep the search with a string query but we can create a function in entity that returns a tuple  query, raw_query

airgun/entities/base.py
abalakh: Thought we need to perform search so query can be automatically filled. Turns out we don't need to do that, just typing into search field is enough. Will update, thanks for suggestion.

airgun/entities/bookmark.py
abalakh: @oshtaier can't agree 2 exactly the same conditions are better than one.
@ldjebran i was thinking the same way, thought rewriting the same var isn't that optimal approach. But maybe more readable, will update, thanks for suggestion.

airgun/entities/base.py
abalakh: Both use cases will work, although for second one searchbox will be filled anyways which is slight redundancy.
To be clear, are you suggesting
```python
if search_query:
    view.searchbox.search_field.fill(search_query)
```
instead?

airgun/entities/bookmark.py
abalakh: @oshtaier i'd like to, but changing search method won't really help us as we can't use it inside navigate step as it uses navigation itself.
@ldjebran not sure if i got your idea right, at least can't image how i can implement that it won't look ugly and overcomplicated. Any more details or example, please?

airgun/entities/base.py
oshtaier: Nope, both scenario can't work as they mutually exclusive, but I thought and read more your code. And yes, exacly:
```
if search_query:
    view.searchbox.search_field.fill(search_query)
```

airgun/entities/bookmark.py
ldjebran: That will be fine even OS dev :)

airgun/entities/bookmark.py
ldjebran: at module level
```python
def get_entity_query_data(entity_name, controller=None):
    row_query = {'name': entity_name}
    query = 'name = "{}"'.format(entity_name)
    if controller:
        query = '{} and controller = "{}"'.format(query, controller)
        row_query['controller'] = controller
    return query, raw_query
```
and use this function in entity and here
```python
query, raw_query = get_entity_query_data(entity_name, controller=controller)
```
This is only a suggestion to not repeat code.

abalakh:
```python
pytest -v tests/foreman/ui_airgun/test_bookmark.py
platform darwin -- Python 3.6.4, pytest-3.6.1, py-1.6.0, pluggy-0.6.0 -- /Users/andrii/workspace/py3a/bin/python3.6
cachedir: .pytest_cache
plugins: xdist-1.22.0, services-1.2.1, mock-1.10.0, forked-0.2, env-0.6.2
collecting 1 item                                                              2018-10-12 15:57:06 - conftest - DEBUG - BZ deselect is disabled in settings

collected 1 item

tests/foreman/ui_airgun/test_bookmark.py::test_positive_create_bookmark_public PASSED [100%]

========================== 1 passed in 157.30 seconds ==========================
```

PR #225, opened 2018-10-17
vm_import + hostgroup

lhellebr:
Add vm_import to computeresource and hostgroup entity only with functionality required for compute resource testing (deploy_on can't be set using API).

### DIFF
--- a/airgun/entities/computeresource.py
+++ b/airgun/entities/computeresource.py
@@ -8,6 +8,7 @@ from airgun.views.computeresource import (
     ResourceProviderEditView,
     ResourceProviderDetailView,
     ResourceProviderProfileView,
+    ResourceProviderVMImport,
 )
 
 
@@ -74,6 +75,13 @@ class ComputeResourceEntity(BaseEntity):
         vm = self.list_vms(rhev_name, vm_name)
         if vm['Power'].widget.read() == 'On':
             vm['Actions'].widget.click(handle_alert=True)
+
+    def vm_import(self, entity_name, vm_name, hostgroup, location):
+        """Imports the specified VM"""
+        view = self.navigate_to(self, 'VMImport',
+                                entity_name=entity_name, vm_name=vm_name)
+        view.fill({'host.hostgroup': hostgroup, 'host.location': location})
+        view.submit.click()
 
     def update_computeprofile(self, computeresource, computeprofile, values):
         """Update specific compute profile attributes through CR detail view"""
@@ -153,3 +161,17 @@ class ComputeResourceProfileDetail(NavigateStep):
         entity_name = kwargs.get('entity_name')
         self.parent.compute_profiles.table.row(compute_profile=entity_name)[
             'Compute profile'].widget.click()
+
+
+@navigator.register(ComputeResourceEntity, 'VMImport')
+class ComputeResourceVMImport(NavigateStep):
+    VIEW = ResourceProviderVMImport
+
+    def prerequisite(self, *args, **kwargs):
+        entity_name = kwargs.get('entity_name')
+        return self.navigate_to(self.obj, 'Detail', rhev_name=entity_name)
+
+    def step(self, *args, **kwargs):
+        vm_name = kwargs.get('vm_name')
+        self.parent.virtual_machines.search(vm_name)
+        self.parent.virtual_machines.actions.fill("Import")
--- a/airgun/views/computeresource.py
+++ b/airgun/views/computeresource.py
@@ -18,6 +18,7 @@ from airgun.widgets import (
     MultiSelect,
 )
 from widgetastic_patternfly import BreadCrumb
+from airgun.views.host import HostCreateView
 
 
 class ComputeResourcesView(BaseLoggedInView, SearchableViewMixin):
@@ -178,6 +179,7 @@ class ResourceProviderDetailView(BaseLoggedInView):
     class virtual_machines(SatTab, SearchableViewMixin):
         TAB_NAME = 'Virtual Machines'
 
+        actions = ActionsDropdown("//div[contains(@class, 'btn-group')]")
         table = SatTable(
             './/table',
             column_widgets={
@@ -225,4 +227,18 @@ class ResourceProviderProfileView(BaseLoggedInView):
             and self.breadcrumb.locations[0] == 'Compute resources'
             and self.breadcrumb.locations[2] == 'Compute profiles'
             and self.breadcrumb.read().startswith('Edit ')
+        )
+
+
+class ResourceProviderVMImport(HostCreateView):
+    breadcrumb = BreadCrumb()
+
+    @property
+    def is_displayed(self):
+        breadcrumb_loaded = self.browser.wait_for_element(
+            self.breadcrumb, exception=False)
+        return (
+            breadcrumb_loaded
+            and self.breadcrumb.locations[0] == 'Compute Resources Vms'
+            and self.breadcrumb.read().startswith('Import ')
         )


airgun/views/computeresource.py
```diff
@@ -174,6 +178,8 @@ class virtual_machines(SatTab, SearchableViewMixin):
                 'Power': Text('.//span[contains(@class,"label")]'),
             }
         )
+        more_button = Text('//span[@class="caret"]')
```
oshtaier: You can try different solutions here:
1) Use `ActionsDropdown` widget directly
2) Inherit `ActionsDropdown` class for your own custom class implemented here locally and override `dropdown` or `button` elements or whatever you need
3) If neither solution works for you - implement your own widget in that module that will work with dropdown element as with dropdown, not separate buttons and etc

airgun/views/computeresource.py
```diff
@@ -174,6 +178,8 @@ class virtual_machines(SatTab, SearchableViewMixin):
...
+        import_button = Text('//a[text()="Import"]')
```
oshtaier: One more question:
Are you sure that all compute providers ('ec2', 'docker', 'rhev' and etc) have these elements? Otherwise that code will add more regression than value

airgun/views/computeresource.py
```diff
@@ -129,6 +129,10 @@ def is_displayed(self):
         return self.browser.wait_for_element(
             self.name, exception=False) is not None
 
+    @View.nested
```
oshtaier: add same for organizations

airgun/entities/computeresource.py
```diff
@@ -75,6 +76,17 @@ def vm_poweroff(self, rhev_name, vm_name):
         if vm['Power'].widget.read() == 'On':
             vm['Actions'].widget.click(handle_alert=True)
 
+    def vm_import(self, rhev_name, vm_name, hostgroup, location):
+        """Imports the specified VM"""
+        view = self.navigate_to(self, 'Detail', rhev_name=rhev_name)
+        view.virtual_machines.search(vm_name)
+        view.virtual_machines.more_button.click()
```
oshtaier: change
```
        view.virtual_machines.more_button.click()
        view.virtual_machines.import_button.click()
```
as per comment in view to `fill` method for corresponding dropdown

airgun/entities/computeresource.py
```diff
@@ -75,6 +76,17 @@ def vm_poweroff(self, rhev_name, vm_name):
...
+        view.virtual_machines.import_button.click()
+        # we are already at import view
+        view = self.navigate_to(self, 'Import', rhev_name=rhev_name)
```
oshtaier: `self.navigate_to(HostEntity, 'Create')`

airgun/entities/computeresource.py
```diff
@@ -75,6 +76,17 @@ def vm_poweroff(self, rhev_name, vm_name):
...
+        view.fill({'host.hostgroup': hostgroup, 'host.location': location})
```
oshtaier: that is not responsibility of an entity layer. Always use:
`view.fill(values)`

airgun/entities/computeresource.py
```diff
@@ -153,3 +165,8 @@ def step(self, *args, **kwargs):
         entity_name = kwargs.get('entity_name')
         self.parent.compute_profiles.table.row(compute_profile=entity_name)[
             'Compute profile'].widget.click()
+
+
+@navigator.register(ComputeResourceEntity, 'Import')
```
oshtaier: please remove:
```
@navigator.register(ComputeResourceEntity, 'Import')
class Import(NavigateStep):
    VIEW = HostCreateView
```

airgun/entities/hostgroup.py
```diff
@@ -0,0 +1,42 @@
...
+from airgun.navigation import NavigateStep, navigator
+from airgun.views.hostgroup import (
+    HostgroupView,
+    HostgroupEditView,
+)
+
+
+class HostgroupEntity(BaseEntity):
+
+    def set_computeresource(self, hostgroup_name, compute_resource):
```
oshtaier: sounds as completely redundant method. Implement basic `update` method instead

airgun/views/hostgroup.py
```diff
@@ -0,0 +1,42 @@
...
+
+from airgun.views.common import BaseLoggedInView, SearchableViewMixin
+from airgun.widgets import (
+    ActionsDropdown,
+    FilteredDropdown,
+    SatTable,
+)
+
+
+class HostgroupView(BaseLoggedInView, SearchableViewMixin):
```
oshtaier: in general - we do not work like this. If you start your coding for entity, so please implement it completely, not like few buttons you need

airgun/views/hostgroup.py
```diff
@@ -0,0 +1,42 @@
...
+
+    @property
+    def is_displayed(self):
+        return self.browser.wait_for_element(
+            self.title, exception=False) is not None
+
+
+class HostgroupEditView(BaseLoggedInView):
+    breadcrumb = BreadCrumb()
+    title = Text("//span[contains(text(),'Edit')]")
```
oshtaier: remove that element

airgun/views/hostgroup.py
```diff
@@ -0,0 +1,42 @@
...
+    compute_resource = FilteredDropdown(id='hostgroup_compute_resource')
```
oshtaier: I don't see compute resource for default host group:
image

Do we have special conditions for that?

lhellebr: Compute resource == Deploy on

airgun/views/hostgroup.py
lhellebr: Done

airgun/views/hostgroup.py
lhellebr: Don't forget Airgun is here to make Robottelo development easier. And contrary to what you say, there are many missing features in Airgun, even in already implemented entities. Developing UI tests in Robottelo means constantly finding a way to do something while API doesn't support many things and so does not Airgun.

My workflow is that I need to create a certain test and need tools for that - if they do not exist, I have to fix the existing ones. My workflow, contrary to Airgun developer's, is not creating a framework and then writing test based on what is implemented in the framework. I need to write very specific tests and as a test writer, I only do the changes that are necessary in framework.

As an Airgun developer, I'm planning to implement the rest of hostgroup entity later when I have time for that. Right now, however, it is out of scope of this PR. If you need the entity fully implemented immediately, feel free to contribute.

airgun/entities/hostgroup.py
lhellebr: I will implement the basic update method once there is complete view implemented, together with the rest of CRUD methods. I am planning to do that later, see my longest comment in this PR.

airgun/entities/computeresource.py
lhellebr: I am using an unnamed dictionary rather than putting it into a variable first. What's the difference?

airgun/entities/computeresource.py
lhellebr: Done. Cool, I wasn't sure if this is a good idea.

airgun/views/computeresource.py
lhellebr: Done, for the sake of consistency

airgun/entities/computeresource.py
oshtaier: in your implementation you will be able to fill only two fields, in mine hundreds

airgun/entities/computeresource.py
lhellebr: Done

airgun/views/hostgroup.py
jhutar: Hello @oshtaier . IMO if you need 2 users, you first have to have 1 user. Yes, this is no full HG support in AirGun, yes issue describing what needs to be added to have it complete would be nice (but not required, it can be created by anybody). I do not see a reason why partial HG support could not be committed?

airgun/entities/hostgroup.py
jhutar: Hello @oshtaier . [Contribution info](https://airgun.readthedocs.io/en/latest/#contributing) does not mention anything like this. Or am I missing something?

lhellebr: I have changed the method to Update so we do not need to change the interface later.

airgun/views/hostgroup.py
lhellebr: @oshtaier , as I said, this doesn't make sense from the test writer point of view. We need to implement tests that test real scenarios. Not to implement tests based on what Airgun can do. Not to implement considerable parts of the framework so we can use some minor functionality. That's not what we are here for.
I've fixed some of the real issues you have pointed out in the PR. Please, review and merge this if there are no further issues.

abalakh: We can't force people to fully implement entities and views if they need only small part of it, as it's out of scope and other parts of entity may appear complex or require some extra knowledge on how to work with them, so i agree with @lhellebr and @jhutar here. But only in case that partial implementation doesn't bring any redundant/temporary code which will add extra work to update/remove it in future.
The only thing we should probably think of is how to mark such entities as "WIP" ones, so others won't get confused why some parts of functionality doesn't work for them.

airgun/entities/hostgroup.py
abalakh: And here i fully agree with @oshtaier - like i said in https://github.com/SatelliteQE/airgun/pull/225#discussion_r235436500, partial implementation is ok, but extra temporary methods which will become useless and redundant with full implementation is definitely not ok.
Thanks @lhellebr for addressing it.

airgun/entities/computeresource.py
```diff
@@ -75,6 +76,13 @@ def vm_poweroff(self, rhev_name, vm_name):
         if vm['Power'].widget.read() == 'On':
             vm['Actions'].widget.click(handle_alert=True)
 
+    def vm_import(self, cr_name, vm_name, hostgroup, location):
```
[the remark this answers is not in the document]
lhellebr: Thanks for pointing this out, I've fixed it for "cr_name".
Because "rhev_name" is used incorrectly in this sense in multiple places unrelated to this PR, I'll fix that in another PR.

PR #227, opened 2018-10-23
Fix subscription waiter helper

abalakh:
Test results in SatelliteQE/robottelo#6424

### DIFF
--- a/airgun/entities/subscription.py
+++ b/airgun/entities/subscription.py
@@ -22,7 +22,7 @@ class SubscriptionEntity(BaseEntity):
         :param name: Name of running task
         :param has_manifest: Should manifest exist after task ended?
         """
-        view = self.navigate_to(self, 'All')
+        view = SubscriptionListView(self.browser, logger=self.browser.logger)
         wait_for(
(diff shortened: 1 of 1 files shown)


airgun/entities/subscription.py
```diff
@@ -22,7 +22,7 @@ def _wait_for_process_to_finish(self, name, has_manifest=False):
         :param name: Name of running task
         :param has_manifest: Should manifest exist after task ended?
         """
-        view = self.navigate_to(self, 'All')
+        view = SubscriptionListView(self.browser, logger=self.browser.logger)
```
abalakh: Navigate triggers `pre_navigate()`, which ensures fade widget is gone before proceeding. The thing is that fade is appearing when task is started and is usually gone when task is finished (pretty often *after* progress bar disappeared and flash notifications shown), so we can't wait till fade is gone as the next verifications may fail.

oshtaier: hmm, so why do we need complex wait mechanism if we apply that 'not so elegant' workaround anyway?

PR #229, opened 2018-10-23
Add more functionality to sync plans

oshtaier:


airgun/entities/syncplan.py
```diff
@@ -12,37 +12,51 @@
...
 
     def update(self, entity_name, values):
+        """Update sync plan with necessary values"""
         view = self.navigate_to(self, 'Edit', entity_name=entity_name)
         filled_values = view.fill(values)
         view.flash.assert_no_error()
         view.flash.dismiss()
         return filled_values
 
+    def add_product(self, entity_name, products_list):
```
abalakh: `add_product` but `products_list`, that's a bit confusing.

abalakh: If no better alternatives - at least add a line in docstring explaining either single product or list is accepted.

PR #232, opened 2018-10-30
Add host group entity

oshtaier:


airgun/views/hostgroup.py
```diff
@@ -0,0 +1,124 @@
...
+
+    @View.nested
+    class ansible_roles(SatTab):
+        TAB_NAME = 'Ansible Roles'
+        resources = MultiSelect(id='ms-hostgroup_ansible_role_ids')
+
+    @View.nested
+    class puppet_classes(SatTab):
+        TAB_NAME = 'Puppet Classes'
+        pass
```
ldjebran: to remove

airgun/views/hostgroup.py
```diff
@@ -0,0 +1,124 @@
...
+        operating_system = FilteredDropdown(id='hostgroup_operatingsystem')
+        media_type = RadioGroup(
+            locator="//div[label[contains(., 'Media Selection')]]")
+        media = FilteredDropdown(id='hostgroup_medium')
+        ptable = FilteredDropdown(id='hostgroup_ptable')
+        root_password = TextInput(id='hostgroup_root_pass')
+
+    @View.nested
+    class parameters(SatTab):
+        pass
```
ldjebran: can we add tab name as for puppet classes?
perhaps make a TODO comment say:
```
# TODO to implement
```

airgun/views/hostgroup.py
ldjebran: perhaps put a TODO comment say:
```
# TODO to implement
```

ldjebran: OK than no need to break the rule than

PR #243, opened 2018-11-20
Fixes subscriptions and contenthosts entities

ldjebran:


airgun/entities/subscription.py
```diff
@@ -71,7 +64,12 @@ def read_delete_manifest_message(self):
         """Read message displayed on 'Confirm delete manifest' dialog"""
         view = self.navigate_to(self, 'Delete Manifest Confirmation')
         view.wait_animation_end()
-        return view.message.read()
+        delete_message = view.message.read()
+        view.cancel_button.click()
+        manage_view = ManageManifestView(self.browser)
+        if manage_view.is_displayed:
+            manage_view.close_button.click()
```
abalakh: imo some comment explaining why do we need that would be helpful (if i got you right - you're closing modal window to unblock further auto steps, aren't you?)

airgun/entities/subscription.py
```diff
@@ -142,6 +141,15 @@ def pre_navigate(self, _tries, *args, **kwargs):
 
     def step(self, *args, **kwargs):
         self.view.menu.select('Content', 'Subscriptions')
+        wait_for_spinner(self.view)
+        self.parent.browser.plugin.ensure_page_safe()
+        wait_for(
+            lambda: self.am_i_here(*args, **kwargs),
```
abalakh: any other options? i don't get why we aren't on requested page after all the ajax finished and spinner is gone

airgun/entities/subscription.py
```diff
@@ -195,17 +203,32 @@ class SubscriptionDetails(NavigateStep):
...
+        virt_who: Whether this is a virt who client subscription.
     """
     VIEW = SubscriptionDetailsView
 
     def prerequisite(self, *args, **kwargs):
         return self.navigate_to(self.obj, 'All')
 
     def am_i_here(self, *args, **kwargs):
-        subscription_name = kwargs.get('subscription')
+        subscription_name = kwargs.get('entity_name')
```
abalakh: is it still useful? By default we have am_i_here which checks for entity_name in breadcrumb. The only difference is default check is `self.view.breadcrumb.locations[1] == entity_name` but here it's `self.view.breadcrumb.locations[-1] == entity_name`.  Is the full path longer than 2 nodes?

airgun/views/subscription.py
```diff
@@ -21,6 +21,16 @@
 )
 
 
+def wait_for_spinner(widget):
```
abalakh: We have pretty similar helper for rhrepo. Maybe it's time to move it out somewhere? I think there might be more new style pages (or more upcoming).

airgun/views/subscription.py
```diff
@@ -186,14 +202,21 @@ class SubscriptionDetailsView(BaseLoggedInView):
     @View.nested
     class details(SatTab):
 
+        associations = SatTable(
```
abalakh: Are you sure 'associations' is better than our standard 'table'? Or maybe there're other tables and no search for sure? Pls explain your choice.

airgun/views/subscription.py
```diff
@@ -21,6 +21,16 @@
 )
 
 
+def wait_for_spinner(widget):
+    """Wait for any spinner to disappear"""
+    wait_for(
+        lambda: not widget.browser.elements("//div[contains(@class, 'spinner')]"),
```
abalakh: Oh, basically the key question - since we have multiple pages with such spinner, maybe it's not worth separate helper which has to be manually called each time but rather worth adding new check to ensure_page_safe? That would ensure spinner isn't present in all possible cases (way more than covered ones) and won't require extra lines. Thoughts?

airgun/entities/subscription.py
oshtaier: +1

airgun/entities/subscription.py
```diff
@@ -195,17 +203,32 @@ class SubscriptionDetails(NavigateStep):
...
         subscription_name = kwargs.get('entity_name')
+        virt_who = kwargs.get('virt_who')
+        search_string = 'name = "{0}"'.format(subscription_name)
+        if virt_who:
+            search_string = 'virt_who = true and {0}'.format(search_string)
+        self.parent.search(search_string)
         self.parent.table.row(name=subscription_name)['Name'].widget.click()
+        wait_for_spinner(self.view)
+        self.parent.browser.plugin.ensure_page_safe()
+        wait_for(
```
oshtaier: +1 here
we have too many methods that validate our navigation is done. Can we have something more generic?

airgun/views/subscription.py
```diff
@@ -104,6 +114,12 @@ def is_displayed(self):
         return self.browser.wait_for_element(
             'div#subscriptions-table', timeout=10, exception=False) is not None
 
+    def search(self, query):
+        """Customized search to make sure that the table spinner is hidden, after search"""
+        self.searchbox.search(query)
+        wait_for_spinner(self)
```
oshtaier: same here. We start to add special methods just for waiter. We need to move that logic on a level above in our hierarchy

airgun/entities/subscription.py
```diff
@@ -55,9 +50,7 @@ def add_manifest(self, manifest_file):
         """
         view = self.navigate_to(self, 'Manage Manifest')
         view.wait_animation_end()
-        view.fill({
-            'manifest.manifest_file': manifest_file,
-        })
+        view.manifest.manifest_file.fill(manifest_file)
```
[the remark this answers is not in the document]
ldjebran: This related to the fact that when filling this field the dialog form disappear and manifest loading begin, this is not how view filing method works, if we let it as the fill method try to access other widgets in the view that lead to exception, to resolve this behavior we two ways,   
1. Do it as I realized here 
2. Inherit fill view method and verify if manifest file is in values, if yes fill it and exit ignoring all other fields otherwise continue with base method.

I chosen the simplest way, as we already know that only that field should be involved and not inherit fill method  :)

airgun/entities/subscription.py
ldjebran: If we do not close that dialog, The further navigation will be broken, especially when we want to navigate to subscription page again  as it will find that we are at that page already and fail.

airgun/entities/subscription.py
ldjebran: The page make some delays to load and yes after all that, without the waiter it work 2 time but fail 3 other times, any way I do see any inconvenient here to put that waiter as we here clicked that button and we already know what we are expecting, with this waiter the things are very stable eg: if we do not do it better to  register a bug about this and put on hold that issue.

airgun/entities/subscription.py
ldjebran: the full path is exactly 2 nodes for example :
Subscriptions > Red Hat Enterprise Linux Server, Premium (8 sockets) (Unlimited guests)

do not understand here what you mean

ldjebran: I simply changed that as we are passing entity_name and not subscription

airgun/views/subscription.py
ldjebran: I really do not understand what you mean , but associations here is a table in subscription details page:
screenshot from 2018-11-22 11-33-48-sub-details

airgun/entities/subscription.py
abalakh: Sure. I meant comment in the code :)

airgun/entities/subscription.py
abalakh: Please drop this method completely then. Take a look at airgun/navigation.py::NavigateStep.am_i_here() - it already does the same; the reason this method was implemented - custom name `subscription` instead of generic `entity_name`. But as you've changed the name to default one, default navigatestep's am_i_here should work great.

airgun/views/subscription.py
abalakh: Are there others spinners on rhrepo page, can't we just use the same locator?

airgun/views/subscription.py
abalakh: Ok, great, i thought it was regular table with entities in it. For such table custom name makes sense, thx.

airgun/views/subscription.py
abalakh: @oshtaier and your thoughts, pls?

airgun/entities/subscription.py
abalakh: If there's a bug - better to submit it, but let's not put PR on hold - fixing that bug may take months as it's not critical.

airgun/views/subscription.py
oshtaier: It depends on the impact on performance. If we talk about milliseconds for that validation, so it is not a problem. Anyway, can we override `ensure_page_safe` or similar idea just for that module? I meant, I don't like that we have more than one (actually at least 3) methods to wait for different things. It is just bad style

airgun/views/subscription.py
ldjebran: nope, I wanted to replay and find my self repeating what I said above :)

ldjebran:
@abalakh @oshtaier comments addressed test output https://github.com/SatelliteQE/robottelo/pull/6483#issuecomment-441185587.

ldjebran:
Main changes:
1- The spinner waiter has been moved to browser plugin to used in rh repos (for a future fix). 
2- The waiters has been moved to post_step navigation (by using the new approach) in a class that is inherited by the navigation clsses that need it

oshtaier:
Looks much better.
@ldjebran @abalakh 
1) Didn't expect that waiter to be moved so high. Could we override `ensure_page_safe` on class/module/entity level?
2) Can we override it on entity level, so we should not call it explicitly in scope of any method in that class? Like no matter I do for subscriptions, it always call overridden, not basic validation
3) Can we put that post_navigation idea with `am_I_here` on higher level hierarchy to be used across whole framework? Or it is too major change?

ldjebran:
@oshtaier @abalakh changes done
experimented with 
```
item.style.display!="none" && item.style.visibility != "hidden" 
```
but seems not working with some entities.

The current implentation was tested also with test_positive_read_from_edit_page , test_positive_create_host_with_parameters and test architecture::test_positive_end_to_end
and seems working as expected
Note: test_positive_read_from_edit_page failed on last assert as seems current_user is not more readable from page and need to be fixed that is out of scope of this PR.


```
(robottelo) dlezz@dlezz:~/projects/robottelo-fork$ pytest -v tests/foreman/ui_airgun/test_subscription.py 
platform linux -- Python 3.6.6, pytest-3.6.1, py-1.5.4, pluggy-0.6.0 -- /home/dlezz/.pyenv/versions/3.6.6/envs/robottelo/bin/python
cachedir: .pytest_cache
plugins: wait-for-1.0.9, xdist-1.22.5, services-1.3.0, mock-1.10.0, forked-0.2, cov-2.5.1
collecting 5 items                                                                                           2018-11-23 18:12:05 - conftest - DEBUG - BZ deselect is disabled in settings

collected 5 items                                                                                            

tests/foreman/ui_airgun/test_subscription.py::test_positive_end_to_end PASSED                          [ 20%]
tests/foreman/ui_airgun/test_subscription.py::test_positive_access_with_non_admin_user_without_manifest PASSED [ 40%]
tests/foreman/ui_airgun/test_subscription.py::test_positive_access_with_non_admin_user_with_manifest SKIPPED [ 60%]
tests/foreman/ui_airgun/test_subscription.py::test_positive_view_vdc_subscription_products PASSED      [ 80%]
tests/foreman/ui_airgun/test_subscription.py::test_positive_view_vdc_guest_subscription_products PASSED [100%]

=================================== 4 passed, 1 skipped in 3421.67 seconds ===================================
```

airgun/views/subscription.py
```diff
@@ -104,6 +104,15 @@ def is_displayed(self):
         return self.browser.wait_for_element(
             'div#subscriptions-table', timeout=10, exception=False) is not None
 
+    def search(self, query):
```
abalakh: Is this search still needed? Looks pretty much similar to default one

ldjebran: search is needed, as when there is no manifest the entity is searchable but there is no table element and the base function fail in that table element not found.

ldjebran:
@abalakh @oshtaier changes done
```
(robottelo) dlezz@dlezz:~/projects/robottelo-fork$ pytest tests/foreman/ui_airgun/test_subscription.py -v -k "test_positive_end_to_end or test_positive_access_with_non_admin_user_without_manifest"
platform linux -- Python 3.6.6, pytest-4.0.1, py-1.5.4, pluggy-0.8.0 -- /home/dlezz/.pyenv/versions/robottelo/bin/python3.6
cachedir: .pytest_cache
plugins: wait-for-1.0.9, xdist-1.22.5, services-1.3.0, mock-1.10.0, forked-0.2, cov-2.5.1
collecting ... 2018-12-04 15:29:32 - conftest - DEBUG - BZ deselect is disabled in settings

collected 5 items / 3 deselected                                                                             

tests/foreman/ui_airgun/test_subscription.py::test_positive_end_to_end PASSED                                                                 [ 50%]
tests/foreman/ui_airgun/test_subscription.py::test_positive_access_with_non_admin_user_without_manifest PASSED                                [100%]

=============================================== 2 passed, 3 deselected, 2 warnings in 251.24 seconds ================================================
```

PR #244, opened 2018-11-21
In the vertical menu, "Host Groups" has been changed to "Host Group"

lhellebr:


lhellebr:
I'm discussing it with developers, trying to actually find where the change originated. I agree this might have been an unintentional change.

lhellebr:
It's caused by the translation commit from 2016 and after 2 days, we still don't know how it only happened to 6.5. However, the menu has been reworked so that might be the reason.

I created a BZ #1652677 about this. There is an easy fix but we don't know when it's gonna be in compose. I am pro keeping this my commit for now as not to block tests. I will revert it once the BZ is verified.

Issue #19, opened 2018-02-22
Implement delete entity mixin


abalakh:
Similar to [robottelo.ui.Base.delete](https://github.com/SatelliteQE/robottelo/blob/master/robottelo/ui/base.py#L199) we should create a helper method which will handle entity deletion and ensure deleted entity can't be found using search.
Couple of improvements which would be nice to have (thanks @oshtaier for pointing them):
- cover the case when deleted entity was the last one available, so searchbar is not present anymore on entity list page
- `delete` should return True/False whether deletion succeeded so we can easily assert deletion in tests like:
```python
assert session.architecture.delete('foo')
```
so we could mark the test as failed (not errored) in case element was found.

mfalesni:
I think using exceptions may be more beneficial, you just do:

```python
with some_expected_error_thing(FlashMessageException):
    create_bad_entity()
```

If a different error happens, it throws a different exception and that can be categorized.

PR #248, opened 2018-11-27
Fixed breadcrumb checks for Edit pages and few other improvements

### LINKED ISSUE #246: Typo in is_diplayed causing am_i_here() always return False
One of common checks inside view's is_dispayed property looks like this:
```python
self.breadcrumb.read().starts_with('Edit ')
```
But str's method is called `startswith`, with no underscore.
This causes exception, and corresponding am_i_here() always returns False.

abalakh:
* Fixed breadcrumb checks for Edit pages (fixes #246)
* Added manifest refresh, improvements for organization delete and subscription waiters

### DIFF
--- a/airgun/entities/organization.py
+++ b/airgun/entities/organization.py
@@ -1,4 +1,5 @@
 from navmazing import NavigateToSibling
+from wait_for import wait_for
 
 from airgun.entities.base import BaseEntity
 from airgun.navigation import NavigateStep, navigator
@@ -27,6 +28,13 @@ class OrganizationEntity(BaseEntity):
         self.browser.handle_alert()
         view.flash.assert_no_error()
         view.flash.dismiss()
+        view = self.navigate_to(self, 'All')
+        wait_for(
+         lambda: not view.search(entity_name),
+         timeout=120,
+         delay=2,
+         logger=view.logger
+        )
 
     def read(self, entity_name):
         view = self.navigate_to(self, 'Edit', entity_name=entity_name)
--- a/airgun/entities/organization.py
+++ b/airgun/entities/organization.py
@@ -30,10 +30,10 @@ class OrganizationEntity(BaseEntity):
         view.flash.dismiss()
         view = self.navigate_to(self, 'All')
         wait_for(
-         lambda: not view.search(entity_name),
-         timeout=120,
-         delay=2,
-         logger=view.logger
+            lambda: not view.search(entity_name),
+            timeout=120,
+            delay=2,
+            logger=view.logger
         )
 
     def read(self, entity_name):
@@ -92,10 +92,9 @@ class SelectOrganizationContext(NavigateStep):
 
     def am_i_here(self, *args, **kwargs):
         org_name = kwargs.get('org_name')
-        current_org = self.view.taxonomies.current_org()
         if len(org_name) > 30:
             org_name = org_name[:27] + '...'
-        return current_org == org_name
+        return org_name == self.view.taxonomies.current_org
 
     def step(self, *args, **kwargs):
         org_name = kwargs.get('org_name')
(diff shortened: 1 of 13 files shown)


airgun/widgets.py
```diff
@@ -691,6 +692,7 @@ def select_loc(self, loc_name):
         )
         self.browser.plugin.ensure_page_safe()
 
+    @property
```
oshtaier: should not you change https://github.com/SatelliteQE/airgun/blob/master/airgun/entities/location.py#L87 too?

airgun/widgets.py
```diff
@@ -676,6 +676,7 @@ def select_org(self, org_name):
...
```
oshtaier: should not you change https://github.com/SatelliteQE/airgun/blob/master/airgun/entities/organization.py#L87 too?

airgun/entities/organization.py
```diff
@@ -27,6 +28,13 @@ def delete(self, entity_name):
         self.browser.handle_alert()
         view.flash.assert_no_error()
         view.flash.dismiss()
+        view = self.navigate_to(self, 'All')
+        wait_for(
+         lambda: not view.search(entity_name),
```
oshtaier: one space indent?

airgun/entities/organization.py
```diff
@@ -27,6 +28,13 @@ def delete(self, entity_name):
         self.browser.handle_alert()
         view.flash.assert_no_error()
         view.flash.dismiss()
+        view = self.navigate_to(self, 'All')
```
oshtaier: that part become too common over the code. We need to think about post delete procedure. Maybe create an issue for that

airgun/entities/subscription.py
```diff
@@ -39,6 +39,8 @@ def _wait_for_process_to_finish(self, name, has_manifest=False):
                 handle_exception=True, timeout=10,
                 logger=view.logger
         )
+        # uncomment when BZ#1651981 is fixed
```
oshtaier: what is a reason for that change? Does someone approve that workaround?
If all cases for subscription fail, so bug is a blocker and only after confirmation you can do changes like this

airgun/widgets.py
abalakh: No idea how pycharm's search didn't reveal that to me. Big thanks for catching!

airgun/entities/organization.py
abalakh: Yeah, we already have #19 for that

airgun/entities/subscription.py
abalakh: I'm not sure it's a good idea to fail all the tests which are using subscriptions page when all the functionality being tested is working ok.

oshtaier: it is not okay, as user will see 404 error on the screen. We cannot omit that message. For WebUI that  issue is a blocker, so we cannot test functionality if we see anything like this. Except the case when everybody in team is informed and team lead is approve to skip that error

abalakh:
@oshtaier comments addressed.
>pending tests:
>few random for organization

```python
pytest -v tests/foreman/ui_airgun/test_organization.py -k 'test_positive_update_user or test_positive_update_environment or test_positive_update_location'
============================== test session starts ===============================
platform darwin -- Python 3.6.4, pytest-3.6.1, py-1.6.0, pluggy-0.6.0 -- /Users/andrii/workspace/py3a/bin/python3.6
cachedir: .pytest_cache
plugins: xdist-1.22.0, services-1.2.1, mock-1.10.0, forked-0.2, env-0.6.2
collecting 12 items                                                              2018-11-27 20:32:14 - conftest - DEBUG - BZ deselect is disabled in settings

collected 12 items / 9 deselected

tests/foreman/ui_airgun/test_organization.py::test_positive_update_user PASSED [ 33%]
tests/foreman/ui_airgun/test_organization.py::test_positive_update_environment PASSED [ 66%]
tests/foreman/ui_airgun/test_organization.py::test_positive_update_location PASSED [100%]

==================== 3 passed, 9 deselected in 253.99 seconds ====================
```

>and for that refresh manifest method.
>If you do them already - just link to them

https://github.com/SatelliteQE/robottelo/pull/6498#issuecomment-442162757

PR #253, opened 2018-12-06
Fixed issues with content view filter rule table

abalakh:
Test results:
```python
pytest -v tests/foreman/ui_airgun/test_contentview.py -k 'test_positive_add_package_filter or test_positive_update_inclusive_filter_package_version or test_positive_update_filter_affected_repos'
============================== test session starts ===============================
platform darwin -- Python 3.6.4, pytest-3.6.1, py-1.6.0, pluggy-0.6.0 -- /Users/andrii/workspace/py3a/bin/python3.6
cachedir: .pytest_cache
shared_function enabled - OFF - scope:  - storage: file
rootdir: /Users/andrii/workspace/robottelo, inifile:
plugins: xdist-1.22.0, services-1.2.1, mock-1.10.0, forked-0.2, env-0.6.2
collecting 38 items                                                              2018-12-06 12:48:48 - conftest - DEBUG - BZ deselect is disabled in settings

collected 38 items / 35 deselected

tests/foreman/ui_airgun/test_contentview.py::test_positive_add_package_filter PASSED [ 33%]
tests/foreman/ui_airgun/test_contentview.py::test_positive_update_inclusive_filter_package_version PASSED [ 66%]
tests/foreman/ui_airgun/test_contentview.py::test_positive_update_filter_affected_repos PASSED [100%]

=================== 3 passed, 35 deselected in 1028.01 seconds ===================
```

PR #254, opened 2018-12-07
Fixes compute resource entity and implement images and profiles

ldjebran:
tests output: https://github.com/SatelliteQE/robottelo/pull/6534

### DIFF
--- a/airgun/entities/computeresource.py
+++ b/airgun/entities/computeresource.py
@@ -260,7 +260,7 @@ class ComputeResourceVMImport(NavigateStep):
 
 class ComputeResourceImageProvider(NavigateStep):
     """Base class for image create and edit views, that need to dynamically define the view type
-    (that depend from from compute resource provider) before reaching navigation destination.
+    (that depend from compute resource provider) before reaching navigation destination.
     """
     PROVIDER_VIEWS = dict()
 
--- a/airgun/entities/computeresource.py
+++ b/airgun/entities/computeresource.py
@@ -160,7 +160,7 @@ class ComputeResourceEntity(BaseEntity):
         view = self.navigate_to(self, 'Detail', entity_name=entity_name)
         view.images.search(image_name)
         view.images.table.row(name=image_name)['Actions'].widget.fill('Destroy')
-        self.browser.handle_alert(wait=2.0)
+        self.browser.handle_alert()
         self.browser.plugin.ensure_page_safe()
         view.flash.assert_no_error()
         view.flash.dismiss()
@@ -191,16 +191,15 @@ class EditExistingComputeResource(NavigateStep):
     def prerequisite(self, *args, **kwargs):
         return self.navigate_to(self.obj, 'All')
 
-    def pre_navigate(self, _tries, *args, **kwargs):
-        # Force this view to navigate away in any case, as conflicts arise when editing after
-        # reading.
-        self.navigate_to(self.obj, 'All')
-
     def step(self, *args, **kwargs):
         entity_name = kwargs.get('entity_name')
         self.parent.search(entity_name)
         self.parent.table.row(
             name=entity_name)['Actions'].widget.fill('Edit')
+
+    def post_navigate(self, _tries, *args, **kwargs):
+        """Select Compute resource tab for initialization"""
+        self.view.compute_resource.click()
 
 
 @navigator.register(ComputeResourceEntity, 'Detail')
@@ -251,11 +250,12 @@ class ComputeResourceVMImport(NavigateStep):
         self.parent.virtual_machines.actions.fill("Import")
 
 
-class ComputeResourceImageProvider(NavigateStep):
-    """Base class for image create edit views, that need to dynamically define the view type
-    (that depend from from compute resource provider) before reaching navigation destination.
-    """
-    PROVIDER_VIEWS = dict()
+@navigator.register(ComputeResourceEntity, 'Create Image')
+class ComputeResourceImageCreate(NavigateStep):
+
+    PROVIDER_VIEWS = dict(
+        RHV=ComputeResourceRHVImageCreateView,
+    )
 
     def prerequisite(self, *args, **kwargs):
         entity_name = kwargs.get('entity_name')
@@ -268,29 +268,35 @@ class ComputeResourceImageProvider(NavigateStep):
         self.VIEW = view
         return parent_view
 
+    def step(self, *args, **kwargs):
+        self.parent.create_image.click()
+
     def am_i_here(self, *args, **kwargs):
         return self.view.is_displayed
 
 
-@navigator.register(ComputeResourceEntity, 'Create Image')
-class ComputeResourceImageCreate(ComputeResourceImageProvider):
-
-    PROVIDER_VIEWS = dict(
-        RHV=ComputeResourceRHVImageCreateView,
-    )
-
-    def step(self, *args, **kwargs):
-        self.parent.create_image.click()
-
-
 @navigator.register(ComputeResourceEntity, 'Edit Image')
-class ComputeResourceImageEdit(ComputeResourceImageProvider):
+class ComputeResourceImageEdit(NavigateStep):
 
     PROVIDER_VIEWS = dict(
         RHV=ComputeResourceRHVImageEditView,
     )
 
+    def prerequisite(self, *args, **kwargs):
+        entity_name = kwargs.get('entity_name')
+        parent_view = self.navigate_to(self.obj, 'Detail', entity_name=entity_name)
+        provider = parent_view.compute_resource.table.row(property='Provider')['Value'].read()
+        view = self.PROVIDER_VIEWS.get(provider)
+        if not view:
+            raise ValueError('Provider type "{0}" for class "{1}" not implemented'.format(
+                provider, self.__class__.__name__))
+        self.VIEW = view
+        return parent_view
+
     def step(self, *args, **kwargs):
         image_name = kwargs.get('image_name')
         self.parent.images.search(image_name)
         self.parent.images.table.row(name=image_name)['Actions'].widget.fill('Edit')
+
+    def am_i_here(self, *args, **kwargs):
+        return self.view.is_displayed
--- a/airgun/entities/computeresource.py
+++ b/airgun/entities/computeresource.py
@@ -201,6 +201,14 @@ class EditExistingComputeResource(NavigateStep):
         """Select Compute resource tab for initialization"""
         self.view.compute_resource.click()
 
+    def am_i_here(self, *args, **kwargs):
+        entity_name = kwargs.get('entity_name')
+        return (
+                self.view.is_displayed
+                and self.view.breadcrumb.locations[0] == 'Compute Resources'
+                and self.view.breadcrumb.read() == 'Edit {}'.format(entity_name)
+        )
+
 
 @navigator.register(ComputeResourceEntity, 'Detail')
 class ComputeResourceDetail(NavigateStep):
@@ -250,53 +258,46 @@ class ComputeResourceVMImport(NavigateStep):
         self.parent.virtual_machines.actions.fill("Import")
 
 
+class ComputeResourceImageProvider(NavigateStep):
+    """Base class for image create and edit views, that need to dynamically define the view type
+    (that depend from from compute resource provider) before reaching navigation destination.
+    """
+    PROVIDER_VIEWS = dict()
+
+    def prerequisite(self, *args, **kwargs):
+        entity_name = kwargs.get('entity_name')
+        parent_view = self.navigate_to(self.obj, 'Detail', entity_name=entity_name)
+        provider = parent_view.compute_resource.table.row(property='Provider')['Value'].read()
+        view = self.PROVIDER_VIEWS.get(provider)
+        if not view:
+            raise ValueError('Provider type "{0}" for class "{1}" not implemented'.format(
+                provider, self.__class__.__name__))
+        self.VIEW = view
+        return parent_view
+
+    def am_i_here(self, *args, **kwargs):
+        return self.view.is_displayed
+
+
 @navigator.register(ComputeResourceEntity, 'Create Image')
-class ComputeResourceImageCreate(NavigateStep):
+class ComputeResourceImageCreate(ComputeResourceImageProvider):
 
     PROVIDER_VIEWS = dict(
         RHV=ComputeResourceRHVImageCreateView,
     )
 
-    def prerequisite(self, *args, **kwargs):
-        entity_name = kwargs.get('entity_name')
-        parent_view = self.navigate_to(self.obj, 'Detail', entity_name=entity_name)
-        provider = parent_view.compute_resource.table.row(property='Provider')['Value'].read()
-        view = self.PROVIDER_VIEWS.get(provider)
-        if not view:
-            raise ValueError('Provider type "{0}" for class "{1}" not implemented'.format(
-                provider, self.__class__.__name__))
-        self.VIEW = view
-        return parent_view
-
     def step(self, *args, **kwargs):
         self.parent.create_image.click()
 
-    def am_i_here(self, *args, **kwargs):
-        return self.view.is_displayed
-
 
 @navigator.register(ComputeResourceEntity, 'Edit Image')
-class ComputeResourceImageEdit(NavigateStep):
+class ComputeResourceImageEdit(ComputeResourceImageProvider):
 
     PROVIDER_VIEWS = dict(
         RHV=ComputeResourceRHVImageEditView,
     )
 
-    def prerequisite(self, *args, **kwargs):
-        entity_name = kwargs.get('entity_name')
-        parent_view = self.navigate_to(self.obj, 'Detail', entity_name=entity_name)
-        provider = parent_view.compute_resource.table.row(property='Provider')['Value'].read()
-        view = self.PROVIDER_VIEWS.get(provider)
-        if not view:
-            raise ValueError('Provider type "{0}" for class "{1}" not implemented'.format(
-                provider, self.__class__.__name__))
-        self.VIEW = view
-        return parent_view
-
     def step(self, *args, **kwargs):
         image_name = kwargs.get('image_name')
         self.parent.images.search(image_name)
         self.parent.images.table.row(name=image_name)['Actions'].widget.fill('Edit')
-
-    def am_i_here(self, *args, **kwargs):
-        return self.view.is_displayed
--- a/airgun/entities/computeresource.py
+++ b/airgun/entities/computeresource.py
@@ -1,5 +1,5 @@
-
 from navmazing import NavigateToSibling
+
 from airgun.entities.base import BaseEntity
 from airgun.navigation import NavigateStep, navigator
 from airgun.views.computeresource import (
@@ -7,6 +7,8 @@ from airgun.views.computeresource import (
     ResourceProviderCreateView,
     ResourceProviderEditView,
     ResourceProviderDetailView,
+    ComputeResourceRHVImageCreateView,
+    ComputeResourceRHVImageEditView,
     ResourceProviderProfileView,
     ResourceProviderVMImport,
 )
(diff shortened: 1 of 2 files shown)


airgun/views/computeresource.py
```diff
@@ -151,22 +161,31 @@ def is_displayed(self):
             self.breadcrumb, exception=False)
         return (
             breadcrumb_loaded
-            and self.breadcrumb.locations[0] == 'Compute resources'
+            and self.breadcrumb.locations[0] == 'Compute Resources'
             and self.breadcrumb.read().startswith('Edit ')
         )
 
 
+class ComputeResourceTabTable(SatTable):
```
oshtaier: `SatTable` is deprecated now according to https://github.com/SatelliteQE/airgun/pull/253/files. Could you try to use `Table` widget directly?

airgun/views/computeresource.py
```diff
@@ -199,34 +218,207 @@ class compute_profiles(SatTab):
             }
         )
 
+    class images(SatTab, SearchableViewMixin):
```
oshtaier: same here. Could you try to use `Table` instead?

airgun/views/computeresource.py
```diff
@@ -199,34 +218,207 @@ class compute_profiles(SatTab):
...
+            return False
+        breadcrumb_text = self.breadcrumb.read()
         return (
-            breadcrumb_loaded
-            and self.breadcrumb.locations[0] == 'Compute resources'
-            and self.breadcrumb.locations[2] == 'Compute profiles'
-            and self.breadcrumb.read().startswith('Edit ')
+                self.breadcrumb.locations[0] == 'Compute Resources'
+                and self.breadcrumb.locations[2] == 'Compute Profiles'
+                and (breadcrumb_text.startswith('Edit ') or breadcrumb_text.startswith('New '))
```
oshtaier: any reason to have both views in one class? Usually we differentiate them

airgun/views/computeresource.py
```diff
@@ -199,34 +218,207 @@ class compute_profiles(SatTab):
...
+    def read(self):
+        """Read all items"""
+        return [item.read() for item in self.items]
+
+    def fill(self, values):
+        """Fill all items.
+        :param values: A list of values to fill the item widgets with.
+        """
+        if not values:
+            values = []
```
oshtaier: why not `return False`?

airgun/entities/computeresource.py
```diff
@@ -83,27 +84,87 @@ def vm_import(self, entity_name, vm_name, hostgroup, location):
...
+    def delete_image(self, entity_name, image_name):
+        """Delete compute resource image.
+
+        :param str entity_name: The compute resource name.
+        :param str image_name: The existing compute resource image name to delete.
+        """
+        view = self.navigate_to(self, 'Detail', entity_name=entity_name)
+        view.images.search(image_name)
+        view.images.table.row(name=image_name)['Actions'].widget.fill('Destroy')
+        self.browser.handle_alert(wait=2.0)
```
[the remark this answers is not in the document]
ldjebran: what do you mean here?

airgun/views/computeresource.py
ldjebran: This is SatTab and not SatTable

airgun/views/computeresource.py
ldjebran: if we need a way to clear the form, so fill(None) will clear the form,  return False, will do nothing here ,otherwise we can return False

airgun/views/computeresource.py
ldjebran: This is not my implementation, I only fixing here, but if you want can split them.

airgun/views/computeresource.py
ldjebran: Ok will test that with Table

airgun/entities/computeresource.py
```diff
@@ -145,10 +211,18 @@ def prerequisite(self, *args, **kwargs):
...
+        entity_name = kwargs.get('entity_name')
         self.parent.search(entity_name)
         self.parent.table.row(name=entity_name)['Name'].widget.click()
 
+    def am_i_here(self, *args, **kwargs):
+        entity_name = kwargs.get('entity_name')
+        return (
+                self.view.is_displayed
+                and self.view.breadcrumb.locations[0] == 'Compute Resources'
+                and self.view.breadcrumb.read() == entity_name
```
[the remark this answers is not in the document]
ldjebran: This is a details view it has no Edit
EG: Compute Resources > entity_name

airgun/entities/computeresource.py
```diff
@@ -169,9 +243,54 @@ class ComputeResourceVMImport(NavigateStep):
...
+        provider = parent_view.compute_resource.table.row(property='Provider')['Value'].read()
+        view = self.PROVIDER_VIEWS.get(provider)
+        if not view:
+            raise ValueError('Provider type "{0}" for class "{1}" not implemented'.format(
+                provider, self.__class__.__name__))
+        self.VIEW = view
+        return parent_view
+
+    def am_i_here(self, *args, **kwargs):
+        return self.view.is_displayed
```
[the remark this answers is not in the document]
ldjebran: The difference is here https://github.com/SatelliteQE/airgun/blob/master/airgun/navigation.py#L73
as we have entity_name and breadcrumb, but we are not at entity edit view

airgun/entities/computeresource.py
```diff
@@ -169,9 +243,54 @@ class ComputeResourceVMImport(NavigateStep):
...
-        return self.navigate_to(self.obj, 'Detail', rhev_name=entity_name)
+        return self.navigate_to(self.obj, 'Detail', entity_name=entity_name)
 
     def step(self, *args, **kwargs):
         vm_name = kwargs.get('vm_name')
         self.parent.virtual_machines.search(vm_name)
         self.parent.virtual_machines.actions.fill("Import")
+
+
+class ComputeResourceImageProvider(NavigateStep):
```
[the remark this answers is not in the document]
ldjebran: This a base class for create and edit navigation there to navigation classes that depend from this (this is done to not repeat all code for the two forms ComputeResourceImageCreate, and ComputeResourceImageEdit

airgun/views/computeresource.py
ldjebran: With table having this result when no rows found
```
[{'Name': 'Suggestions Check the javascript regular expression syntax of the search term. Check that the correct menu option is chosen (token ID vs. user ID). Use wildcards (* to match 0 or more characters, + to match 1 or more characters, ? to match 0 or 1 character). Clear the search field, then click Search to return to the 20 most recent records.', 'Operating System': 'Suggestions Check the javascript regular expression syntax of the search term. Check that the correct menu option is chosen (token ID vs. user ID). Use wildcards (* to match 0 or more characters, + to match 1 or more characters, ? to match 0 or 1 character). Clear the search field, then click Search to return to the 20 most recent records.', 'Username': 'Suggestions Check the javascript regular expression syntax of the search term. Check that the correct menu option is chosen (token ID vs. user ID). Use wildcards (* to match 0 or more characters, + to match 1 or more characters, ? to match 0 or 1 character). Clear the search field, then click Search to return to the 20 most recent records.', 'UUID': 'Suggestions Check the javascript regular expression syntax of the search term. Check that the correct menu option is chosen (token ID vs. user ID). Use wildcards (* to match 0 or more characters, + to match 1 or more characters, ? to match 0 or 1 character). Clear the search field, then click Search to return to the 20 most recent records.', 'User data': 'Suggestions Check the javascript regular expression syntax of the search term. Check that the correct menu option is chosen (token ID vs. user ID). Use wildcards (* to match 0 or more characters, + to match 1 or more characters, ? to match 0 or 1 character). Clear the search field, then click Search to return to the 20 most recent records.', 'Actions': 'Suggestions Check the javascript regular expression syntax of the search term. Check that the correct menu option is chosen (token ID vs. user ID). Use wildcards (* to match 0 or more characters, + to match 1 or more characters, ? to match 0 or 1 character). Clear the search field, then click Search to return to the 20 most recent records.'}]
```
screenshot from 2018-12-10 19-02-07

ldjebran: are we deprecating a working scenario in place of not working one?

airgun/views/computeresource.py
ldjebran: But this for is a bit tricky as we are not creating or editing, but associating

airgun/entities/computeresource.py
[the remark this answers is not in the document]
oshtaier:
```
in (
                        entity_name,
                        'Edit {}'.format(entity_name))
                )
```

airgun/entities/computeresource.py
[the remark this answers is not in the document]
oshtaier: Is `wait=2.0` mean wait 2 seconds? If yes then solution is bad

airgun/views/computeresource.py
oshtaier: oops

airgun/views/computeresource.py
oshtaier: If that is a thing you need then it is fine, but maybe you have to call `clear` when you want to clear? as that will make more sense

airgun/views/computeresource.py
oshtaier: yeah, we are moving to use solely `Table` in all places, so you can verify whether your `Name` is not equal to expected (when you delete entity for example) or that part of expected message (because it is too long) is present in `Name`

airgun/views/computeresource.py
oshtaier: it is not critical

airgun/entities/computeresource.py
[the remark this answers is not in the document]
oshtaier: I understand the idea, but suggest to use another approach as it is more readable and simple even if code is duplicated.

airgun/entities/computeresource.py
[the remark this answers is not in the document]
ldjebran: Ok, will remove that. but be aware that default wait is 30.0 seconds

airgun/views/computeresource.py
ldjebran: This is not that evident (This is not a first class view but embedded in an other view it's even not an entity view but a sub entity view and you have not direct access to that clear function) as the whole view is considered as a widget, as input box when you pass none it will be cleared first. if we had clear that list implicitly , we have to create two function in entity clear_rhv_image_interface_networks and clear_rhv_image_storage and this functions will be relevant only to RHV. It's more simple to clear them by passing a None or empty list.

airgun/entities/computeresource.py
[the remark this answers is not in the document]
ldjebran: By this the location of edit and details will be the same (am_i_here will be true for both locations) and we will be unable to navigate from edit view to details view and vice versa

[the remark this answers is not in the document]
oshtaier: It is already implemented in that way, but anyway, what is difference between edit and details views?

ldjebran:
@oshtaier 
changes done: testoutput https://github.com/SatelliteQE/robottelo/pull/6534#issuecomment-446162532

airgun/entities/computeresource.py
```diff
@@ -169,9 +242,61 @@ class ComputeResourceVMImport(NavigateStep):
...
+
+
+@navigator.register(ComputeResourceEntity, 'Create Image')
+class ComputeResourceImageCreate(NavigateStep):
+
+    PROVIDER_VIEWS = dict(
+        RHV=ComputeResourceRHVImageCreateView,
+    )
+
+    def prerequisite(self, *args, **kwargs):
```
oshtaier:
```
parent_view = self.navigate_to(self.obj, 'Detail', entity_name=entity_name)
self.VIEW = ComputeResourceRHVImageCreateView
```

airgun/entities/computeresource.py
```diff
@@ -169,9 +242,61 @@ class ComputeResourceVMImport(NavigateStep):
...
+
+
+@navigator.register(ComputeResourceEntity, 'Edit Image')
+class ComputeResourceImageEdit(NavigateStep):
+
+    PROVIDER_VIEWS = dict(
+        RHV=ComputeResourceRHVImageEditView,
+    )
+
+    def prerequisite(self, *args, **kwargs):
```
oshtaier:
```
```
parent_view = self.navigate_to(self.obj, 'Detail', entity_name=entity_name)
self.VIEW = ComputeResourceRHVImageEditView
return parent_view
```

File: airgun/entities/computeresource.py (hunk 9 above)
Comment (ldjebran) [reply airgun/entities/computeresource.py]: I do not agree with this change.

File: airgun/entities/computeresource.py (hunk 10 above)
Comment (ldjebran) [reply airgun/entities/computeresource.py]: I do not agree with this change.

**Comment by ldjebran** [comment]
do not agree with proposed changes, think has implemented a more general solution for new provider images EC2, Libvirt , VMware ...., we have only to put the provider with it's view in the dict to support it.

File: airgun/views/computeresource.py (hunk 1 above)
Comment (ldjebran) [reply airgun/views/computeresource.py]: class completely deleted and Table implemented

File: airgun/entities/computeresource.py (hunk 6 above)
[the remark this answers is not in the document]
Comment (ldjebran) [reply airgun/entities/computeresource.py]: Edit view:
Compute Resources > Edit bEXNJAxXBB
Details view 
Compute Resources > bEXNJAxXBB

File: airgun/entities/computeresource.py [hunk 11]
```diff
@@ -169,9 +243,54 @@ class ComputeResourceVMImport(NavigateStep):
...
+        return self.navigate_to(self.obj, 'Detail', entity_name=entity_name)
 
     def step(self, *args, **kwargs):
         vm_name = kwargs.get('vm_name')
         self.parent.virtual_machines.search(vm_name)
         self.parent.virtual_machines.actions.fill("Import")
+
+
+class ComputeResourceImageProvider(NavigateStep):
+    """Base class for image create edit views, that need to dynamically define the view type
```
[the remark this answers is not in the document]
ldjebran: deleted as not accepted by reviewer

airgun/entities/computeresource.py
```diff
@@ -169,9 +243,54 @@ class ComputeResourceVMImport(NavigateStep):
...
+    (that depend from from compute resource provider) before reaching navigation destination.
```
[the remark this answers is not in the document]
ldjebran: deleted as not accepted by reviewer

airgun/entities/computeresource.py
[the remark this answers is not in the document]
ldjebran: deleted as not accepted by reviewer

airgun/entities/computeresource.py
```diff
@@ -130,6 +191,11 @@ class EditExistingComputeResource(NavigateStep):
     def prerequisite(self, *args, **kwargs):
         return self.navigate_to(self.obj, 'All')
 
+    def pre_navigate(self, _tries, *args, **kwargs):
```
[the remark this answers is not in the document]
ldjebran: removed

airgun/entities/computeresource.py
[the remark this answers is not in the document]
ldjebran: removed

ldjebran:
also it make no sense to return RHV image  view for EC2 provider knowing that they are different and the things will crash later on locator not found.

airgun/entities/computeresource.py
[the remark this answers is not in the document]
oshtaier: so where is your view for Edit?

oshtaier:
okay, now I see your point. Sorry.
From that point of view, I am not sure what approach is better. I pushed my previous comments only to remove extra code, but if that code is generic, so no code is removed. If you like more your previous implementation of one generic class, you may return it. Or leave like this

ldjebran:
@oshtaier yeah the first implementation with generic super class was better, will restore it.

airgun/entities/computeresource.py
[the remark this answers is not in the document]
oshtaier: I mean, we do not have Edit View implemented? If it is going to be added in future then we can go with some implementation, but at that moment generic method will suffice our needs?

oshtaier:
Thanks

ldjebran:
@oshtaier changes done.
test output https://github.com/SatelliteQE/robottelo/pull/6534#issuecomment-446952641

airgun/views/computeresource.py
```diff
@@ -199,34 +215,207 @@ class compute_profiles(SatTab):
...
-        id='s2id_compute_attribute_vm_attrs_availability_zone')
-    subnet = FilteredDropdown(id='s2id_compute_attribute_vm_attrs_subnet_id')
-    security_groups = MultiSelect(
-        id='ms-compute_attribute_vm_attrs_security_group_ids')
-    managed_ip = FilteredDropdown(
-        id='s2id_compute_attribute_vm_attrs_managed_ip')
+    compute_profile = FilteredDropdown(id='s2id_compute_attribute_compute_profile_id')
+    compute_resource = FilteredDropdown(id='s2id_compute_attribute_compute_resource_id')
+
+    provider_content = ConditionalSwitchableView()
```
abalakh: Why such complex logic with `_is_provider` and partial calls instead of ConditionalSwitchableView(reference='current_provider') and extending current_provider to properly grab only provider part?

airgun/views/computeresource.py
```diff
@@ -199,34 +215,207 @@ class compute_profiles(SatTab):
...
+        ROOT = ".//input[contains(@id, 'bootable')]"
+
+        def _is_checked(self):
+            return self.browser.get_attribute('checked', self)
+
+        def read(self):
+            return self.browser.is_selected(self)
+
+        def fill(self, value):
+            if value is True and not self.browser.is_selected(self):
```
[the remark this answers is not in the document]
ldjebran: This is a radio buttom, we have to select it only if it's not select it , to unselect it we must select an other option.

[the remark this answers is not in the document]
ldjebran: eg , we cannot unselect it. only when selecting an other option

airgun/views/computeresource.py
ldjebran: we have not to change that provider value when reading it, as it may be used by some checking reading tests , remain variant to use a callable.

ldjebran: using a callable is allowed by the API it self.

ldjebran: also do not want to use regex to extract that provider that can lead to errors, the way it's implemented now will work value that end with "provider)".

airgun/views/computeresource.py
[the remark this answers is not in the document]
abalakh: We never create a separate widget for single radio option, only for entire radio (button) group and its `fill()` method accepts option name to select.

airgun/views/computeresource.py
abalakh: No need to change value, yea, i was suggesting using regex. Why do you think it will lead to errors?
As a last resort option you can use `rsplit('(', 1)` to have the same effect. 
Imo current implementation is too tricky and hard to understand from the first glance.

ldjebran: @abalakh will see what I do do here.

ldjebran:
@oshtaier @abalakh updated 
test output for compute profiles tests EC@ and RHV
```
(robottelo) dlezz@dlezz:~/projects/robottelo-fork$ pytest -v tests/foreman/ui_airgun/test_computeresource.py::test_positive_associate_with_custom_profile tests/foreman/ui_airgun/test_computeresource_ec2.py::test_positive_default_end_to_end_with_custom_profile 
platform linux -- Python 3.6.6, pytest-3.6.1, py-1.5.4, pluggy-0.6.0 -- /home/dlezz/.pyenv/versions/robottelo/bin/python3.6
cachedir: .pytest_cache
plugins: wait-for-1.0.9, xdist-1.22.5, services-1.3.1, mock-1.10.0, forked-0.2, cov-2.5.1
collecting 2 items                                                                                           2018-12-18 16:32:14 - conftest - DEBUG - BZ deselect is disabled in settings

collected 2 items                                                                                            

tests/foreman/ui_airgun/test_computeresource.py::test_positive_associate_with_custom_profile PASSED    [ 50%]
tests/foreman/ui_airgun/test_computeresource_ec2.py::test_positive_default_end_to_end_with_custom_profile PASSED [100%]

========================================= 2 passed in 253.27 seconds =========================================
```

airgun/views/computeresource.py
ldjebran: Done via regex

airgun/views/computeresource.py
```diff
@@ -199,34 +215,203 @@ class compute_profiles(SatTab):
...
+        # Returns the items length
+        return len(self.browser.elements(self.ITEMS, parent=self))
+
+    @property
+    def items(self):
+        """Return all the items widget instances"""
+        return [self.get_item_at_index(index) for index in range(self.items_length)]
+
+    def clear(self):
+        """Remove all items if item remove button att defined."""
```
oshtaier: `att`?  should it be `attr`?

PR #257, opened 2018-12-11
Added support for new field custom cron

omkarkhatavkar:
This is related to Satellite 6.5 Feature to support custom_cron. 

- Added this Create View Page 
- Same also added in Edit View Page with Recurring Logic 

Corresponding tests will be added in robottelo soon.

### DIFF
--- a/airgun/views/syncplan.py
+++ b/airgun/views/syncplan.py
@@ -36,6 +36,7 @@ class SyncPlanCreateView(BaseLoggedInView):
     name = TextInput(id='name')
     description = TextInput(id='description')
     interval = Select(id='interval')
+    cron_expression = TextInput(id='cron_expression')
     date_time = DateTime()
     submit = Text("//button[contains(@ng-click, 'handleSave')]")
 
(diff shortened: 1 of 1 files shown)


airgun/views/syncplan.py
```diff
@@ -36,6 +36,7 @@ class SyncPlanCreateView(BaseLoggedInView):
     name = TextInput(id='name')
     description = TextInput(id='description')
     interval = Select(id='interval')
+    cron_expression = TextInput(id='cron_expression')
```
oshtaier: that should be implemented using conditional view, let's me check that

oshtaier: okay, that will probably add unnecessary complexity

airgun/views/syncplan.py
```diff
@@ -73,7 +74,9 @@ class details(SatTab):
         next_sync = ReadOnlyEntry(name='Next Sync')
         enabled = EditableEntryCheckbox(name='Sync Enabled')
         interval = EditableEntrySelect(name='Interval')
+        cron_expression = EditableEntry(name='Cron Logic')
```
oshtaier: just change order to have fields in the same way we have them on UI:
```
name = EditableEntry(name='Name')
description = EditableEntry(name='Description')
date_time = EditableDateTime(name='Start Date')
next_sync = ReadOnlyEntry(name='Next Sync')
recurring_logic = ReadOnlyEntry(name='Recurring Logic')
enabled = EditableEntryCheckbox(name='Sync Enabled')
interval = EditableEntrySelect(name='Interval')
cron_expression = EditableEntry(name='Cron Logic')
products_count = ReadOnlyEntry(name='Products')
```

omkarkhatavkar: Done!

PR #258, opened 2018-12-12
Added support of alert screen which appears after switching organization

abalakh:
So there's a new screen which appears when switching organization while organization-specific entity is opened, looks like that:
image
Adding extra step to click 'Back' button while switching context if such screen appears.

PR #260, opened 2018-12-12
Revert "Added support of alert screen which appears after switching organization"

oshtaier:
Reverts SatelliteQE/airgun#258

Need update from widgetastic side


abalakh:
Yep, needs RedHatQE/widgetastic.core#131 to be merged & released first

PR #264, opened 2018-12-21
Fixes content host and errata entities

ldjebran:


airgun/entities/contenthost.py
```diff
@@ -34,6 +34,11 @@ def read(self, entity_name):
         view = self.navigate_to(self, 'Edit', entity_name=entity_name)
         return view.read()
 
+    def read_details(self, entity_name):
```
oshtaier: that was a long time plan. If that become as necessity, maybe we should discuss a solution

ldjebran: @oshtaier perhaps we should think about def read(self, tabs=None)
if tabs is None we read all, if tabs is specified read only the tabs needed
for example : read(tabs=['details'])

ldjebran: @oshtaier as some entities have many tabs with complicated content as in content host

ldjebran: @oshtaier or def read(self, widgets=None)
and we can read read(['widget.subwidget', ...])

abalakh: I'd vote for `tab=` or `tabs=`. We should convert UI tab name to our class name ('Repository Sets' to repository_sets) as we probably don't want end user to deal with internal names and rely on UI only and to add that extra condition for reading tab(s) or entire view. The tricky part is figuring out a way to do that in one place so we don't have to copy+paste all that logic in every single `read()`.

ldjebran: @abalakh @oshtaier 
 we can implement globally in BaseLoggedInView:
```python
   class BaseLoggedInView(View):
       ...
       def _get_widget_by_name(self, widget_name):
        """Return a widget by it's name, where widget can be a sub widget.

        Example:
             widget_name = 'details'
             or
             widget_name = 'details.subscription_status'
             or
             widget_name = 'details.subscriptions.resources'
        """
        widget = self
        for sub_widget_name in widget_name.split('.'):
            name = sub_widget_name
            if name not in widget.widget_names:
              # if UI name was passed
              # this allow to convert "Repository Sets" to "repository_sets"
                name = name.replace(' ', '_')
                name = name.lower()
                if name not in widget.widget_names:
                    raise AttributeError('Object <{0}> has no widget name "{1}"'.format(
                        widget.__class__, sub_widget_name))
            widget = getattr(widget, name)
        return widget.read()

    def read(self, widget_names=None):
        """Reads the contents of the view and presents them as a dictionary.

        widget_names: If specified , will read only the widgets which names is in the list.

        Returns:
            A :py:class:`dict` of ``widget_name: widget_read_value`` where the values are retrieved
            using the :py:meth:`Widget.read`.
        """
        if widget_names is None:
            return super().read()
        values = {}
        for widget_name in widget_names:
            values[widget_name] = self._get_widget_by_name(widget_name)
        return values
```
and of course use this in entity read function

What do you you think?

ldjebran: resolved separately in PR https://github.com/SatelliteQE/airgun/pull/265

ldjebran:
changed to use the new implemented feature https://github.com/SatelliteQE/airgun/pull/265
test output https://github.com/SatelliteQE/robottelo/pull/6584#issuecomment-453131229

PR #273, opened 2019-01-15
Added few docker-related tabs & helpers; fixed SatTabWithDropdown

abalakh:


airgun/views/common.py
```diff
@@ -138,6 +138,16 @@ class mytab(SatTabWithDropdown):
...
         'contains(@class, "tab-content")]')
 
+    @property
+    def is_displayed(self):
+        return 'ng-hide' not in self.parent_browser.classes(self.TAB_LOCATOR)
+
+    def read(self):
+        """Do not attempt to read hidden tab contents"""
+        if not self.is_displayed:
+            do_not_read_this_widget()
```
abalakh: TabWithDropdown is always present in DOM (just with extra class when its hidden), so before this fix we always received misleading values for hidden tabs with dropdowns

PR #277, opened 2019-01-28
Fixes contentview entity add_search_content_view

ldjebran:
test output https://github.com/SatelliteQE/robottelo/pull/6683

### DIFF
--- a/airgun/entities/contentview.py
+++ b/airgun/entities/contentview.py
@@ -77,6 +77,22 @@ class ContentViewEntity(BaseEntity):
         view.flash.assert_no_error()
         view.flash.dismiss()
 
+    def search_content_view(self, entity_name, value, assigned=True):
+        """Search a composite content view for content views with query value in assigned or
+        unassigned resources.
+        """
+        view = self.navigate_to(self, 'Edit', entity_name=entity_name)
+        assert view.content_views.is_displayed, (
+            'Could not find "Content Views" tab. '
+            'Make sure {} is composite content view'
+            .format(entity_name)
+        )
+        if assigned:
+            resource_widget = view.content_views.resources.list_remove_tab
+        else:
+            resource_widget = view.content_views.resources.add_tab
+        return resource_widget.search(value)
+
     def add_cv(self, entity_name, cv_name):
         """Add content view to selected composite content view."""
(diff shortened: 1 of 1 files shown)


oshtaier:
if that functionality is needed, please use it directly in robottelo as it has nothing to do with framework itself.

ldjebran:
@oshtaier test layer has no access to the resources view to be able to do search in that area and need to read all the content view tabs and widgets to verify only one table content (if not using this search function, we should filter by content if we need a component, that can be also hidden/overridden by the default table limit of 20 elements), sounds also not logical.

mirzal:
As for `assert view.content_views.is_displayed` - we already use that idiom in `add_cv` function, so I would let it pass.

As I understand it, the problem is that we use the same entity for content view page and for composite content view page, but these pages are not exactly the same. Possibly it would make sense to subclass composite content view entity and view out of `ContentViewEntity`? Not sure how much work it is - I guess quite a lot and that's why we didn't do it in the first place. @abalakh added `add_cv`, maybe he knows more on why things are this way.

-----------------------

As for second block - other `search_*` functions go to Content view, select version, open specific tab ("rpm Packages" etc.) and search in table that is displayed there. So it searches content that **was part of specific content view version in the past**, not content that **can be part of next version**. I wouldn't confuse these two scenarios - if both are useful, we at least need different naming convention to distinguish between them.

But we managed to do without searching in content that can be part of next version (where you have two tabs: "List/Remove" and "Add") until now. Maybe that is something that you can reasonably work around in test itself?

abalakh:
@mirzal Regarding assertion - i've done it in `add_cv` to make debugging a bit easier, simple as that. Since content view is kinda tricky entity (or more like 2 somewhat similar entities combined into one), i thought that it might be pretty common mistake at test layer to use wrong cv type, and in such case `make sure it's composite cv` sounds way better to me than some `NavigationTriesExceeded` error with long traceback.

Regarding `search_content_view` method itself - we had a meeting and decided that we don't need some special search in such case - as our intend is to fetch all available content views and pagination may hide some of them from us, then it's issue with `read`, not `search`; we should update entity's `read` to correctly handle pagination and grab all the details, not just the first page.

ldjebran:
As discussion offline, Closing this PR to implement a more general table data retrieval using pagination.

PR #280, opened 2019-01-29
quick cv fix

oshtaier:


airgun/views/contentview.py
```diff
@@ -106,7 +106,7 @@ def is_displayed(self):
 
 class ContentViewEditView(BaseLoggedInView):
     breadcrumb = BreadCrumb()
-    publish = Text("//button[contains(., 'Publish New Version')]")
+    publish = Button("Publish New Version")
```
abalakh: Any details? Change looks pretty much similar to what it was. What does it fix?

oshtaier: now it is real button with all related events, validators and etc. Can't find exact code piece, but I don't need to do it. In result, if button is not on a page, it is correctly return exception, not clicking empty place on a page

PR #284, opened 2019-02-04
Added more views and actions in entities for selectable customizable …

latran:
Added view and entity files for selectable customizable column.

### DIFF
--- a/airgun/entities/subscription.py
+++ b/airgun/entities/subscription.py
@@ -107,6 +107,23 @@ class SubscriptionEntity(BaseEntity):
         view = self.navigate_to(self, 'All')
         return view.search(value)
 
+    def filter_columns(self, columns=None):
+        """Filters column headers
+        :param columns: dict mapping column name to boolean value
+        :return: tuple of the name of the headers
+        """
+        view = self.navigate_to(self, 'All')
+        if columns is not None:
+            view.columns_filter_checkboxes.fill(columns)
+            wait_for(
+                lambda: view.table.headers is not None,
+                timeout=10,
+                delay=1,
+                handle_exception=True,
+                logger=view.logger
+            )
+        return view.table.headers
+
     def provided_products(self, entity_name, virt_who=False):
         """Read list of all products provided by subscription.
         :param entity_name: Name of subscription
--- a/airgun/views/subscription.py
+++ b/airgun/views/subscription.py
@@ -80,6 +80,54 @@ class ProductContentItemsList(GenericLocatorWidget):
         raise ReadOnlyWidgetError('Widget is read only, fill is prohibited')
 
 
+class SubscriptionColumnsFilter(GenericLocatorWidget):
+    """This is the list of interaction items for when opening up the selectable customizable
+    checkboxes"""
+
+    ITEMS_LOCATOR = "//div[@id='subscriptionTableTooltip']//li/span"
+    CHECKBOX_LOCATOR = "//div[@id='subscriptionTableTooltip']//li/span[.='{}']/preceding::input[1]"
+
+    @property
+    def is_open(self):
+        return 'tooltip-open' in self.browser.classes(self)
+
+    def open(self):
+        if not self.is_open:
+            self.click()
+
+    def close(self):
+        if self.is_open:
+            self.click()
+
+    def checkboxes(self):
+        labels = [
+            line.text
+            for line in self.browser.elements(self.ITEMS_LOCATOR)
+        ]
+        return {
+            label: Checkbox(self, locator=self.CHECKBOX_LOCATOR.format(label))
+            for label in labels
+        }
+
+    def read(self):
+        """Read values of checkboxes"""
+        self.open()
+        values = {
+            name: checkbox.read()
+            for name, checkbox in self.checkboxes.items()
+        }
+        self.close()
+        return values
+
+    def fill(self, values):
+        """Check or uncheck one of the checkboxes """
+        self.open()
+        checkboxes = self.checkboxes()
+        for name, value in values.items():
+            checkboxes[name].fill(value)
+        self.close()
+
+
 class SubscriptionListView(BaseLoggedInView, SubscriptionSearchableViewMixin):
     """List of all subscriptions."""
     table = SatSubscriptionsViewTable(
@@ -89,12 +137,15 @@ class SubscriptionListView(BaseLoggedInView, SubscriptionSearchableViewMixin):
             'Name': Text("./a"),
         }
     )
+
     add_button = Button(href='subscriptions/add')
     manage_manifest_button = Button('Manage Manifest')
     export_csv_button = Button('Export CSV')
     delete_button = Button('Delete')
     progressbar = ProgressBar('//div[contains(@class,"progress-bar-striped")]')
     confirm_deletion = DeleteSubscriptionConfirmationDialog()
+    columns_filter_checkboxes = SubscriptionColumnsFilter(
+        ".//form[div[contains(@class, 'filter')]]/div/i")
 
     @property
     def is_displayed(self):


airgun/entities/subscription.py
```diff
@@ -145,6 +145,48 @@ def delete(self, entity_name):
         self._wait_for_process_to_finish(
                 'Delete Upstream Subscription', has_manifest=True)
 
+    def read_header(self):
```
oshtaier: doesn't Table object contain information about its columns headers?

airgun/entities/subscription.py
```diff
@@ -145,6 +145,48 @@ def delete(self, entity_name):
...
+        """Read the head columns under Subscription"""
+        view = self.navigate_to(self, 'All')
+        return view.table.headers
+
+    def reset_scc(self):
```
oshtaier: that should be placed on test layer whether in test module helper or if you have only one test case for that - then in test itself

airgun/entities/subscription.py
```diff
@@ -145,6 +145,48 @@ def delete(self, entity_name):
...
+            view.req_virt_who,
+            view.type,
+            view.consumed,
+            view.entitlements]
+        for item in scc_checkbox_list:
+            if item.selected:
+                item.click()
+        view.scc_button.click()
+
+    def default_scc(self):
```
oshtaier: same here

airgun/views/subscription.py
```diff
@@ -89,13 +89,25 @@ class SubscriptionListView(BaseLoggedInView, SubscriptionSearchableViewMixin):
...
     progressbar = ProgressBar('//div[contains(@class,"progress-bar-striped")]')
     confirm_deletion = DeleteSubscriptionConfirmationDialog()
 
+    name = Checkbox(id='id')
+    sku = Checkbox(id='product_id')
+    contract = Checkbox(id='contract_number')
+    start_date = Checkbox(id='start_date')
+    end_date = Checkbox(id='end_date')
+    req_virt_who = Checkbox(id='virt_who')
+    type = Checkbox(id='type')
```
oshtaier: `type` is a reserved name, of course it is not important, but please rename that variable

airgun/entities/subscription.py
latran: I'm not too sure what you mean here.

airgun/entities/subscription.py
latran: I can only find table.headers within view = self.navigate_to(self, 'All').  If there's a better way to display the header info than this, please advise on how that is implemented.

airgun/entities/subscription.py
oshtaier: I mean that framework should not contain such helpers. If you need select any checkbox, so do it in the test directly. If you need to select all of them (I am not sure why you need it), so just do that still in the test

airgun/entities/subscription.py
oshtaier: Right
so why do you need separate method to read them?

latran: Here's my initial reasoning for needing a separate method:
1. I'm separating reading the headers and checking/unchecking the checkboxes for assertion purposes in ui_airgun/test_subscription.py
2. Expanding the test cases in the future may just need checking/unchecking the checkboxes without need of reading the headers.

After thinking about it, I see what you mean by not separating read header into a method because reason 1 and 2 wouldn't be a large blocker to include or not include the read header in there.

I'll play around with this a bit more and fix the changes based on your suggestion.

airgun/entities/subscription.py
latran: When you say "if you need to select any checkboxes, so do it in the test directly" do you mean write that in ui_airgun/test_subscription.py or where?

I only did the select all or not select all because I'm ensuring that all or no headers show up correctly based on those 2 filtering.

latran:
I redid the implementation to make it more dynamic based on dlezz's help and suggestion.

airgun/entities/subscription.py
```diff
@@ -107,6 +107,18 @@ def search(self, value):
         view = self.navigate_to(self, 'All')
         return view.search(value)
 
+    def filter_columns(self, columns=None):
+        """Filters column headers
+        :param columns: name of column header to filter by
```
mirzal: That suggests we can pass in list or tuple, which isn't really true. We need to pass in dict that contains column labels as keys and boolean values as values. Maybe `dict with column label: boolean value pairs`?

airgun/entities/subscription.py
```diff
@@ -107,6 +107,18 @@ def search(self, value):
...
+        :return: tuple of the name of the headers
+        """
+        view = self.navigate_to(self, 'All')
+        if columns is not None:
+            view.columns_filter_checkboxes.fill(columns)
+            self.browser.plugin.ensure_page_safe()
+            view = self.navigate_to(self, 'All')
```
mirzal: Is this `navigate_to` necessary? You are on `All` and as far as I can tell, interacting with columns filter can't navigate you out of that page. So it kind of looks like code that doesn't do anything useful.

I suppose you added this because this is React page and without it, your script was trying to continue too early (i.e. before table columns were changed). Would it make sense to identify which exact part of `navigate_to` stabilizes your script and add it here directly?

airgun/views/subscription.py
```diff
@@ -89,12 +134,14 @@ class SubscriptionListView(BaseLoggedInView, SubscriptionSearchableViewMixin):
...
         }
     )
+
     add_button = Button(href='subscriptions/add')
     manage_manifest_button = Button('Manage Manifest')
     export_csv_button = Button('Export CSV')
     delete_button = Button('Delete')
     progressbar = ProgressBar('//div[contains(@class,"progress-bar-striped")]')
     confirm_deletion = DeleteSubscriptionConfirmationDialog()
+    columns_filter_checkboxes = SubscriptionColumnsFilter(".//form[div[contains(@class, 'filter')]]/div/i")
```
mirzal: Personally it would make sense for me to add it above, simply because this is the first element on the page you will see going left-right, top-bottom.

But usually we don't really care about order of widgets, so you can leave it as is.

airgun/entities/subscription.py
latran: @mirzal You are correct in that assumption!  I'm still trying to figure out which part of navigate_to is used, but unable to identify the exact part of it.  This still fails from time to time as is so I do need help in this.

mirzal: I think it would be one of two places:
* [`am_i_here`](https://github.com/SatelliteQE/airgun/blob/master/airgun/navigation.py#L57) which would call [`view is_displayed`](https://github.com/SatelliteQE/airgun/blob/master/airgun/views/subscription.py#L100)
* subscription's own [`post_navigate`](https://github.com/SatelliteQE/airgun/blob/master/airgun/entities/subscription.py#L152), which eventually calls view is_displayed as well.

But `is_displayed` waits for div with id `subscriptions-table`, which - once it appears on page - does not disappear. When changing columns, React will first remove all cells in table heading one by one, then add expected set of cells, and then remove entire body of table, only to add it back (this time with new columns included). I'd say calling `self.navigate_to` doesn't actually wait for React to finish, it just delays further execution for long enough that React completes in the meantime.

As far as I know, there is no nice and clean solution. React doesn't expose internal state, so we have no way of knowing if it does something (and we should wait) or not.

One idea that comes to my mind is waiting for first row in table to have exact number of cells (which you should be able to compute based on number of columns before any action and results of changes from `checkbox.fill()`). This won't catch the case when you add and remove the same number of columns in one transaction. But I think you only care about displaying `virt-who` column, so it should suffice for now.

airgun/entities/subscription.py
latran: @mirzal 
I'll just update the comment of the param to reflect that it's a dict mapping column name to boolean value if that's alright.

airgun/entities/subscription.py
latran: I've tried those implementations before and they don't work as well as `navigate_to.`  I'm leaving it as is for now since it's working more consistently every time I've run it.

mirzal: Thanks for looking into this and trying other options as well. It's a bit unfortunate that they didn't work as well as navigation, but I take automation stability over everything else, so I'm OK with leaving it as is.

latran:
@oshtaier Can I get an approval so we can begin the merging process?

airgun/entities/subscription.py
```diff
@@ -107,6 +107,18 @@ def search(self, value):
...
+    def filter_columns(self, columns=None):
+        """Filters column headers
+        :param columns: dict mapping column name to boolean value
+        :return: tuple of the name of the headers
+        """
+        view = self.navigate_to(self, 'All')
+        if columns is not None:
+            view.columns_filter_checkboxes.fill(columns)
+            self.browser.plugin.ensure_page_safe()
+            view = self.navigate_to(self, 'All')
```
oshtaier: sorry, but we cannot submit code without exact reason, even if that block something. What issue do you have here? `StaleElementException`? We can look on alternative solutions. E.g. waiter that will check for necessary element is visible.

oshtaier:
Can you share server details with functionality? (in email)

airgun/entities/subscription.py
latran: @oshtaier yes, I encountered the `StaleElementReferenceException` here.  I've tried several solutions that was suggested while working with @ldjebran on this piece and this was the only one that worked well enough during the time.

oshtaier:
Hi @latran 

I finally looked into the issue.

1) Functionality is completely fine for automation. I thought about another filtering that is present on every page for every table. Here you have special one
2) Framework part looks good on 95%. I will provide comments later once we see that everything works properly
3) I think that we need improve test part, make it more complex, but not sure how at that moment.

According the issue. I am still in the debug. That is a problem of fundamental part of the framework. Not airgun, but widgetastic core itself.

`sleep(5)` - works
but
```
try:
except StaleElementReferenceException:
    time.sleep(0.5)
```
doesn't

```
    def filter_columns(self, columns=None):
        """Filters column headers
        :param columns: dict mapping column name to boolean value
        :return: tuple of the name of the headers
        """
        if columns is not None:
            view.columns_filter_checkboxes.fill(columns)
            sleep(5)
        return view.table.headers
```
works fine, but it is not a solution

Btw:
```
        if columns is not None:
            view.columns_filter_checkboxes.fill(columns)
```
doesn't work for me 100 times from 100 times. Probably, you have old code for navigation

latran:
@oshtaier Can the complex test be filtering and drilling into a link versus filtering and not drilling into a link?

oshtaier:
Hi @latran Can we move further with that? Can I help you in any way?

airgun/entities/subscription.py
```diff
@@ -107,6 +107,18 @@ def search(self, value):
         view = self.navigate_to(self, 'All')
         return view.search(value)
 
+    def filter_columns(self, columns=None):
```
oshtaier:
```
        if columns is not None:
            view.columns_filter_checkboxes.fill(columns)
            wait_for(
                lambda: view.table.headers is not None,
                timeout=10,
                delay=1,
                handle_exception=True,
                logger=view.logger
            )
        return view.table.headers
```

airgun/views/subscription.py
```diff
@@ -17,7 +17,7 @@
     ItemsListReadOnly,
     ProgressBar,
     SatTable,
-    Search,
+    Search
```
oshtaier: initial line was correct

airgun/views/subscription.py
```diff
@@ -80,6 +80,51 @@ def fill(self, value):
         raise ReadOnlyWidgetError('Widget is read only, fill is prohibited')
 
 
+class SubscriptionColumnsFilter(GenericLocatorWidget):
```
oshtaier: please provide docstring for that class as we do for every class

airgun/views/subscription.py
```diff
@@ -89,12 +134,15 @@ class SubscriptionListView(BaseLoggedInView, SubscriptionSearchableViewMixin):
...
         }
     )
+
     add_button = Button(href='subscriptions/add')
     manage_manifest_button = Button('Manage Manifest')
     export_csv_button = Button('Export CSV')
     delete_button = Button('Delete')
     progressbar = ProgressBar('//div[contains(@class,"progress-bar-striped")]')
     confirm_deletion = DeleteSubscriptionConfirmationDialog()
+    columns_filter_checkboxes = \
```
oshtaier:
```
columns_filter_checkboxes = SubscriptionColumnsFilter(
    ".//form[div[contains(@class, 'filter')]]/div/i")
```

latran:
Can I get some approvals for this?

oshtaier:
make sure that code is tested

latran:
@oshtaier yup it has been tested!

oshtaier:
then we need one more reviewer

latran:
@mirzal can you review?  Thanks.

PR #286, opened 2019-02-07
Fixes host entity schedule_remote_job

ldjebran:
test output https://github.com/SatelliteQE/robottelo/pull/6659#issuecomment-461393340

### DIFF
--- a/airgun/entities/host.py
+++ b/airgun/entities/host.py
@@ -104,12 +104,14 @@ class HostEntity(BaseEntity):
         view.export.click()
         return self.browser.save_downloaded_file()
 
-    def schedule_remote_job(self, entities_list, values, timeout=60):
+    def schedule_remote_job(self, entities_list, values, timeout=60, wait_for_results=True):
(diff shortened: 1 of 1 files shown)


airgun/entities/host.py
```diff
@@ -104,12 +104,14 @@ def export(self):
...
         return self.browser.save_downloaded_file()
 
-    def schedule_remote_job(self, entities_list, values, timeout=60):
+    def schedule_remote_job(self, entities_list, values, timeout=60, wait_for_results=True):
         """Apply Schedule Remote Job action to the hosts names in entities_list
 
         :param entities_list: The host names to apply the remote job.
         :param values: the values to fill The Job invocation view.
         :param timeout: The time to wait for the job to finish.
+        :param wait_for_results: Whether to wait for the job to finish execution and return the
```
oshtaier: just `Whether to wait for the job to finish execution` without `and return the final job status results.` as we return status no matter what

PR #287, opened 2019-02-13
Fixes Discovery Rule entity

ldjebran:
test output https://github.com/SatelliteQE/robottelo/pull/6736#issue-252762785

### DIFF
--- a/airgun/views/discoveryrule.py
+++ b/airgun/views/discoveryrule.py
@@ -11,8 +11,8 @@ from airgun.views.common import (
     SatTab,
 )
 from airgun.widgets import (
-    AutoCompleteSearchInput,
     ActionsDropdown,
+    AutoCompleteTextInput,
     FilteredDropdown,
     MultiSelect,
     SatTable,
@@ -55,7 +55,7 @@ class DiscoveryRuleCreateView(BaseLoggedInView):
     @View.nested
     class primary(SatTab):
         name = TextInput(id='discovery_rule_name')
-        search = AutoCompleteSearchInput(id='search')
+        search = AutoCompleteTextInput(id='search')
         host_group = FilteredDropdown(id='discovery_rule_hostgroup_id')
         hostname = TextInput(id='discovery_rule_hostname')
         hosts_limit = TextInput(id='discovery_rule_max_count')
--- a/airgun/views/discoveryrule.py
+++ b/airgun/views/discoveryrule.py
@@ -11,6 +11,7 @@ from airgun.views.common import (
     SatTab,
 )
 from airgun.widgets import (
+    AutoCompleteSearchInput,
     ActionsDropdown,
     FilteredDropdown,
     MultiSelect,
@@ -36,19 +37,6 @@ class DiscoveryRulesView(BaseLoggedInView):
             self.title, exception=False) is not None
 
 
-class SearchInput(TextInput):
-    """Discovery Search input field, This field has no value attribute, and The only way to set or
-    read it values correctly is using javascript.
-    """
-
-    @property
-    def value(self):
-        return self.browser.execute_script('return arguments[0].value;', self)
-
-    def fill(self, value):
-        self.browser.execute_script('arguments[0].value=arguments[1];', self, value)
-
-
 class DiscoveryRuleCreateView(BaseLoggedInView):
     submit = Text('//input[@name="commit"]')
     cancel = Text('//a[text()="Cancel"]')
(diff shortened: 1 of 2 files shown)


mirzal:
I understand that after this change, test case passes. I assume it is stable, i.e. passes every time you run it.

But my concern is, code like that shouldn't be necessary. Airgun's `TextInput.value()` will call Selenium `get_attribute`, which - after going through webdriver - will eventually call JavaScript's `element[attribute]`. And for this particular field, it does work and returns expected output. At least in Chrome webdev tools:

```
> document.getElementById('search')['value']
" activation_key_id  = 4"
```

Note that JavaScript's Node object property `value` is distinct from HTML `<input>` attribute `value`. JavaScript `value` is current value of field. HTML `value` is **default** value of `<input>`. If user changes field value, JavaScript will display changed string, HTML will display old string.   
Other fields on that page are missing HTML attribute `value` as well, but they don't require this workaround.

My wild guess would be that problem is not that we need to set/read value using JavaScript directly, but that this field has autocompletion elements that somehow mess up with Airgun. Could it be that when Airgun is trying to fill next field ("host group"), it is below "search" autocompletion? Could you share some more details on when and how your test failed before this change?

--------------

On separate note, this field is AutoCompletion component from React. As of now, there are only two places in entire Satellite where this is used - one that you found here, another one on "Create Filter" page, if "Unlimited?" checkbox is not checked. But as Satellite will gradually move to React, we will probably see it in other places as well. Then, we might want to move it to `widgets.py`.

Just something for other reviewers to consider.

airgun/views/discoveryrule.py
```diff
@@ -36,6 +36,19 @@ def is_displayed(self):
             self.title, exception=False) is not None
 
 
+class SearchInput(TextInput):
+    """Discovery Search input field, This field has no value attribute, and The only way to set or
+    read it values correctly is using javascript.
```
oshtaier: we can't even read value?

ldjebran: Yes the standard read, return the value from the list and not the current visible one.

ldjebran:
Please do not merge for now some changes appeared in last snap and value attribute is rule search input field is available now and have to make some testing and perhaps some changes.

abalakh:
I'm currently working with container tag input which has suggestions popping up with exactly the same problem @mirzal mentioned - suggestions are covering next field, causing issues. And i was planning to add extra widget for that, but it looks like there's a chance we facing the same issue. @ldjebran let's collaborate and try to create some generic widget (if our cases are actually similar).

ldjebran:
@mirzal @oshtaier thanks for review updated the PR thanks to @abalakh for coworking on this, the problem was after fill the autocompletion list does not disappear immediately or does not disappear at all in other entities (@abalakh case).

airgun/widgets.py
```diff
@@ -1718,3 +1718,15 @@ def fill(self, values):
                 widget = get_widget_by_name(self, key)
                 if widget:
                     widget.fill(value)
+
+
+class AutoCompleteSearchInput(TextInput):
```
abalakh: Any other ideas for widget name? It doesn't necessarily mean it's search input, basically any text input with auto complete, smth like 'text input with autocomplete'. `AutoCompleteTextInput` maybe? Or smth similar

airgun/views/discoveryrule.py
```diff
@@ -11,6 +11,7 @@
     SatTab,
 )
 from airgun.widgets import (
+    AutoCompleteSearchInput,
```
abalakh: nitpick: pls sort alphabetically :)

abalakh:
And btw, @mirzal thanks a lot for catching and bringing attention to the issue 🍻

oshtaier:
and test result for affected discovery test case

ldjebran:
@oshtaier changes done
```
(robottelo) dlezz@dlezz:~/projects/robottelo-fork$ pytest -v tests/foreman/ui_airgun/test_discoveryrule.py::test_positive_end_to_end
platform linux -- Python 3.6.6, pytest-4.0.2, py-1.5.4, pluggy-0.8.0 -- /home/dlezz/.pyenv/versions/robottelo/bin/python3.6
cachedir: .pytest_cache
plugins: xdist-1.26.0, timeout-1.3.3, services-1.3.1, mock-1.10.0, forked-0.2, cov-2.6.1
collecting ... 2019-02-25 09:04:21 - conftest - DEBUG - BZ deselect is disabled in settings

collected 1 item                                                                                             

tests/foreman/ui_airgun/test_discoveryrule.py::test_positive_end_to_end PASSED                         [100%]

=================================== 1 passed, 2 warnings in 169.66 seconds ===================================

```

PR #290, opened 2019-02-19
Fixes Role and Filter entity

ldjebran:
test output https://github.com/SatelliteQE/robottelo/pull/6751
dependency: https://github.com/SatelliteQE/airgun/pull/292

### DIFF
--- a/airgun/entities/filter.py
+++ b/airgun/entities/filter.py
@@ -29,6 +29,24 @@ class FilterEntity(BaseEntity):
             self, 'Edit', role_name=role_name, entity_name=entity_name)
         return view.read(widget_names=widget_names)
 
+    def read_all(self, role_name):
+        """Read all the available role filters table values"""
+        view = self.navigate_to(self, 'All', role_name=role_name)
+        return view.table.read()
+
+    def read_permissions(self, role_name):
+        """Return all role filters permissions as a dict with resources as keys and permission
+        names as values.
+        """
+        rows = self.read_all(role_name)
+        permissions = {}
+        for row in rows:
+            resource = row['Resource']
+            if resource not in permissions:
+                permissions[resource] = []
+            permissions[resource].extend(row['Permissions'].split(', '))
+        return permissions
+
     def update(self, role_name, entity_name, values):
         """Update filter values"""
         view = self.navigate_to(
--- a/airgun/widgets.py
+++ b/airgun/widgets.py
@@ -25,7 +25,7 @@ from widgetastic_patternfly import (
     VerticalNavigation,
 )
 
-from airgun.exceptions import ReadOnlyWidgetError
+from airgun.exceptions import DisabledWidgetError, ReadOnlyWidgetError
 from airgun.utils import get_widget_by_name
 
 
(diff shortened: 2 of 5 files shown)


airgun/widgets.py
```diff
@@ -766,6 +766,79 @@ def fill(self, value):
         self.filter_content.fill(value)
 
 
+class ButtonDropdown(GenericLocatorWidget):
```
oshtaier: so ActionsDropdown doesn't cover your case?

ldjebran:
Remove pagination table changes to be implemented in it's own PR

airgun/entities/filter.py
```diff
@@ -29,6 +29,24 @@ def read(self, role_name, entity_name, widget_names=None):
...
+        """Return all role filters permissions as a dict with resources as keys and permission
+        names as values.
+        """
+        rows = self.read_all(role_name)
+        permissions = {}
+        for row in rows:
+            resource = row['Resource']
+            if resource not in permissions:
+                permissions[resource] = []
+            permissions[resource].extend(row['Permissions'].split(', '))
```
mirzal: You can drop `if` and use this instead: 
```
permissions.setdefault(resource, []).extend(row['Permissions'].split(', '))
```

airgun/entities/filter.py
```diff
@@ -29,6 +29,24 @@ def read(self, role_name, entity_name, widget_names=None):
             self, 'Edit', role_name=role_name, entity_name=entity_name)
         return view.read(widget_names=widget_names)
 
+    def read_all(self, role_name):
```
mirzal: Not sure about this name... in test code, we would have `session.filter.read_all(role_name)`. If I saw that, I think I would have trouble understanding what "all" refers to.

Maybe use `read_filters` instead? `session.filter.read_filters(role_name)` is a bit redundant, but at least it's clear we read all filters for specified role. It's also consistent with `read_permissions`.

airgun/entities/filter.py
ldjebran: Nope we need to return a pure dict, (that was my first implementation :) )

ldjebran: Nope my first implementation was with permissions = defaultdict([]), any way do not think this can be a blocker as it's really complicate the reading of this line.

ldjebran: if everybody vote for this will implement it

airgun/entities/filter.py
ldjebran: we are at filter entity so we reading filters values , many entities already use this function in the same context , why we should rename it here?.

mirzal: Other entities with that function don't require argument here, so there is no ambiguity if "all" refers to entity (filter) or argument (role).

Feel free to disagree and resolve conversation. I don't require this to be changed, I'm only raising concern.

ldjebran: if you look at all the functions of this entity:
```python
class FilterEntity(BaseEntity):
    def create(self, role_name, values):

    def search(self, role_name, value):
    
    def read(self, role_name, entity_name, widget_names=None):

    def update(self, role_name, entity_name, values):

    def delete(self, role_name, entity_name):
``` 
why this function has to be an exception?

mirzal: OK, let's keep it your way.

airgun/entities/filter.py
abalakh: I don't usually encourage redundant conditions when there's shorter version, but i have to agree in this case option with `if` is better readable, 3 func calls in a single line is too much and it's hard to understand what's going on from the first glance.

oshtaier: it is not so important

PR #293, opened 2019-02-26
Add media entity

oshtaier:


airgun/entities/media.py
```diff
@@ -0,0 +1,83 @@
...
+    def delete(self, entity_name):
+        """Delete media"""
+        view = self.navigate_to(self, 'All')
+        view.search(entity_name)
+        view.table.row(name=entity_name)['Actions'].widget.click(
+            handle_alert=True)
+
+
+@navigator.register(MediaEntity, 'All')
+class ShowAllMedias(NavigateStep):
```
abalakh: Maybe we should use 'Medium' instead of 'Medias' as a plural form?

airgun/views/media.py
```diff
@@ -0,0 +1,64 @@
...
+        breadcrumb_loaded = self.browser.wait_for_element(
+            self.breadcrumb, exception=False)
+        return (
+            breadcrumb_loaded
+            and self.breadcrumb.locations[0] == 'Installation Media'
+            and self.breadcrumb.read() == 'Create Medium'
+        )
+
+    @View.nested
+    class medium(SatTab):
```
abalakh: especially as 'medium' is already used due to UI naming :)

airgun/views/media.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+class MediasView(BaseLoggedInView, SearchableViewMixin):
+    title = Text("//h1[text()='Installation Media']")
+    new = Text("//a[contains(@href, '/media/new')]")
+    table = SatTable(
+        './/table',
+        column_widgets={
+            'Name': Text('./a'),
+            'Actions': Text('.//a[@data-method="delete"]'),
```
abalakh: Why not Button? Are we gonna enforce using buttons btw or it still remains optional?

oshtaier: That is probably under discussion. I would suggest to use Button for all unique buttons that can be addressed without xpath and in all places where you actually need to track events and etc.
Yeah, we can discuss that later

airgun/entities/media.py
oshtaier: whatever

PR #296, opened 2019-02-27
Fixes, new views and helpers for Container entity

abalakh:
Test results in https://github.com/SatelliteQE/robottelo/pull/6773#issue-256759373

### DIFF
--- a/airgun/entities/container.py
+++ b/airgun/entities/container.py
@@ -39,6 +39,7 @@ class ContainerEntity(BaseEntity):
         else:
             raise ValueError(
                 'Wrong `power` value passed. Supported values are True/False/"On"/"Off"')
+        # workaround for BZ1683348
         view.flash.assert_no_error(
             ignore_messages=['Error - wrong number of arguments (given 1, expected 0)'])
         view.flash.dismiss()
--- a/airgun/entities/container.py
+++ b/airgun/entities/container.py
@@ -1,7 +1,7 @@
 from navmazing import NavigateToSibling
 from airgun.entities.base import BaseEntity
 from airgun.navigation import NavigateStep, navigator
-from airgun.views.container import ContainerCreateView, ContainerView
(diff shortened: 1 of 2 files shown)


airgun/entities/container.py
```diff
@@ -19,11 +19,44 @@ def search(self, value):
...
+        view = self.navigate_to(self, 'Details', entity_name=entity_name)
+        if power in (True, 'On'):
+            view.power_on.click()
+        elif power in (False, 'Off'):
+            view.power_off.click(handle_alert=True)
+        else:
+            raise ValueError(
+                'Wrong `power` value passed. Supported values are True/False/"On"/"Off"')
+        view.flash.assert_no_error(
+            ignore_messages=['Error - wrong number of arguments (given 1, expected 0)'])
```
oshtaier: what is that one about?

abalakh: [BZ1683348](https://bugzilla.redhat.com/show_bug.cgi?id=1683348) just some weird message appears, but container stops/starts successfully

oshtaier: I think comment with BZ id will not harm much

airgun/entities/container.py
```diff
@@ -37,4 +70,26 @@ class AddNewContainer(NavigateStep):
...
+    VIEW = ContainerDetailsView
+
+    def prerequisite(self, *args, **kwargs):
+        return self.navigate_to(self.obj, 'All')
+
+    def am_i_here(self, *args, **kwargs):
+        entity_name = kwargs.get('entity_name')
+        return (
+            self.view.is_displayed
+            and self.view.breadcrumb.read() == '{} - /{}'.format(
```
oshtaier: strange construct =)

abalakh: You don't say :)
That's how it looks on WebUI:
image

PR #297, opened 2019-03-04
Add hardware model entity and update dropdown widget

oshtaier:


airgun/widgets.py
```diff
@@ -757,11 +758,18 @@ def read(self):
...
+        self.clear_filter.click()
+
     def fill(self, value):
         """Select specific item from the drop-down
 
         :param value: string with item value
         """
+        if value == '':
+            self.clear()
+            return
```
abalakh:
```suggestion
            return True
```

PR #304, opened 2019-03-15
Fixes rhai entities

ldjebran:
test output https://github.com/SatelliteQE/robottelo/pull/6813

### DIFF
--- a/airgun/entities/rhai/base.py
+++ b/airgun/entities/rhai/base.py
@@ -1,12 +1,25 @@
-from airgun.exceptions import InsightsOrganizationPageError
+from airgun.exceptions import DestinationNotReachedError
 from airgun.navigation import NavigateStep
 from airgun.views.rhai import InsightsOrganizationErrorView
+
+
+class InsightsOrganizationPageError(Exception):
+    """Raised when navigating to insight plugin pages and the organization is not selected
+     or the current selected organization has no manifest.
+    """
 
 
 class InsightsNavigateStep(NavigateStep):
 
     def post_navigate(self, _tries, *args, **kwargs):
-        """raise Error if Organization Error page is displayed"""
-        org_err_page_view = InsightsOrganizationErrorView(self.view.browser)
-        if org_err_page_view.is_displayed:
-            raise InsightsOrganizationPageError(org_err_page_view.read())
+        """Raise Error if Destination view is not displayed or Organization Error page
+        is displayed.
+        """
+        if not self.view.is_displayed:
+            org_err_page_view = InsightsOrganizationErrorView(self.view.browser)
+            if org_err_page_view.is_displayed:
+                raise InsightsOrganizationPageError(org_err_page_view.read())
+            raise DestinationNotReachedError(
+                'Navigation destination view "{0}" not reached'.format(
+                    self.view.__class__.__name__)
+            )
--- a/airgun/entities/rhai/base.py
+++ b/airgun/entities/rhai/base.py
@@ -1,0 +1,12 @@
+from airgun.exceptions import InsightsOrganizationPageError
+from airgun.navigation import NavigateStep
+from airgun.views.rhai import InsightsOrganizationErrorView
+
+
+class InsightsNavigateStep(NavigateStep):
+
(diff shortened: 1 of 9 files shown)


airgun/views/rhai.py
```diff
@@ -35,9 +58,22 @@ def is_displayed(self):
         return self.title.is_displayed and self.search.is_displayed
 
 
+class InventoryHostRule(GenericRemovableWidgetItem):
+    """Insights inventory host rule widget"""
+    # There is no remove button in this widget
```
oshtaier: sounds weird. `GenericRemovableWidgetItem` and `There is no remove button in this widget`. `Removable`, but `no remove button`. That is not logical. Any alternative

airgun/entities/rhai/base.py
```diff
@@ -0,0 +1,12 @@
+from airgun.exceptions import InsightsOrganizationPageError
+from airgun.navigation import NavigateStep
+from airgun.views.rhai import InsightsOrganizationErrorView
+
+
+class InsightsNavigateStep(NavigateStep):
```
oshtaier: propose to not use that navigation at all

ldjebran:
output to latest changes https://github.com/SatelliteQE/robottelo/pull/6813#issuecomment-473348979

airgun/views/rhai.py
```diff
@@ -10,7 +10,30 @@
...
+    title = Text("//article[@id='content']/section/h1")
+    message = Text("//article[@id='content']/section/p")
+
+    @property
+    def is_displayed(self):
+        return (self.browser.wait_for_element(self.title, exception=False) is not None
+                and self.browser.wait_for_element(self.message, exception=False) is not None)
+
+    def read(self, widget_names=None):
+        if widget_names:
```
oshtaier: as far as I understand, we know that `if not self.view.is_displayed:`  and we know that no widgets are displayed, so I suggest to use plain:
`return '{0}: {1}'.format(self.title.read(), self.message.read())`

airgun/views/rhai.py
oshtaier: Maybe we can create an issue per our discussion? To create better architecture for such kind of widgets?

ldjebran:
@oshtaier changes done and issue https://github.com/SatelliteQE/airgun/issues/305 created

airgun/views/rhai.py
```diff
@@ -10,7 +10,28 @@
...
+    title = Text("//article[@id='content']/section/h1")
+    message = Text("//article[@id='content']/section/p")
+
+    @property
+    def is_displayed(self):
+        return (self.browser.wait_for_element(self.title, exception=False) is not None
+                and self.browser.wait_for_element(self.message, exception=False) is not None)
+
+    def read(self, widget_names=None):
+        return '{0}: {1}'.format(self.title.read(), self.message.read())
```
abalakh: I'm not sure we should mess with custom read, especially as there's no real need to - both title and message will be returned correctly with default read implementation.

airgun/views/rhai.py
```diff
@@ -10,7 +10,28 @@
...
+class InsightsOrganizationErrorView(BaseLoggedInView):
+    """View displayed when no Organization is selected or when the current organization has no
+    manifest
+    """
+    title = Text("//article[@id='content']/section/h1")
+    message = Text("//article[@id='content']/section/p")
+
+    @property
+    def is_displayed(self):
+        return (self.browser.wait_for_element(self.title, exception=False) is not None
```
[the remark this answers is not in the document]
ldjebran: 1- This has a very specific usage in insight 
2- If you look at the locators 
title = Text("//article[@id='content']/section/h1")
message = Text("//article[@id='content']/section/p")
you will never fiver find a normal page with these 2 element at the same time for forman / kattelo there is not page like this even with only title, in the insight also there is no page with locator except this error page

airgun/views/rhai.py
ldjebran: This really has no importance, is used in one place to get that error message, if we do not pack it here , we will pack else where, but really do not see, why we should have them separated, it is an error message.

PR #311, opened 2019-04-01
added discovered hosts widget for dashboard

vijay8451:
- Added Discovered Host widget .

### DIFF
--- a/airgun/views/dashboard.py
+++ b/airgun/views/dashboard.py
@@ -1,4 +1,5 @@
 from widgetastic.widget import (
+    Table,
     Text,
     View,
     Widget,
@@ -98,6 +99,11 @@ class DashboardView(BaseLoggedInView):
         """
         self.searchbox.search(query)
         return self.read()
+
+    @View.nested
+    class DiscoveredHosts(View):
+        ROOT = ".//li[@data-name='Discovered Hosts']"
+        hosts = Table('.//table')
 
     @View.nested
     class HostConfigurationStatus(View):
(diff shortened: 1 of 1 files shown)


airgun/views/dashboard.py
```diff
@@ -99,6 +100,11 @@ def search(self, query):
         self.searchbox.search(query)
         return self.read()
 
+    @View.nested
+    class DiscoveredHosts(View):
+        ROOT = ".//li[@data-name='Discovered Hosts']"
+        hosts = Table('.//table')
+
```
omkarkhatavkar: @vijay8451 thanks for adding airgun PR :+1: 

Two Queries : 
1. Can you please also add the widget header as well?  It is a link containing the number of hosts ( case for assertion). 
2. There is also navigation from widget to details page, Do you think we need to add that flow as user will be doing that mostly ? 

download

vijay8451: @omkarkhatavkar thanks ..  
for queries , 1) Added widget in last commit 2) not looks that much usable one.

vijay8451:
> Strong + for first point in Omkar comment.
> ACK pending comment and check what PR is going to be first merged: mine or yours to resolve `Table` import conflict

Added the new widget for first point , And about `Table` so shall resolve if conflict occur

oshtaier:
sorry, not sure why we do not have conflict with import:
https://github.com/SatelliteQE/airgun/blob/master/airgun/views/dashboard.py#L2
Maybe it is necessary to do rebase
Other than that PR is ready to be merged

vijay8451:
> sorry, not sure why we do not have conflict with import:
> https://github.com/SatelliteQE/airgun/blob/master/airgun/views/dashboard.py#L2
> Maybe it is necessary to do rebase
> Other than that PR is ready to be merged

@oshtaier not sure why this is not showing and ideally that should occur.
I have fixed that now and good to merge it.

PR #298, opened 2019-03-04
Make ProvisioningTemplate inherit from Template. Add ReportTemplate also inheriting from Template.

lhellebr:

PR #314, opened 2019-04-08
Re-implement template functionality

oshtaier:
Initial PR: https://github.com/SatelliteQE/airgun/pull/298
Critical reasons for that PR:
1) Initial PR added redundant layer of abstraction that applied only for single case
2) ./entities and ./views folders contain only modules for entities and views (and that is pretty much logical), but initial PR introduced `template.py` per each folder that has no logic for any specific entity
3) Initial PR introduced multiple inheritance from different views, so add more complexity into the code for no reason
4) Differences between report and provisioning templates introduce many unnecessary overrides and redundancies. With time it will be very hard to track all of them and other people who work with framework should not care about that

```
pytest tests/foreman/ui_airgun/test_reporttemplates.py 
collecting ... 2019-04-08 13:50:10 - conftest - DEBUG - BZ deselect is disabled in settings

collected 16 items                                                                                                                                                                           

tests/foreman/ui_airgun/test_reporttemplates.py ssssss.ssssssss.                                                                                                                       [100%]
==================================================================== 2 passed, 14 skipped, 14 warnings in 212.24 seconds ==================================================================
```
```
pytest tests/foreman/ui_airgun/test_provisioningtemplate.py 
collecting ... 2019-04-08 14:00:15 - conftest - DEBUG - BZ deselect is disabled in settings

collected 2 items                                                                                                                                                                            

tests/foreman/ui_airgun/test_provisioningtemplate.py ..                                                                                                                                [100%]
========================================================================== 2 passed, 19 warnings in 256.61 seconds ========================================================================
```


oshtaier:
@lhellebr @mirzal @schlupov 
To people who worked on initial PR. Please make sure to look on the current PR and understand what was wrong with previous approach
At least, you should know about `RemovableWidgetsItemsListView` widget and its usage as we re-implemented whole idea with Inputs tabs a long time ago

lhellebr:
Actually, this is not true because there are also at least job_templates. Effectively reverting my commit will lead to duplicating code in entities that are almost the same. Actually, these are also one entity in Satellite itself.
I agree with point 2, though.
Overally, I don't think this was a good thing to do.

oshtaier:
Not sure what you mean by comment about `job_templates` as you didn't touch it in your initial PR, so true or not true is out of scope of our discussion.
Anyway, that 'almost the same' is whole reason we have isolated implementation per each entity.

lhellebr:
I didn't touch it because my main goal was to add reporting. And when I was doing it, I wanted to do it correctly. I could fix job_templates sometime later, because, as you said, those were out of scope.

"Almost the same" is an absolutely typical reason for subclassing, isn't it? Now, when something changes in templates in Satellite, it changes in all 3 types and we will have to do the same changes in the code in 3 places. That is violation of one of the most basic principles of clean code.

I understand why you didn't like a new file in "entities" directory that is actually not entity. But couldn't we find a better solution?

oshtaier:
and if we get changes only in two, but not all three, so you will override 99% of code or what? Sounds like fundamental mistake in architecture, isn't it?

oshtaier:
And there are more points. Do you understand why next statements are bad?:
```
class ProvisioningTemplateCreateView(TemplateCreateView, ProvisioningTemplateDetailsView):
```
or
```
class ProvisioningTemplateEntity(TemplateEntity):
    pass
```

lhellebr:
I understand the pitfalls but I figured those are better left in the code for the sake of consistency. Also, multiple inheritance is not apriori a bad thing - in this case, it is pretty understandable.

Let's conclude that both approaches have their pros and cons. Since airgun is "your" framework and it's not such a big issue and this PR is merged already, let's go with yours :)

oshtaier:
It is not "mine" framework and anyway we are waiting for one more reviewer =)

abalakh:
Overall i like an idea of removing redundant code when the entities are the same. But since they aren't , multiple inheritance is probably not the best way, plus such 'exception' stands out of our common approaches introducing some unnecessary complexity. Imo we could try python mixins instead, but at this point it will require fundamental changes to framework (it shouldn't be done for 1 entity only, but framework itself should be designed accordingly) so i'm not so sure if that's worth it atm.
Good thing is you guys both agreed we can proceed with current approaches so i don't have to convince anyone, as both approaches have pros and cons and i don't have strong point against any of them (except in case of moving similar code outside we'd definitely have to find better way to do that).

mirzal:
@lhellebr can you add ACK if you are OK with these changes? Reading comments, it seems that you and Alex reached agreement, but I wouldn't like to merge these changes if there is some fundamental concern from you.

lhellebr:
I still think this will be change to the worse. But I understand the reasons for this PR and that it's a matter of opinion. I can confirm this will likely not break anything.

oshtaier:
@mirzal can you merge please?

PR #315, opened 2019-04-09
Expand oscap views

oshtaier:


mirzal:
Looks good.

Do we have any OpenSCAP tests that upload files?

PR #319, opened 2019-04-18
Add audit entity and widget for table without headers

oshtaier:


rplevka:
quick question:
does this support expanding the individual rows and does it also parse the table inside each of them?
| old value | new value |

Also, how about special kind of audit records which contain the ACE editor, e.g. audits for Provisioning templates?

oshtaier:
@rplevka 
1) Yes and that is covered in every test like here (expanding):
https://github.com/SatelliteQE/robottelo/pull/6865/files#diff-fb92185da8432f6160876b5f596eaee3R72
and here(old and new values):
https://github.com/SatelliteQE/robottelo/pull/6865/files#diff-fb92185da8432f6160876b5f596eaee3R139

Of course, solution is not ideal as it requires to specify good enough search criteria to leave only one row in the table, but at least we can operate with any data inside of it

2) Yes, we can add support for ACE editor, but we need to slightly improve logic to not raise exception when it is not present on the page as it present only for templates + it application doesn't work at that moment for create event:
image

Only first letter from word 'test' is displayed.

rplevka:
@oshtaier thanks Alex, it's much better than I expected!

PR #321, opened 2019-04-24
Add trend entity

oshtaier:


oshtaier:
Do not cover actual trend graphs in scope of that PR, because it is out of scope of porting efforts and basically such efforts can be huge and not cost efficient for low priority functionality units. Briefly, I have no idea how to cover charts like this with such type of DOM

airgun/views/trend.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+class TrendsView(BaseLoggedInView):
+    title = Text("//h1[text()='Trends']")
+    new = Text("//a[contains(@href, '/trends/new')]")
+    table = Table(
+        './/table',
+        column_widgets={
+            'Name': Text('./a'),
+            'Action': ActionsDropdown("./div[contains(@class, 'btn-group')]"),
```
vijay8451: I think above would match the two elements as below are classes contain the class `btn-group` in it:
```
class="btn-group"
class="dropup pagination-pf-pagesize btn-group"
```
Can we use below one it provides the uniq element:
```suggestion
            'Action': ActionsDropdown("./div[@class='btn-group']"),
```

airgun/entities/trend.py
```diff
@@ -0,0 +1,103 @@
...
+class TrendEntity(BaseEntity):
+
+    def create(self, values):
+        """Create new trend in the system
+
+        Please note that we have static amount of Trend types and trendables, so it is not
+        possible to generate values for them. In case you want to interact with them through
+        automation, it is must have condition to provide unique elements from corresponding
+        lists. In most cases it will not be possible to re-run tests due condition of
+        uniqueness described above.
```
mirzal: This wording seems a little too strong. If test creates new trend and properly deletes it, it can be re-run multiple times in sequence, can't it? Of course if test fails after creating trend, manual intervention is needed to re-run.

airgun/views/trend.py
```diff
@@ -0,0 +1,62 @@
...
+        return (
+            breadcrumb_loaded
+            and self.breadcrumb.locations[0] == 'Trends'
+            and self.breadcrumb.read() == 'Create Trend'
+        )
+
+
+class TrendEditView(BaseLoggedInView):
+
+    breadcrumb = BreadCrumb()
```
mirzal: Not blocking: this empty line isn't needed.

airgun/entities/trend.py
oshtaier: yes, that means if you delete -> you get your unique element back

airgun/views/trend.py
oshtaier: nope, it will not. In `./div...` we have `.` before `/` that means we have relative path for current ROOT which is `./table`, so we looking inside Table element only

vijay8451: ahh , right :)

oshtaier:
Thanks anyway for comment. That help to find issue that will make test fail every time on new instance

oshtaier:
and btw it will not work with `/div[@class='btn-group']` as locators for ActionDropdown are very complex. Weird in some way, but it is what it is =)

oshtaier:
```
pytest tests/foreman/ui_airgun/test_trend.py 
collecting ... 2019-04-24 17:10:06 - conftest - DEBUG - BZ deselect is disabled in settings

collected 1 item                                                                                                                                                                             

tests/foreman/ui_airgun/test_trend.py .                                                                                                                                                [100%]

====================================================================================== warnings summary ======================================================================================
tests/foreman/ui_airgun/test_trend.py::test_positive_end_to_end
  /usr/local/lib/python3.6/site-packages/selenium/webdriver/remote/webdriver.py:796: DeprecationWarning: use driver.switch_to.alert instead
    warnings.warn("use driver.switch_to.alert instead", DeprecationWarning)
  /usr/local/lib/python3.6/site-packages/selenium/webdriver/remote/webdriver.py:796: DeprecationWarning: use driver.switch_to.alert instead
    warnings.warn("use driver.switch_to.alert instead", DeprecationWarning)

=========================================================================== 1 passed, 2 warnings in 75.29 seconds ==========================================================================
```

oshtaier:
@mirzal @vijay8451 could you please ack and merge?

vijay8451:
> and btw it will not work with `/div[@class='btn-group']` as locators for ActionDropdown are very complex. Weird in some way, but it is what it is =)

strange as same is able to find the uniq element at my end.

oshtaier:
@vijay8451 yes, you can find it in browser, but if you put it in code and run - you will get an exception. I feel that relative paths that based on this locator further will not be found

vijay8451:
> @vijay8451 yes, you can find it in browser, but if you put it in code and run - you will get an exception. I feel that relative paths that based on this locator further will not be found

I saw the [1] and found it only use the format `./div[contains(@class, 'btn-group')]` and not flexible to go with `./div[@class='btn-group']` ..

 [1] https://github.com/SatelliteQE/airgun/blob/master/airgun/widgets.py#L573-L576

oshtaier:
@vijay8451 proposed locator is fine when need to click on button part of element, but raise exception when click on dropdown. Maybe, it can be improved in some way, but it is not important at that moment

PR #322, opened 2019-04-30
reporttemplate.is_locked

lhellebr:


lhellebr:
What do you mean, my approach doesn't work for unlocked entities? I think it does. I have tests that pass using lock, unlock and read in both states.

airgun/entities/report_template.py
```diff
@@ -55,6 +55,14 @@ def unlock(self, entity_name):
         view.flash.assert_no_error()
         view.flash.dismiss()
 
+    def is_locked(self, entity_name):
+        """Check if report template is locked for editing"""
+        view = self.navigate_to(self, 'All')
+        view.search(entity_name)
+        return "This template is locked for editing." in \
+            view.table.row(name=entity_name)['Locked'].widget.\
+            browser.element('.').get_property('innerHTML')
```
mirzal: Text widget has `text` property, so you should be able to use `view.table.row(name=entity_name)['Locked'].text` instead. Did you try? Did you encounter any errors?

lhellebr: Yes, text and innerHTML are different things. Text "This template is locked for editing." is actually a HTML tag parameter, not a text node:
```
<span class="glyphicon glyphicon-lock" title="" data-original-title="This template is locked for editing.">
        </span>
```
Also, I can't just directly access the `<span>` node because it is only there if the rt is locked. I don't really like my solution, either, but I don't know any better one and I think it's not that awful after all.

lhellebr:
```

collected 5 items                                                                                      

```

airgun/entities/report_template.py
mirzal: I took a closer look at this and now better understand your problem and why you have chosen this solution.

I have some ideas for cleaner solutions, like making 'locked' property of reporttemplate entity, or implementing custom widget for this column, but... nah, this has been opened long enough.

oshtaier:
okay, finally back from all these PTOs

Question is still actual. That element has its own unique appearance, so not sure why idea with separate widget was ignored

lhellebr:
> Question is still actual. That element has its own unique appearance, so not sure why idea with separate widget was ignored

Mainly because https://github.com/SatelliteQE/airgun/pull/322#discussion_r286929932 , there is no element for unlocked entities.

lhellebr:
@mirzal ACKed already, is this OK to get merged, @oshtaier ?

oshtaier:
Okay, I just wanted to be able to re-use that functionality. And with proposed solution, I will need to copy paste that method everywhere, for every template like provision template and etc.

Okay, if no other options then just fix formatting and let's merge it. 
something like:
```
return "This template is locked for editing." in view.table.row(
    name=entity_name)['Locked'].widget.browser.element('.').get_property('innerHTML')
```

PR #323, opened 2019-05-02
Download manager fix

lhellebr:


oshtaier:
Please provide reasons for that change and test result to prove that change is fixing things.

mirzal:
I can confirm that on Chrome 73, which is not even the latest version, these objects already have `fileUrl` instead of `file_url`. So yeah, if it's not needed yet, it will be eventually.

But we need to somehow coordinate it with upgrade of Chrome that automation uses. Is there any task for that anywhere? Something that we could subscribe to, to be in the loop? Or @lhellebr , are you going to keep track of that and follow up here when right time comes?

oshtaier:
yeap, we have to have static browser version as we usually did in the past, so we do not get completely different results for different environments. Let's say that Jenkins environment should be an example for everyone.

lhellebr:
@mirzal , this is a  change we can do anytime in Jenkins. Also, doing it will cause some tests to fail until this PR is merged. So it's up to us (as in all automation stakeholders).

lhellebr:
```

collected 5 items                                                                                      

```

oshtaier:
So that PR will fix code for your version of Chrome, but instead make fail tests on mine? I don't mind to move on newer version, but that will require some global re-test. If we all agreed to move, I will ACK that PR on 110%

lhellebr:
Exactly, that's why I said this should have been a Draft PR and I don't expect it to be merged unless we decide to use newer Chrome in automation.

oshtaier:
okay, sorry then =)

tstrych:
This PR should be closed, change was already merged in [this PR](https://github.com/SatelliteQE/airgun/pull/405)

lhellebr:
Right, Tomas's PR should solve the issue. Closing.

PR #324, opened 2019-05-02
Reporttemplate functions

lhellebr:


airgun/views/report_template.py
```diff
@@ -85,3 +85,17 @@ def is_displayed(self):
             and self.breadcrumb.locations[0] == 'Report Templates'
             and self.breadcrumb.read() == 'Create Template'
         )
+
+class ReportTemplateGenerateView(BaseLoggedInView):
+    breadcrumb = BreadCrumb()
+    submit = Text('//input[@name="commit"]')
```
oshtaier: please add two fields that form has

lhellebr: Input fields are template-specific

oshtaier: Then it is necessary describe them. At least most commonly used through conditional view

lhellebr:
Depends on https://github.com/SatelliteQE/airgun/pull/325

airgun/views/report_template.py
lhellebr: It would be awful to hardcode values for some report templates. Instead, I coded EditableEntryMultiTextInput widget (PR https://github.com/SatelliteQE/airgun/pull/325) which is more general. Seems reasonable?

airgun/widgets.py
```diff
@@ -1321,6 +1321,42 @@ class EditableEntryMultiCheckbox(EditableEntry):
     edit_field = CheckboxGroup(locator='.//form')
 
 
+class EditableEntryMultiTextInput(GenericLocatorWidget):
```
mirzal: This is the same thing as in #325 , right? If so, I think it should be removed from this PR and this PR should be merged only after #325 is merged.

lhellebr: I made that commit part of this PR - due to some issues with branching, it had to be done. I think the best way to handle this would be to close https://github.com/SatelliteQE/airgun/pull/325/commits and accept this PR as-is. Agree?

oshtaier: Hmm, if you just copy/pasted content of other PR here, so yeah, we can close #325 and address comments in it there, but again, you adding mess where it should not be added in the first place.

lhellebr: Actually I rebased this PR to contain the other PR. Let's close that one, I will resolve comments from @ldjebran here.

lhellebr:
```

collected 5 items                                                                                      

```

airgun/entities/report_template.py
```diff
@@ -55,6 +56,27 @@ def unlock(self, entity_name):
...
 
+    def export(self, entity_name):
+        """Export report template.
+
+         :return str: path to saved file
+        """
+        view = self.navigate_to(self, 'All')
+        view.search(entity_name)
+        view.table.row(name=entity_name)['Actions'].widget.fill('Export')
+        view.flash.assert_no_error()
```
oshtaier: that is redundant as it just can't happen

airgun/widgets.py
```diff
@@ -1321,6 +1321,45 @@ class EditableEntryMultiCheckbox(EditableEntry):
     edit_field = CheckboxGroup(locator='.//form')
 
 
+class TextInputsGroup(GenericLocatorWidget):
+    """
+    A set of text inputs
+    """
+    ITEMS_LOCATOR = './/label'
```
oshtaier: that sounds too ambiguous. What are `ITEMS`? Is it TEXTINPUT` group?

airgun/entities/report_template.py
lhellebr: How do you know? I'm not saying that you're wrong but how can you know some error won't happen, let alone in the future releases?

oshtaier: Once you do `view.table.row(name=entity_name)['Actions'].widget.fill('Export')` you lose control on a browser completely as opened dialog block Selenium from access till it closed, so you can't read anything no matter what happen

airgun/widgets.py
lhellebr: Items are field lablels. TEXTINPUT_LOCATOR are text inputs found by their label.

But I've found another issue: the locator has changed in 6.6. I'll change the locator so it works for 6.6 and will fix it again for 6.5 during cherrypick, is that ok?

airgun/entities/report_template.py
lhellebr: That makes sense, I'll remove the line.

airgun/widgets.py
oshtaier: that is intended procedure. So can we rename ITEMS to FIELD_LABELS or whatever you like?

lhellebr:
Comments resolved

lhellebr:
@oshtaier , this has already been ACKed by @mirzal , do we really need @ldjebran ?

oshtaier:
Yes, as he commented your PR that you closed. You moved discussion here, so let's have his ACK for moved things. Ping him in IRC and I believe you get response very quickly

airgun/widgets.py
```diff
@@ -1321,6 +1321,46 @@ class EditableEntryMultiCheckbox(EditableEntry):
...
+        return {
+            name: textinput.read()
+            for name, textinput in self.textinputs.items()
+        }
+
+    def fill(self, values):
+        """Fill one of the text inputs
+
+        :param value: string with specification of fields' values
+            Example: value={'inputs': {'Hosts filter': 'name=host11.example.com',
```
ldjebran: please update the doc string as seems inputs key not used in the code.

airgun/entities/report_template.py
```diff
@@ -55,6 +56,26 @@ def unlock(self, entity_name):
...
+        return self.browser.save_downloaded_file()
+
+    def generate(self, entity_name, values={}):
+        """Generate report template
+
+         :return str: path to saved file
+        """
+        view = self.navigate_to(self, 'Generate', entity_name=entity_name)
+        view.fill(values)
+        view.submit.click()
```
ldjebran: is there errors popups in case of error after submitting, validation errors ..., if there is we should check them before return.

lhellebr: @ldjebran , @oshtaier suggested the exact opposite here: https://github.com/SatelliteQE/airgun/pull/324#pullrequestreview-242786044
But yes, there are checks in the form, for example for e-mail format. Can we get to some agreement on what should be done here?

lhellebr: Actually, there is a difference because in 6.6, the report is not downloaded straight away (as it used to in 6.5) but you are redirected to the download page. I suppose that means it IS possible to check for the pop-up.

oshtaier: Yes, we should remove for 'Export' as there is no validation performed here and you get to dialog right away, but for generate, I can see validation, but it is not a pop - up:
image

so not sure what validation we can add, maybe a custom one. Anyway, I feel `save_downloaded_file` should produce an exception in case dialog is absent. We can test that. If yes, we do not need any validation

ldjebran: we can try ValidationErrors from widgets module? to stick with minimum validation errors, as may be useful also for negative tests, and to minimize dealing with the errors when passing to download as they may be very specific.
@oshtaier what do you think?  
anyway I am OK for the positive test cases to crash on download in case of errors and confirm ACK.

oshtaier: I am always like to go with better errors, but if it at least crash on negative scenario, then I am ACK too

ldjebran:
@lhellebr please rebase.

airgun/entities/report_template.py
lhellebr: Is ``view.flash.assert_no_error()`` what we are looking for?

oshtaier: I believe no as we have no `flash` here. Flash errors are ones that appear in top right corner. Screenshot above is for element which directly present in DOM and will not be caught by existing validators. Anyway, I would be happy just see that scenario crash. Exact error is lower priority for me

lhellebr: Your screenshot is what happens on success. However, failure looks like this:

flash

I believe that IS flash error, isn't it?

oshtaier: yeap, that one will be caught =)

oshtaier: so please put that line in the method

oshtaier:
please do rebase

airgun/entities/report_template.py
lhellebr: I did. Now, when I fill in e-mail in incorrect format, the result is:
```
E               AssertionError: assert_no_error: error: 'Could not generate the report, check the form for error messages'
```

airgun/views/report_template.py
```diff
@@ -86,3 +87,21 @@ def is_displayed(self):
             and self.breadcrumb.locations[0] == 'Report Templates'
             and self.breadcrumb.read() == 'Create Template'
         )
+
+
+class ReportTemplateGenerateView(BaseLoggedInView):
+    breadcrumb = BreadCrumb()
+    submit = Text('//input[@name="commit"]')
```
oshtaier: please put widget in the order we have them in UI as they will be filled in that order
and more important, where is `Generate at` field which is present in every view

oshtaier: That field will not get to the `TextInputsGroup`, right?

oshtaier: I mean we created `TextInputsGroup` for dynamic content, and that one is static for every report

lhellebr: The Generate at field should be filled extra, not as part of TextInputsGroup, exactly as you say. I will reorder the fields. I don't want to include the Generate At field right now because it's a datetime picker, which
1) is a new thing in the WebUI
2) is somewhat hard to implement
3) is not necessary now
4) will be reworked soon

oshtaier: Sounds good to me, it is definitely large piece of work and out of scope for current PR

oshtaier:
@lhellebr sorry for all the delays and thanks for your patience.

PR #325, opened 2019-05-07
EditableEntryMultiTextInput

lhellebr:


oshtaier:
sorry @lhellebr we have many holidays here in May, so that is why review is delayed
anyway, I looked on the functionality yesterday and yes, fields combination is pretty random and depends on report template type. Proposed solution is close to one we may need. It will be good if someone else look on that PR meanwhile

Only correction at that moment. That entity has nothing to do with Editable entries, but just FieldGroup

lhellebr:
@oshtaier That's fine, we also have many holidays here. And FYI, I will have PTO the next week.

Right. Is it just about renaming to TextInputGroup?

oshtaier:
not sure if that approach is best, so will ask @ldjebran  or @abalakh check that PR

airgun/widgets.py
```diff
@@ -1321,6 +1321,42 @@ class EditableEntryMultiCheckbox(EditableEntry):
     edit_field = CheckboxGroup(locator='.//form')
 
 
+class EditableEntryMultiTextInput(GenericLocatorWidget):
```
ldjebran: TextInput and EditableEntry can  considered to have the same meaning here, perhaps
LabeledTextInputs
Or TextInputsGroup ...

airgun/widgets.py
```diff
@@ -1321,6 +1321,42 @@ class EditableEntryMultiCheckbox(EditableEntry):
...
+    """
+    ITEMS_LOCATOR = './/label'
+    TEXTINPUT_LOCATOR = './/label[normalize-space(.)="{}"]/following-sibling::div/textarea'
+
+    @cached_property
+    def textinputs(self):
+        labels = [
+            line.text
+            for line in self.browser.elements(self.ITEMS_LOCATOR, parent=self)
+        ]
```
ldjebran: think labels can be as a standalone cached property

lhellebr: I don't understand what you are suggesting.

airgun/widgets.py
lhellebr: Is this just about the class name? If yes then OK, why not.

airgun/widgets.py
lhellebr: @ldjebran

mirzal: I believe @ldjebran means to create `labels` property. Something like that:

```python
    @cached_property
    def labels(self):
        return [
            line.text
            for line in self.browser.elements(self.ITEMS_LOCATOR, parent=self)
        ]

    @cached_property
    def textinputs(self):
        return {
            label: TextInput(self, locator=self.TEXTINPUT_LOCATOR.format(label))
            for label in self.labels
        }
```

mirzal:
@lhellebr We should close this one and continue discussion in #324, is that correct?
Can you close PR yourself, or do you lack necessary permissions?

lhellebr:
Let's continue in https://github.com/SatelliteQE/airgun/pull/324 containing this commit

PR #327, opened 2019-05-20
Improve host entity

oshtaier:


airgun/entities/host.py
```diff
@@ -70,14 +70,18 @@ def update(self, entity_name, values):
         view.flash.assert_no_error()
         view.flash.dismiss()
 
-    def delete(self, entity_name):
+    def delete(self, entity_name, get_alert_text=False):
```
ldjebran: To support more tests We can add a confirm=True kwargs :
in case of True  we handle the alert with Ok, otherwise we push cancel button,
and always return the alert text.

airgun/views/host.py
```diff
@@ -232,6 +239,14 @@ class ansible_roles(SatTab):
 
         resources = MultiSelect(id='ms-host_ansible_role_ids')
 
+    @View.nested
+    class virtual_machine(SatTab):
```
ldjebran: We can use the compute resource profile view as it's the same for each compute resource type, this will need some code code refactoring in compute resource with moving the VMs views to a base class to compute resource profile, this will allow to use that VM views in host , compute resource profile and  profile entities.

ldjebran: The class concerned is ResourceProviderProfileView

airgun/views/computeresource.py
```diff
@@ -241,15 +241,29 @@ class containers(SatTab, SearchableViewMixin):
...
+    network = FilteredDropdown(id="bridge")
+    nic_type = FilteredDropdown(id="model")
+
+
+class ComputeResourceLibvirtProfileStorageItem(GenericRemovableWidgetItem):
+    """Libvirt Compute Resource profile "Storage" item widget"""
+    storage_pool = FilteredDropdown(id="pool_name")
+    size = TextInput(locator=".//input[contains(@id, 'capacity')]")
+    storage_type = FilteredDropdown(id="format_type")
+
```
vijay8451: can we include the `Allocation (GB)`  too  ..
what do you think ?

airgun/views/host.py
```diff
@@ -99,6 +102,13 @@ def override(self, value=True):
             self.remove_override_button.click()
 
 
+class ComputeResourceLibvirtProfileStorageItem(GenericRemovableWidgetItem):
+    """Libvirt Compute Resource profile "Storage" item widget"""
+    storage_pool = FilteredDropdown(id="pool_name")
+    size = TextInput(locator=".//input[contains(@id, 'capacity')]")
+    storage_type = FilteredDropdown(id="format_type")
+
```
vijay8451: same as above

airgun/views/computeresource.py
oshtaier: @vijay8451 Let's me move that out of the scope of that PR as it is probably require separate widget for it:
image

oshtaier:
Created issue for that:
https://github.com/SatelliteQE/airgun/issues/330

airgun/views/computeresource.py
vijay8451: @oshtaier yeah it looks more a new widget but I have seen same used by multiple Compute resource hence worth to have that. 
as new issue #330 opened , I am fine with it :)

lhellebr:
@oshtaier I believe this PR has broken create_fake_host:
``E               ValueError: Could not find a corresponding registered view.`` where the view is supposed to be ``provider_content``.

PR #329, opened 2019-05-21
Update NavigateStep.go, add use_resetter and wait_for_view kwargs

mshriver:
These kwargs were proposed as update to navmazing, but rejected for core.

Pull in the Navmazing.NavigateStep.go code, adapt to local use of view/am_i_here

### DIFF
--- a/airgun/navigation.py
+++ b/airgun/navigation.py
@@ -85,11 +85,50 @@ class NavigateStep(navmazing.NavigateStep):
         """Wrapper around :meth:`navmazing.NavigateStep.go` which returns
         instance of view after successful navigation flow.
 
+        Optional kwargs:
+            use_resetter: boolean control over resetter being called (default True)
+            wait_for_view: control over am_i_here being called after step (default False)
+                            Pass a value for duration of wait_for loop (str or int)
+
+        :raises:
+            wait_for.TimedOutError if wait_for_view is passed and am_i_here does not return true
+
         :return: view instance if class attribute ``VIEW`` is set or ``None``
             otherwise
         """
-        super(NavigateStep, self).go(_tries=_tries, *args, **kwargs)
+        _tries += 1
+        use_resetter = kwargs.pop('use_resetter', True)
+        wait_for_view = kwargs.pop('wait_for_view', None)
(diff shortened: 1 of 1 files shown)


mshriver:
@oshtaier The navmazing core update was rejected, thoughts on pulling in the code from navmazing as I've done here, including my updates to support control of `resetter` and waiting for the view after navigation?

oshtaier:
Does it make sense to change code just in that way?:
```
wait_for_view = kwargs.pop('wait_for_view', None)
super(NavigateStep, self).go(_tries=_tries, *args, **kwargs)
view = self.view if self.VIEW is not None else None
if wait_for_view is not None and view is not None:
    self.logger.info("NAVIGATE: Waiting for am_i_here for {}".format(self._name))
    wait_for(
        self.am_i_here, func_args=args, func_kwargs=kwargs,
        logger=self.logger,
        timeout=wait_for_view
)
```
I feel that we do not use `resetter` much and it doesn't worth to remove `super` due that small fact. I like that we still linked to navmazing.
What do you think @mshriver ?

mshriver:
@oshtaier absolutely up to you! Figured you would want control over use_resetter as well, and that we might as well pull in the source so it can be modified to your liking here for your navigation loop.

Navmazing itself is very unlikely to change its implementation of `go`, as its intended to be overwritten, with the existing implementation as a basic example of the flow.

We actually wrap every method call within go with a function that checks for failures in the browser along the way - making sure there aren't alerts, that JS async is 'complete', etc.

Wanted to get this conversation moving, feel free to close this PR and open a separate one with just the content you want, like support for wait_for_view.  You could even flip its implementation, defaulting to some timeout, only skipping wait when `False` or `None` is explicitly passed. This is just the implementation choice that MIQ made - to default to _not_ waiting and let the user explicitly wait for it.

lhellebr:
Nothing changed here for a long time, do you still plan to finish this sometime, @mshriver?

latran:
@mshriver Any updates on this?

lhellebr:
I suggest closing this. @mshriver ?

lhellebr:
No activity for a long time and no protests against closing -> closing.

Issue #331, opened 2019-05-30
Fix manage_packages() in hostcollection.py


swadeley:
When running a test that calls manage_packages() in hostcollection.py, and using the default 'via Katello Agent' method, the Update Packages dialogue does not close.

PR #332, opened 2019-05-30
call done.click() if via Katello Agent (#331)

swadeley:
Proposing an if statement to close the Update Packages dialogue. I did try an else statement at the end but got errors.

Edit: I found this was required while working on https://github.com/SatelliteQE/robottelo/pull/6974

### DIFF
--- a/airgun/entities/hostcollection.py
+++ b/airgun/entities/hostcollection.py
@@ -97,6 +97,8 @@ class HostCollectionEntity(BaseEntity):
         view.apply_action(action, action_via=action_via)
         view.flash.assert_no_error()
         view.flash.dismiss()
+        if action_via == 'via Katello Agent':
+            view.done.click()
         if action_via == 'via remote execution - customize first':
             # After this step the user is redirected to remote execution job
             # create view.


swadeley:
Hello

The `manage_packages()` function is used a number of times in `test_hostcollection.py`. But here we are only interested in those that do not use `action_via` (so that it defaults to using katello agent). I see  `test_positive_install_package` is a test without `action_via`. That test passes.

I then ran a test using the complete `tests/foreman/ui_airgun/test_hostcollection.py` file. All test pass except `test_positive_install_errata` which does not use, does not call, `manage_packages()`.  I think that could be caused by the bug mentioned in https://github.com/SatelliteQE/robottelo/pull/6974. I will apply the update and test again.

mirzal:
@swadeley I think it will be enough if you add output of something like:
```
pytest -v -k 'test_positive_install_package or test_positive_remove_package' tests/foreman/ui_airgun/test_hostcollection.py
```

I mean, it's patch adding two lines, it should have been merged already. Let's not make it longer and harder than necessary.

swadeley:
robottelo]$ pytest -v tests/foreman/ui_airgun/test_hostcollection.py::test_positive_install_package
============================== test session starts =============================
platform linux -- Python 3.6.6, pytest-4.0.2, py-1.8.0, pluggy-0.12.0 -- .virtualenvs/robottelo-master/bin/python3
cachedir: .pytest_cache
rootdir: /sat6/robottelo, inifile:
plugins: services-1.3.1, forked-1.0.2, xdist-1.28.0, cov-2.7.1, mock-1.10.0
collecting ... 2019-06-04 06:17:11 - conftest - DEBUG - BZ deselect is disabled in settings

collected 1 item                                                                                        

tests/foreman/ui_airgun/test_hostcollection.py::test_positive_install_package PASSED              [100%]

======================== 1 passed, 2 warnings in 662.97 seconds ===================

swadeley:
thank you @mirzal

oshtaier:
@mirzal It doesn't matter how many lines we have in PR, if you change basic framework, it is required to provide test result

PR #333, opened 2019-06-04
Improve gpg key entity

oshtaier:


airgun/views/contentcredential.py
```diff
@@ -68,9 +68,15 @@ class details(SatTab):
         repos = ReadOnlyEntry(name='Repositories')
 
     @View.nested
-    class products(SatTab):
-        table = SatTable(locator=".//table")
+    class products(SatTab, SearchableViewMixin):
+        table = SatTable(
```
omkarkhatavkar: Do you think we need to use `SatTable` here or `Table`?

airgun/views/contentcredential.py
```diff
@@ -68,9 +68,15 @@ class details(SatTab):
...
+            './/table',
+            column_widgets={'Name': Text('./a')}
+        )
 
     @View.nested
-    class repositories(SatTab):
-        table = SatTable(locator=".//table")
+    class repositories(SatTab, SearchableViewMixin):
+        table = SatTable(
```
omkarkhatavkar: Do you think we need to use `SatTable` here or `Table`?

oshtaier:
@omkarkhatavkar @vijay8451 Addressed comments

PR #341, opened 2019-06-19
 command_executor can be configured for selenium/remote 

eko999:
This is an enhancement for the selenium/remote, the scenario is:
if I have a remote selenium webdriver, such as http://10.66.xxx.xxx:4444/wd/hub, and the browserName is firefox, I want to use it for robottelo UI testing, I can configure it in robottelo.properties: 
[robottelo]
browser=selenium
webdriver=remote
webdriver_desired_capabilities=platform=Linux,browserName=firefox
command_executor=http://10.66.xxx.xxx:4444/wd/hub

The key point is I don't need to care about how to deploy my remote webdriver and where is the webdriver,  you only need to provide an available command_executor to access the remote webdriver.

### DIFF
--- a/airgun/browser.py
+++ b/airgun/browser.py
@@ -202,7 +202,8 @@ class SeleniumBrowserFactory(object):
             elif browserName == 'firefox':
                 capabilities = webdriver.DesiredCapabilities.FIREFOX.copy()
             elif browserName == 'ie':
-                capabilities = webdriver.DesiredCapabilities.INTERNETEXPLORER.copy()
+                capabilities = (
+                        webdriver.DesiredCapabilities.INTERNETEXPLORER.copy())
             elif browserName == 'edge':
                 capabilities = webdriver.DesiredCapabilities.EDGE.copy()
             else:
@@ -214,8 +215,8 @@ class SeleniumBrowserFactory(object):
                 capabilities.update(
                     vars(settings.webdriver_desired_capabilities))
             self._webdriver = webdriver.Remote(
-                command_executor = settings.selenium.command_executor,
-                desired_capabilities = capabilities
+                command_executor=settings.selenium.command_executor,
+                desired_capabilities=capabilities
             )
             idle_timeout = settings.webdriver_desired_capabilities.idleTimeout
             if idle_timeout:
--- a/airgun/browser.py
+++ b/airgun/browser.py
@@ -196,10 +196,31 @@ class SeleniumBrowserFactory(object):
             self._webdriver = webdriver.PhantomJS(
                 service_args=['--ignore-ssl-errors=true'])
         elif self.browser == 'remote':
-            capabilities = vars(settings.webdriver_desired_capabilities)
+            browserName = settings.webdriver_desired_capabilities.browserName
+            if browserName == 'chrome':
+                capabilities = webdriver.DesiredCapabilities.CHROME.copy()
+            elif browserName == 'firefox':
+                capabilities = webdriver.DesiredCapabilities.FIREFOX.copy()
+            elif browserName == 'ie':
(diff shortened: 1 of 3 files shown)


airgun/browser.py
```diff
@@ -196,10 +196,32 @@ def _get_selenium_browser(self):
             self._webdriver = webdriver.PhantomJS(
                 service_args=['--ignore-ssl-errors=true'])
         elif self.browser == 'remote':
-            capabilities = vars(settings.webdriver_desired_capabilities)
+            browserName = settings.webdriver_desired_capabilities.browserName
```
mirzal: Do you need to introduce new variable and setting here? Can't you re-use `self.browser` (defaults to `settings.selenium.webdriver`)?

webdriver capability `browserName` will be set by `webdriver.DesiredCapabilities.<provider>.copy()` anyway.

airgun/browser.py
```diff
@@ -196,10 +196,32 @@ def _get_selenium_browser(self):
...
+            elif browserName == 'edge':
+                capabilities = webdriver.DesiredCapabilities.EDGE.copy()
+            else:
+                raise ValueError(
+                    '"{}" browserName is not supported. Please use one of {}'
+                    .format(browserName, ('chrome', 'firefox', 'ie', 'edge'))
+                )
+            if settings.webdriver_desired_capabilities:
+                capabilities.update(
+                    vars(settings.webdriver_desired_capabilities))
```
mirzal: Is this whole part (including `idle_timeout` below) so different from code in `_get_saucelabs_browser` that we should have it independent? Can't we extract it to separate private method and just run for both saucelab and remote? In fact, saucelab is just a special case of remote.

Also, we always set `download.prompt_for_download` to False and pass disable-web-security to Chrome, and set `acceptSslCerts` and `javascriptEnabled` to True for Edge. I think you want to do it in this path, too.

As a side note, this change shows that `SeleniumBrowserFactory` is difficult to extend and there is a lot of code duplication. It probably should be rewritten to use separate classes, so Sauce and Docker can simply inherit from Remote (and Remote should inherit from normal local webdriver, where all capabilities are set?). But refactoring is probably out of scope for this PR.

airgun/browser.py
jyejare: +1 to what @mirzal said.

eko-index: Because there are different browsers used for remote, in my opinion, remote should not be a value for 'settings.selenium.webdriver', it should be a provider for 'settings.selenium.browser' 
provider = ('selenium', 'docker', 'saucelabs', 'remote')
browser = ('chrome', 'firefox', 'ie', 'edge', 'phantomjs')

airgun/browser.py
eko-index: Yes, I absolutely agree with you, the capabilities for different browser can be public, Remote/Sauce can use the same capabilities, I'd like to create a new PR for this.

airgun/browser.py
mirzal: > Because there are different browsers used for remote, in my opinion, remote should not be a value for `settings.selenium.webdriver`, it should be a provider for `settings.selenium.browser`:
> provider = ('selenium', 'docker', 'saucelabs', 'remote')
> browser = ('chrome', 'firefox', 'ie', 'edge', 'phantomjs')

Fully agreed. In fact, my comment made sense only if you assume such situation - of course you can't use `self.browser` in these ifs now, because `self.browser` is `remote`.

Is moving `remote` to `provider` something you can work on as part of this PR?
As you have shown, `remote` hardly works right now at all, so I'm not worried too much about backwards compatibility here.

eko999:
I will close this PR, and create a new one as above discussion.

PR #343, opened 2019-06-20
Added search for ldap auth

ntkathole:
Fixes https://github.com/SatelliteQE/robottelo/issues/6687

### DIFF
--- a/airgun/entities/ldap_authentication.py
+++ b/airgun/entities/ldap_authentication.py
@@ -1,5 +1,5 @@
 from navmazing import NavigateToSibling
-from widgetastic.exceptions import RowNotFound
+from widgetastic.exceptions import RowNotFound, NoSuchElementException
 
 from airgun.entities.base import BaseEntity
 from airgun.navigation import NavigateStep, navigator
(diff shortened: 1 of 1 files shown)


oshtaier:
both `read_table_row` and `search` should be removed. Let's me find time and provide examples. Basically, `Table` widget and our basic `search` should already handle that situation. I think you agree that it will be really bad idea to add same two methods in other 50 modules.

oshtaier:
Maybe LDAP entity doesn't have welcome screen (https://github.com/SatelliteQE/airgun/blob/master/airgun/views/common.py#L407). Hmm

oshtaier:
Anyway, I feel proposed solution is a little bit redundant. Maybe, it will be better to stop at that moment on:
```
    def read_table_row(self, entity_name):
        """Read values for corresponding table row from LDAP Authentication
        title page. Return None in case row is not present in the table
        """
        try:
            row_value = view.table.row(name=entity_name).read()
        except RowNotFound, NoSuchElementException:
            row_value = None
        return row_value
```
It is not perfect solution, but not the worst and better will require some time for investigation

ntkathole:
@oshtaier Thanks. It works perfect.

oshtaier:
It just can catch some other invalid situation =), but we use solution very specifically, so it is not big problem with that 'None'

Issue #249, opened 2018-11-27
 self.browser.clear(self.search_field) doesn't work on foreman search boxes


rplevka:
The `clear` method has no effect on foreman search boxes. The method does not error out though, so the effect of repeated `session.<entity>.search("foo")` is that the search query is being chained.
Workaround for now is to manually aim at the "x" icon next to the search bar that clears the output, Or, the nastier way- navigating away from the page and re-visit it.

abalakh:
Couldn't reproduce for both katello and foreman entities. Tested on 6.5 SNAP4 and SNAP5.
My env: Chrome Version 70.0.3538.102, ChromeDriver 70.0.3538.97.
@rplevka @oshtaier What's your chrome(chromium) and, more importantly, chromedriver version?

oshtaier:
Just to be on the same page. Did you look exactly on a field with 'slash' on a beginning:
![image](https://user-images.githubusercontent.com/10895065/49104624-187aae00-f287-11e8-9d7a-149694be7ae7.png)

abalakh:
@oshtaier yep.
My findings:
Updating to Chrome Version 70.0.3538.110 (latest) didn't change a thing. BUT, after also updating chromedriver to latest - ChromeDriver 2.44.609545 (weird output in comparison to previous one btw) i was able to reproduce it. Seems like an issue in latest chromedriver itself (which doesn't surprise me, there was another regression in previous release).

oshtaier:
I didn't touch my environment for a while, so not sure about the details, but I can reproduce issue constantly

abalakh:
@oshtaier have you disabled chrome autoupdate? Chromedriver supports only few chrome versions, you should keep its version according to your Chrome version, otherwise issues like this may occur.
Ok, so i'm playing with both 
ChromeDriver 70.0.3538.97 (d035916fe243477005bc95fe2a5778b8f20b6ae1)
ChromeDriver 2.44.609545 (c2f88692e98ce7233d2df7c724465ecacfe74df5)
and i can't reproduce it with 70.0.3538.97, but it's always reproducible on 2.44.609545, the issue is definitely there.

oshtaier:
okay, I will check the version tomorrow

abalakh:
Try this one tomorrow:
https://chromedriver.storage.googleapis.com/index.html?path=70.0.3538.97/

vikask18:
It looks like this issue still persists.

My findings so far:
- Navigation across pages clears the search box.
[...]

PR #344, opened 2019-06-24
Enhancement for remote webdriver

eko999:
As discussion before, this PR will:
1). move remote from `self.browser` to `self.provider`
2). add a new function for remote provider:  `_get_remote_browser`
3). add a new function to get the capabilities for remote and sauce providers: ` _get_webdriver_capabilities`

command_executor and browser are required for remote provider, the demo configuration in robottelo.properties is:
[robottelo]
browser=remote
webdriver=firefox
command_executor=http://10.73.xxxx.xxx:4444/wd/hub

### DIFF
--- a/airgun/browser.py
+++ b/airgun/browser.py
@@ -273,7 +273,6 @@ class SeleniumBrowserFactory(object):
         idle_timeout = settings.webdriver_desired_capabilities.idleTimeout
         if idle_timeout:
             self._webdriver.command_executor.set_timeout(int(idle_timeout))
-        # todo: attempt to rename job here
         return self._webdriver
 
     def _get_webdriver_capabilities(self):
--- a/airgun/browser.py
+++ b/airgun/browser.py
@@ -83,10 +83,10 @@ class SeleniumBrowserFactory(object):
         values.
 
         :param str optional provider: Browser provider name. One of
-            ('selenium', 'docker', 'saucelabs'). If none specified -
+            ('selenium', 'docker', 'saucelabs', 'remote'). If none specified -
             :attr:`settings.selenium.browser` is used.
         :param str optional browser: Browser name. One of ('chrome', 'firefox',
-            'ie', 'edge', 'phantomjs', 'remote'). Not required for ``docker``
+            'ie', 'edge', 'phantomjs'). Not required for ``docker``
             provider as it currently supports firefox only. If none specified -
             :attr:`settings.selenium.webdriver` is used.
         :param str optional test_name: Name of the test using this factory. It
@@ -113,10 +113,15 @@ class SeleniumBrowserFactory(object):
             return self._get_saucelabs_browser()
         elif self.provider == 'docker':
             return self._get_docker_browser()
+        elif self.provider == 'remote':
+            return self._get_remote_browser()
         else:
             raise ValueError(
                 '"{}" browser is not supported. Please use one of {}'
-                .format(self.provider, ('selenium', 'saucelabs', 'docker'))
+                .format(
+                    self.provider,
+                    ('selenium', 'saucelabs', 'docker', 'remote')
+                )
             )
 
     def post_init(self):
@@ -147,7 +152,7 @@ class SeleniumBrowserFactory(object):
             or not. Is only used for ``saucelabs`` provider.
         :return: None
         """
-        if self.provider == 'selenium':
+        if self.provider == 'selenium' or self.provider == 'remote':
             self._webdriver.quit()
             return
         elif self.provider == 'saucelabs':
@@ -194,17 +199,12 @@ class SeleniumBrowserFactory(object):
         elif self.browser == 'phantomjs':
             self._webdriver = webdriver.PhantomJS(
                 service_args=['--ignore-ssl-errors=true'])
-        elif self.browser == 'remote':
-            capabilities = vars(settings.webdriver_desired_capabilities)
-            self._webdriver = webdriver.Remote(
-                desired_capabilities=capabilities
-            )
         if self._webdriver is None:
             raise ValueError(
                 '"{}" webdriver is not supported. Please use one of {}'
                 .format(
-                    self.provider,
-                    ('chrome', 'firefox', 'ie', 'edge', 'phantomjs', 'remote')
+                    self.browser,
+                    ('chrome', 'firefox', 'ie', 'edge', 'phantomjs')
                 )
             )
         return self._webdriver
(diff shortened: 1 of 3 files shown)


rplevka:
ACK!
I think this pretty much re-implements the NACKed:

eko999:
> ACK!
> I think this pretty much re-implements the NACKed:
> https://github.com/SatelliteQE/airgun/pull/223/files
> 
> But we were about to resurrect it.

Yes, I think the most difference is I changed the remote as a provider. The key point is everybody can run the UI cases easily with a valid command_executor url.

I created this PR because I'd like to have more choices to run the UI cases, especially when running UI cases in a Jenkins slave which is a container, the remote provider should be another better choice besides Saucelab.

airgun/browser.py
```diff
@@ -298,6 +261,58 @@ def _get_docker_browser(self):
...
+        Note: should not be called directly, use :meth:`get_browser` instead.
+        """
+        self._webdriver = webdriver.Remote(
+            command_executor=settings.selenium.command_executor,
+            desired_capabilities=self._get_webdriver_capabilities()
+        )
+        idle_timeout = settings.webdriver_desired_capabilities.idleTimeout
+        if idle_timeout:
+            self._webdriver.command_executor.set_timeout(int(idle_timeout))
+        # todo: attempt to rename job here
```
mirzal: This comment seems to be copied from `_get_saucelab_browser`. Does it make sense in context of remote browser as well?

airgun/browser.py
```diff
@@ -298,6 +261,58 @@ def _get_docker_browser(self):
         self._webdriver = self._docker.webdriver
         return self._webdriver
 
+    def _get_remote_browser(self):
+        """Returns remote webdriver instance of selected ``browser``.
+
+        Note: should not be called directly, use :meth:`get_browser` instead.
+        """
+        self._webdriver = webdriver.Remote(
+            command_executor=settings.selenium.command_executor,
```
mirzal: What will happen if we pass `None` here (which is the default)? Will it throw exception with meaningful message?
Perhaps we should be a little more defensive, and first ensure that this variable is set?

eko999: When command_executor is None,  webdriver.Remote will try to use ''http://127.0.0.1:4444/wd/hub" as the default command_executor to connect, if the default command_executor is not available, exception message will be raised:
`urllib3.exceptions.MaxRetryError: HTTPConnectionPool(host='127.0.0.1', port=4444): Max retries exceeded with url: /wd/hub/session (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7f5a7ff01080>: Failed to establish a new connection: [Errno 111] Connection refused',))`

I think the exception message is fair, webdriver.Remote has the ability to verify the command_executor's value.

airgun/browser.py
eko999: Yes, `_get_remote_browser` has the same requirement for webdriver_capabilities, as you said, saucelab is also a remote mode, the only difference is how to define command_executor.

eko999:
@mirzal ,
I have created a robottelo.properties to test this PR, the test result as below

eko999:
@mirzal , I have created a robottelo.properties to test this PR, the test result as below:
`
# pytest tests/foreman/ui/test_architecture.py 
============================================================================ test session starts ============================================================================
platform linux -- Python 3.5.3, pytest-4.0.2, py-1.8.0, pluggy-0.12.0
rootdir: /root/eko/robottelo, inifile:
plugins: mock-1.10.4, services-1.3.1, xdist-1.29.0, cov-2.7.1, forked-1.0.2
collecting ... 2019-06-25 17:48:16 - conftest - DEBUG - BZ deselect is disabled in settings

collected 1 item                                                                                                                                                            

tests/foreman/ui/test_architecture.py 
.                                                                                                                               [100%]

============================================================================= warnings summary ==============================================================================
tests/foreman/ui/test_architecture.py::test_positive_end_to_end
  /usr/lib/python3.5/site-packages/widgetastic/browser.py:721: DeprecationWarning: use driver.switch_to.alert instead
    return self.selenium.switch_to_alert()
  /usr/lib/python3.5/site-packages/widgetastic/browser.py:721: DeprecationWarning: use driver.switch_to.alert instead
    return self.selenium.switch_to_alert()

================================================================== 1 passed, 2 warnings in 123.10 seconds ===================================================================
`

airgun/browser.py
mirzal: Agreed, this message looks good. I think people should be able to understand where the problem is after seeing it.
Thanks for checking this.

airgun/browser.py
mirzal: I mean this comment specifically:

> todo: attempt to rename job here

SauceLabs has test runner and test reporter capabilities, so this comment kind of makes sense to me in that context - we would like to change "job name", to have something better in SauceLabs (but why and to what exactly escapes me).

But in context of remote webdriver, I don't understand what would be a "job", why we would like to rename it and what name we would like to use instead.
Perhaps this is just me missing something about remote webdriver, but if you could explain it, I would very much appreciate it.

abalakh: @mirzal saucelabs doesn't know test name, so when session starts, it just gives it unique id. After test finishes, we send a request to update session name to test name and result to either "pass" or "fail". This isn't great by 2 reasons:
1) while tests are running (during test run or when you're debugging smth) its hard to find proper session as its one of those 8+ "d7f9sd97fsd9fsd7fs9d7f" running in parralel
2) if, for some reason, connection with saucelabs breaks, its pretty much impossible to find out what test caused issues using saucelabs interface only (of course, possible via jenkins logs) due to that id instead of name.

So i left this comment to return back in future and attempt to rename job immideately after session start, as it could be a nice improvement (but nothing really critical, just handy in some rare cases).

And this comment is irrelevant for remote webdriver for sure.

eko999: Okay, I copied the code from sauce because the same structure in my understanding, I will remove it for remote provider, thank you all.

eko999:
@rplevka @san7ket @ntkathole , could you help to review this PR? because the Remote provider has the same desired_capabilities requirement as Sauce, I moved the desired_capabilities code from `  _get_saucelabs_browser` and create a new function `_get_webdriver_capabilities which can be used by remote and sauce, there is no influence for the `_get_saucelabs_browser`. 

@mirzal is waiting for your review and then can merge this pr,  thank you guys so much.

san7ket:
@eko999 we will also need a PR in jenkins-config repo, to update the robottelo.properties. As , 
if there is a setting missing , settings.configure() would return 1 and break building of robottelo. I can PM you the details. 

Please DNM this PR , until the config PR is merged. After that we can test and I can review

sghai:
@san7ket @ntkathole do you think changes are good for 2nd ACK ? or are you guys waiting on some comments to be worked on ?

san7ket:
Test results, just one module run would be enough

mirzal:
Jenkins job run is at $JENKINS/view/Satellite%206/job/satellite6-standalone-automation/1594 . It failed, but due to #249 . Tunnel to sauce was correctly created, sauce has full logs and video, test did run for 10 minutes before failing. Clearly problem is not related to this PR, just bad luck when choosing file to test.

@eko999 thank you very much for this work and your patience. Merging now.

PR #365, opened 2019-07-16
Updated the locator for create image

vijay8451:
- Updated the locator for create image
- Test result:
```
tests/foreman/ui/test_computeresource.py::test_positive_image_end_to_end PASSED                                                                                  [ 50%]
tests/foreman/ui/test_computeresource_vmware.py::test_positive_image_end_to_end PASSED                                                                           [100%]

```


vijay8451:
> Too bad this locator changed. id looked like unique and more stable thing to use. Would be good idea to ask devs to add id back, or create PR for that ourselves.
> Of course this one is +1.

As discussed , you are going to send the PR around it ... thanks

vijay8451:
Follow-up opened #369

PR #370, opened 2019-07-21
local web UI tests in headless mode

ntkathole:
Support to run webUI tests in headless mode for chrome

### DIFF
--- a/airgun/browser.py
+++ b/airgun/browser.py
@@ -170,6 +170,7 @@ class SeleniumBrowserFactory(object):
         """
         kwargs = {}
         binary = settings.selenium.webdriver_binary
+        browseroptions = settings.selenium.browseroptions
 
         if self.browser == 'chrome':
             if binary:
@@ -178,6 +179,9 @@ class SeleniumBrowserFactory(object):
             prefs = {'download.prompt_for_download': False}
             options.add_experimental_option("prefs", prefs)
             options.add_argument('disable-web-security')
+            if browseroptions:
+                for opt in browseroptions.split(';'):
+                    options.add_argument(opt)
             kwargs.update({'options': options})
             self._webdriver = webdriver.Chrome(**kwargs)
         elif self.browser == 'firefox':
--- a/airgun/settings.py
+++ b/airgun/settings.py
@@ -44,6 +44,7 @@ class SeleniumSettings(object):
         self.webdriver = None
         self.webdriver_binary = None
         self.command_executor = None
+        self.browseroptions = None
 
 
 class WebdriverCapabilitiesSettings(object):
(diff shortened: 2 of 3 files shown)


airgun/browser.py
```diff
@@ -178,6 +179,8 @@ def _get_selenium_browser(self):
             prefs = {'download.prompt_for_download': False}
             options.add_experimental_option("prefs", prefs)
             options.add_argument('disable-web-security')
+            if headless == "true":
```
mirzal: Just `if headless:`.

First, we will obtain this value from Robottelo, where it is casted to boolean. So it will never be equal string "true".
Second, string that contains value that seems to be of other type is good indicator that we should just change variable type.

ntkathole:
Thanks for review @mirzal I keep it simple as we are not going to use it in production and its common requirement. I agree that, from framework perspective, it should be general configuration settings. We already have `webdriver_desired_capabilities` section https://github.com/SatelliteQE/airgun/blob/master/settings.ini.example#L19 where we can set options as per our need specific to webdrivers. Do you want to have this headless option there ?

mirzal:
No, the place is correct. This should be part of `selenium` section, not `webdriver_desired_capabilities` section. Selenium is moving away from using desired capabilities to set web browser options, so we shouldn't act against them here. Especially since we don't have support for passing arbitrary capabilities anyway.

What I mean is that we shouldn't have `headless` setting at all. We should have `browserOptions` setting, that can be used to pass arbitrary number of options (e.g. `--headless;--incognito`). This way our code will be robust against future changes, and allow for more specific options if they are needed. We can document most often used options in comments, airgun docs, Mojo or anywhere.

I can ACK this change, but on condition that we do not make it part of contract that must be kept for backwards compatibility. By that I mean that we can change to using more generic approach at any time and will not have to support existing `headless` settings. Of course we will change Robottelo and CI configs accordingly, but we can't change existing users settings.

Sounds ok?

ntkathole:
@mirzal Sounds good. Updated, please revisit!

airgun/browser.py
```diff
@@ -178,6 +179,8 @@ def _get_selenium_browser(self):
             prefs = {'download.prompt_for_download': False}
             options.add_experimental_option("prefs", prefs)
             options.add_argument('disable-web-security')
+            if browseroptions:
+                options.add_argument(browseroptions)
```
mirzal: Can you make that:
```python
if browseroptions:
    for opt in browseroptions.split(';'):
        options.add_argument(opt)
```
?

If we name setting "browseroptions", we really should support multiple options ;) .

We will probably need to eventually change this somehow, but using semicolon as delimiter sound like relatively safe idea. I know for sure that some options take argument of comma-delimited list, so we want to use something other than comma.

ntkathole: @mirzal done

Issue #348, opened 2019-06-28
Rename SauceLab job automatically


mirzal:
Issue created based on comment in code: https://github.com/SatelliteQE/airgun/blob/7b37535bc808f599f41f517a47b94d5ba640fda3/airgun/browser.py#L229

See also comment in PR #334 for more details: https://github.com/SatelliteQE/airgun/pull/344#discussion_r297369056

PR #372, opened 2019-07-22
add test_name to the capabilities for `remote` browser

rplevka:
The `remote_browser` did not forward the `self.test_name` as one of the desired_capabilities, leaving the session without a name (causing random uuid to be generated for them and causing chaos in the Dashboard). This fixes it.

### DIFF
--- a/airgun/browser.py
+++ b/airgun/browser.py
@@ -309,6 +309,9 @@ class SeleniumBrowserFactory(object):
         if settings.webdriver_desired_capabilities:
             desired_capabilities.update(
                 vars(settings.webdriver_desired_capabilities))
+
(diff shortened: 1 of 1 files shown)


mirzal:
This change makes it easier to work with Zalenium, correct? Does it have any impact on tests run on SauceLabs? Namely, I wonder if it fixes #348

rplevka:
yeah it turns the random uuids into
img

..not sure about saucelabs tho, as far as i know saucelabs already receive the test name in the capabilities and show them on the dashboard.

rplevka:
ok, so I set up locla saucelabs config and ran the test and it looks like the test name is properly resolved even before the actual session starts:
img

PR #375, opened 2019-07-23
overriding read to handle cv_filter option

omkarkhatavkar:
### PR Objective
CV filter by default is empty making table having empty rows and while reading puppet modules tab it is not been selected. This will read puppet modules based on CV_filter will help to fix the issue for https://github.com/SatelliteQE/robottelo/issues/7217

### Test Result 
```
Testing started at 12:05 PM ...
/home/okhatavk/Satellite/robottelo/env/bin/python /home/okhatavk/Downloads/pycharm-community-2018.1.4/helpers/pycharm/_jb_pytest_runner.py --target test_lifecycleenvironment.py::test_positive_add_puppet_module
Launching py.test with arguments test_lifecycleenvironment.py::test_positive_add_puppet_module in /home/okhatavk/Satellite/robottelo/tests/foreman/ui

2019-07-23 12:05:45 - conftest - DEBUG - Registering custom pytest_configure

2019-07-23 12:05:45 - conftest - DEBUG - Fetching BZs to deselect...

2019-07-23 12:06:03 - conftest - DEBUG - Deselected tests reason: BZ resolution ['1487317', '1156555', '1147100', '1214312', '1278917', '1226425', '1230902', '1414821', '1217635', '1479291', '1310422', '1311113', '1347658', '1204686', '1475443', '1199150']

2019-07-23 12:06:03 - conftest - DEBUG - Deselected tests reason: missing version flag ['1487317', '1156555', '1682940', '1147100', '1214312', '1278917', '1581628', '1625783', '1194476', '1226425', '1230902', '1436209', '1217635', '1489322', '1479291', '1310422', '1311113', '1347658', '1204686', '1716429', '1701132', '1701118', '1378442', '1610309', '1475443', '1199150']

2019-07-23 12:06:03 - conftest - DEBUG - Deselected tests reason: sat-backlog ['1487317', '1156555', '1682940', '1147100', '1214312', '1278917', '1581628', '1625783', '1194476', '1226425', '1230902', '1436209', '1217635', '1489322', '1479291', '1310422', '1311113', '1347658', '1204686', '1716429', '1701132', '1701118', '1378442', '1610309', '1475443', '1199150']

platform linux -- Python 3.4.8, pytest-4.0.2, py-1.7.0, pluggy-0.9.0
rootdir: /home/okhatavk/Satellite/robottelo, inifile:
plugins: xdist-1.26.1, services-1.3.1, mock-1.10.0, forked-1.0.2, cov-2.6.12019-07-23 12:06:03 - conftest - DEBUG - Collected 1 test cases

collected 1 item
=================== 1 passed, 15 warnings in 201.01 seconds ====================

```

### DIFF
--- a/airgun/entities/lifecycleenvironment.py
+++ b/airgun/entities/lifecycleenvironment.py
@@ -67,6 +67,12 @@ class LCEEntity(BaseEntity):
         view.packages.search(package_name, cv=cv_name, repo=repo_name)
         return view.packages.table.read()
 
+    def search_puppet_module(self, entity_name, pupet_module_name, cv_name=None):
+        """Search for specific puppet module inside lifecycle environment"""
+        view = self.navigate_to(self, 'Edit', entity_name=entity_name)
+        view.puppet_modules.search(pupet_module_name, cv=cv_name)
+        return view.puppet_modules.table.read()
+
     def search_module_stream(self, entity_name, module_name, cv_name=None, repo_name=None):
         """Search for specific module stream inside lifecycle environment"""
         view = self.navigate_to(self, 'Edit', entity_name=entity_name)
--- a/airgun/views/lifecycleenvironment.py
+++ b/airgun/views/lifecycleenvironment.py
@@ -177,6 +177,23 @@ class LCEEditView(BaseLoggedInView):
             self.searchbox.search(query)
             return self.table.read()
 
+        def read(self):
+            """Returns puppet modules are available for each content view.
+
+               We get dictionary in next format:
+
+                   {
+                       'CV1': [{'Name': 'httpd', 'Version': '0.2' },]
+                       'CV2': [{'Name': 'Nginx', 'Version': '1.2' },]
+                   }
+
+               """
+            result = {}
+            for option in self.cv_filter.all_options:
+                self.cv_filter.fill(option.text)
+                result[option.text] = self.table.read()
+            return result
+
     @View.nested
     class module_streams(SatTab):
         TAB_NAME = 'Module Streams'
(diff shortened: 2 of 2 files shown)


omkarkhatavkar:
> If not, I wonder if you could build this new `read()` on top of `search()`

We can add a search_puppet_module method in entity and can solve this issue. But while reading Lifecycle Environment e.g. `session.lifecycleenvironment.read(lce.name)` we still end with having an empty result for puppet module tab. This commit will fix that and Do you think is that worth it?  

We have a similar case as well here in this view https://github.com/SatelliteQE/airgun/pull/375/files#diff-c1bf7f2e867f6ffc1fb1bae73e858473R68.

mirzal:
When merging this, please use "Merge" and **not** "Squash and merge". This way it will be easy to re-add `read()` method from first approach.

omkarkhatavkar:
> > We can add a search_puppet_module method in entity and can solve this issue. But while reading Lifecycle Environment e.g. `session.lifecycleenvironment.read(lce.name)` we still end with having an empty result for puppet module tab. This commit will fix that and Do you think is that worth it? I think it worth keeping it.
> 
> I can see this point and I agree `read()` not returning data from puppet module tab might be problematic.
> 
> One of main Airgun pain points is how slow it is, which is caused by framework reading too much data and sometimes seemingly doing nothing (probably reading some more data). That's why I would prefer to try more direct approach first, especially as it's clear that we were supposed to do it through entity in the first place. If later it turns out we are limited by this design and many tests could use `read()` returning puppet data, we can get back to your original idea.
> 
> ACK for new method on entity.

I can agree on your direct approach searching the right resource in the right tab. read()  is always going to be slow as we want to read from all tabs if that is the use case. 

We can not keep reason slow to revert for the accurate result. My intention is that if read() method execute fast does not have correct or enough information does not make sense. But if it is slow compared to others by few secs and resulting in accurate or enough information makes more sense.

Issue #380, opened 2019-07-29
Fix oscapcontent .py and oscaptailoringfile.py views


jameerpathan111:
Edit button has been removed from Satellite 6.6.0 for oscapcontent/oscaptailoringfile. Now we have to click on name of oscapcontent/oscaptailoringfile to edit them.

jameerpathan111:
Fixed in #381

PR #381, opened 2019-07-29
fixes oscapcontent and oscaptailoringfile views

jameerpathan111:
fixes issue #380


jameerpathan111:
> Was the button removed, or was "Edit" option removed from dropdown list?
> 
> If button was removed, I think you should also remove `Actions` from https://github.com/SatelliteQE/airgun/blob/master/airgun/views/oscapcontent.py#L28 and https://github.com/SatelliteQE/airgun/blob/master/airgun/views/oscaptailoringfile.py#L28 . Will there be anything in this table except for name?
> 
> (I don't have access to Satellite at the moment)

yes, the edit button was removed but we still have dropdown list for Delete option.

san7ket:
Product wise looks good

PR #386, opened 2019-08-06
Read all the elements in the detail page

eko999:
1. other more options need to read in the detailed page.
2. get the debug checkbox status by the class  `VirtwhoConfiguresDebug`.

### DIFF
--- a/airgun/views/virtwho_configure.py
+++ b/airgun/views/virtwho_configure.py
@@ -1,4 +1,5 @@
 from widgetastic_patternfly import BreadCrumb
+from widgetastic.exceptions import NoSuchElementException
 
 from widgetastic.widget import (
     Checkbox,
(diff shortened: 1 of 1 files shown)


airgun/views/virtwho_configure.py
```diff
@@ -195,14 +221,21 @@ def is_displayed(self):
     @View.nested
     class overview(SatTab):
         status = VirtwhoConfigureStatus('.')
+        debug = VirtwhoConfiguresDebug()
```
vijay8451:
```suggestion
        debug = Text('.//span[contains(@class,'config-debug')]')
```
Just a query , What if you keep the above for debug ? will you not able to see the status in same ?

eko999: Can't get the debug status by Text(), it debug checkbox is checked, the element is:
`<span class="config-debug"><span class="fa fa-check "></span> </span>`
If unchecked, the html is:
`<span class="config-debug"></span>`

This is why I have to use `VirtwhoConfiguresDebug()` to get the status.

PR #388, opened 2019-08-09
GCE Host Provisioning support and updates to existing helpers

jyejare:
New:

- ComputeResourceGoogleProfileStorageItem
- GoogleResourceForm
- Image support in Host Operating System

Updates to:
- Host Entity is updated to wait for JQuery to finish for 20 minutes
- load_zones moved under Zone class in GCE

### DIFF
--- a/airgun/entities/host.py
+++ b/airgun/entities/host.py
@@ -29,7 +29,8 @@ class HostEntity(BaseEntity):
         """Create new host entity"""
         view = self.navigate_to(self, 'New')
         view.fill(values)
-        view.submit.click()
+        self.browser.click(view.submit, ignore_ajax=True)
+        self.browser.plugin.ensure_page_safe(timeout='600s')
         host_view = HostDetailsView(self.browser)
         wait_for(
             lambda: host_view.is_displayed is True,
--- a/airgun/views/computeresource.py
+++ b/airgun/views/computeresource.py
@@ -83,10 +83,10 @@ class ResourceProviderCreateView(BaseLoggedInView):
         google_project_id = TextInput(id='compute_resource_project')
         client_email = TextInput(id='compute_resource_email')
         certificate_path = TextInput(id='compute_resource_key_path')
-        load_zones = Text("//*[contains(@id,'test_connection_button')]")
 
         @View.nested
         class zone(View):
+            load_zones = Text("//a[contains(@id,'test_connection_button')]")
             value = FilteredDropdown(id='s2id_compute_resource_zone')
 
             def before_fill(self, values=None):
--- a/airgun/views/host.py
+++ b/airgun/views/host.py
@@ -110,6 +110,11 @@ class ComputeResourceLibvirtProfileStorageItem(GenericRemovableWidgetItem):
     storage_type = FilteredDropdown(id="format_type")
 
 
+class ComputeResourceGoogleProfileStorageItem(GenericRemovableWidgetItem):
+    """Google Compute Resource profile "Storage" item widget"""
+    size = TextInput(locator=".//input[contains(@id, 'size')]")
+
+
 class HostInterface(View):
     ROOT = ".//div[@id='interfaceModal']"
     title = Text(".//h4[contains(., 'Interface')]")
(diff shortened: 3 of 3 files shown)


airgun/entities/host.py
```diff
@@ -29,7 +29,8 @@ def create(self, values):
         """Create new host entity"""
         view = self.navigate_to(self, 'New')
         view.fill(values)
-        view.submit.click()
+        self.browser.click(view.submit, ignore_ajax=True)
+        self.browser.plugin.ensure_page_safe(timeout='1200s')
```
omkarkhatavkar: Adding such validation common code not valid. In airgun code, we don't have such conditions. This will impact all host create services such as libvirt, azure. Can you handle this all above things in a specific view e.g. if it exists in GCE compute resource?

airgun/views/computeresource.py
```diff
@@ -83,10 +83,10 @@ class GCEProviderForm(View):
         google_project_id = TextInput(id='compute_resource_project')
         client_email = TextInput(id='compute_resource_email')
         certificate_path = TextInput(id='compute_resource_key_path')
-        load_zones = Text("//*[contains(@id,'test_connection_button')]")
 
         @View.nested
         class zone(View):
+            load_zones = Text("//*[contains(@id,'test_connection_button')]")
```
omkarkhatavkar: Please don't use '*' while adding locator, it is not recommended slows the locating strategy.  Please use direct tags or do you have any issue while adding tag name?

airgun/entities/host.py
jyejare: @omkarkhatavkar , This is being followed by oscap content create code - https://github.com/SatelliteQE/airgun/blob/236d791fbd47766801cbdf0b51334a6a65ffbf43/airgun/entities/oscapcontent.py#L23

Also, there is no impact other than increased timeout for page safe.

omkarkhatavkar: `self.browser.click(view.submit, ignore_ajax=True)` I know this added because of special condition, I saw one email thread going for that. Is that a product bug?
`
self.browser.plugin.ensure_page_safe(timeout='1200s')` is there any possibility we can reduce this timeout, this will really slow down the tests. We wanted to reduce the time required to run the test.

vijay8451: @jyejare can you please try something like in below way ? This is because as this timeout `1200s` will introduce the delay for other CR create host too:
 
https://github.com/SatelliteQE/airgun/blob/403e3840a69233bebd2985bd0671cd06a16afe22/airgun/entities/host.py#L35-L40

jyejare: @omkarkhatavkar , @vijay8451 Even If I try this way the timeout will be the same. I think you are more focusing on delay part, for which the ensure_page_safe is already taking care of it.

A timeout of 1200 seconds doesn't mean that the test will wait until 1200 seconds. The periodic delay/check is being taken care of. And as soon as the page is safe rendered the execution comes out of this loop wheter its 100, 200, 500 seconds at that time. Remember timeout is the maximum time to wait and it's not time.sleep(1200)

To make you guys sure, this entire test is taking 1300 seconds. Will post the result soon.

omkarkhatavkar: @jyejare We do understand the difference between `time.sleep(1200)` and `self.browser.plugin.ensure_page_safe(timeout='1200s')`.

A timeout of 1200 seconds doesn't mean that the test will wait until 1200 seconds. 
==> I don't know did you understand my point,  It will wait for 1200 seconds if the ensure_page_safe does not come. If we have multiple tests hitting this code any problem occurs then whole tests will wait at this condition. I will still recommend less time out, 1200 secs are too much.

jyejare: @omkarkhatavkar, 
First - If the ensure_page_safe doesn't happen even after 1200 seconds, they that's a bug (for which we have written this test). Because we wait for such a long time. And if that is happening for all/most compute resources, congrats we have a caught a common issue for all CRs.
Second - The ensure_page_safe time is different for computing resources and while deciding the timeout, we have to think about the compute resource which takes the max time to perform provisioning here its GCE as of now.
Third - I never understand how timeout can be an issue for other compute resources, it can be an issue(increase in time) for automation. But we cannot compromise the functionality for the sake of 'time' its taking to execute.

jyejare: Finally changed the timeout to 600s and with hidden yum update, it works:
https://github.com/SatelliteQE/robottelo/pull/7264#issuecomment-521255113

PR #389, opened 2019-08-11
 Fix LoginView submit button selector

laviro:
After the new Login page was introduced to Foreman,
the button is no longer an 'input' and currently doesn't have the name attribute,

old implementation:
```
<input type="submit" name="commit" value="Log In" class="btn btn-primary" data-disable-with="Log In">
```

new:
```
<button type="submit" disabled="" class="login-pf-submit-button btn btn-lg btn-primary btn-block">Log In</button>
```

I didn't test this change, but if you would like this to work I guess a small modification like this PR could be made.

also, if you would like to add something more specific, I have added this PR in foreman: https://github.com/theforeman/foreman/pull/6971

so we could wait and than change it into something with an ID or name,
this is how the Login Page button will look like after that PR will get merged:
```
<button type="submit" disabled="" id="login_submit_btn" name="commit" class="login-pf-submit-button btn btn-lg btn-primary btn-block">Log In</button>
```

cc @jhutar - fixes https://projects.theforeman.org/issues/27559

### DIFF
--- a/airgun/views/login.py
+++ b/airgun/views/login.py
@@ -4,7 +4,7 @@ from widgetastic.widget import ClickableMixin, Text, TextInput, View
 class LoginView(View, ClickableMixin):
     username = TextInput(id='login_login')
     password = TextInput(id='login_password')
-    submit = Text('//input[@name="commit"]')
+    submit = Text('//button[@type="submit"]')
 
     @property
     def is_displayed(self):


airgun/views/login.py
```diff
@@ -4,7 +4,7 @@
 class LoginView(View, ClickableMixin):
     username = TextInput(id='login_login')
     password = TextInput(id='login_password')
-    submit = Text('//input[@name="commit"]')
+    submit = Text('//button[@type="submit"]')
```
vijay8451: So far downstream is not impacted around it as this issue exists on Foreman v1.23/24 ... Hence please add the label 'Do Not Merge' on it until comes under downstream part.

laviro: No probs.. lets just keep it open until it becomes relevant

vijay8451: Please add label 'DO NOT MERGE' on this PR so that it will not merge until you requires .

omkarkhatavkar: I have added Do Not Merge label, once we see that test requires. we can merge this.

omkarkhatavkar:
@laviro lets wait for downstream, I can merge this soon.

mirzal:
Is there bugzilla for that? Will it be in Sat 6.6, or after that?
I'd say we want this merged right before snap that has this change, so we don't fail entire UI for that release.

laviro:
Yes @mirzal, this will be in 6.6,
can we check for the Sat version here and based on that use the right selector?

mirzal:
@laviro we have git branches for Satellite versions - 6.5.z, 6.4.z and master is for 6.6. As long as we don't cherry-pick this, new selector won't be used in other Satellite versions.
The thing is that within 6.6, snap X will have old selector, and snap Y will have new selector. We never differentiate between snaps, as they are not released to customers. So it's up to us to merge it between snap X and snap Y.

laviro:
@mirzal no problems, we can also close it for now,
and whenever you'll face the test failing you'll know what to do :)

The selector should find an element that looks like that right now:
```
<button type="submit" id="login_submit_btn" name="commit" class="login-pf-submit-button btn btn-lg btn-primary btn-block">Log In</button>
```

mirekdlugosz:
@laviro @omkarkhatavkar at the end of the day, it was not included in 6.6, right? Should we expect it in 6.6.z, or will it be only in 6.7? I guess first 6.7 snap will already have it?

omkarkhatavkar:
@laviro @mirekdlugosz I can see this change in the latest satellite 6.7 version. I'm going to merge this

omkarkhatavkar:
Currently, I'm tier1 reviewer due to which I'm not able to merge this PR. Let's wait for @mirekdlugosz to merge this up.

omkarkhatavkar:
I merged this on priority because, if we won't merge this change now then it would affect all UI tests for 6.7

mirekdlugosz:
@omkarkhatavkar good call, especially as I was on PTO yesterday.

You have write permission on repository, just like Nikhil who merged PR. I think you should be able to merge.
Although master has branch protection rule saying that you need 2 ACKs to merge. Probably when you tried there was only one ACK here.

laviro:
Thanks @mirekdlugosz @omkarkhatavkar :medal_military: 
and @ares for bringing this up !

PR #397, opened 2019-09-17
fixes #349 added logout method

### LINKED ISSUE #349: Implement logout functionality
`LoginEntity().logout()` does nothing:

https://github.com/SatelliteQE/airgun/blob/7b37535bc808f599f41f517a47b94d5ba640fda3/airgun/entities/login.py#L13-L15

See also #20

omkarkhatavkar:
### PR Objective 
Fixed https://github.com/SatelliteQE/airgun/issues/349

### Test Result
```
(env) [root@okhatavk robottelo]# pytest tests/foreman/ui/test_repository.py -v -k test_positive_discover_repo_via_new_product 
2019-09-17 08:58:34 - conftest - DEBUG - Registering custom pytest_configure

=============================================== test session starts ===============================================
platform linux -- Python 3.4.8, pytest-4.0.2, py-1.7.0, pluggy-0.9.0 -- /home/okhatavk/Satellite/robottelo/env/bin/python3.4
cachedir: .pytest_cache
rootdir: /home/okhatavk/Satellite/robottelo, inifile:
plugins: xdist-1.26.1, services-1.3.1, mock-1.10.0, forked-1.0.2, cov-2.6.1
collecting ... 2019-09-17 08:58:34 - conftest - DEBUG - BZ deselect is disabled in settings

collected 20 items / 19 deselected                                                                                

tests/foreman/ui/test_repository.py::test_positive_discover_repo_via_new_product PASSED                     [100%]

================================================ warnings summary =================================================
tests/foreman/ui/test_repository.py::test_positive_discover_repo_via_new_product
  /home/okhatavk/Satellite/robottelo/env/lib/python3.4/site-packages/widgetastic/widget/select.py:130: DeprecationWarning: The value of convert_charrefs will become True in 3.5. You are encouraged to set the value explicitly.
    parser = html_parser.HTMLParser()
  /home/okhatavk/Satellite/robottelo/env/lib/python3.4/site-packages/widgetastic/widget/select.py:133: DeprecationWarning: The unescape method is deprecated and will be removed in 3.5, use html.unescape() instead.
    for option
  /home/okhatavk/Satellite/robottelo/env/lib/python3.4/site-packages/widgetastic/widget/select.py:130: DeprecationWarning: The value of convert_charrefs will become True in 3.5. You are encouraged to set the value explicitly.
    parser = html_parser.HTMLParser()
  /home/okhatavk/Satellite/robottelo/env/lib/python3.4/site-packages/widgetastic/widget/select.py:133: DeprecationWarning: The unescape method is deprecated and will be removed in 3.5, use html.unescape() instead.
    for option
  /home/okhatavk/Satellite/robottelo/env/lib/python3.4/site-packages/widgetastic/widget/select.py:130: DeprecationWarning: The value of convert_charrefs will become True in 3.5. You are encouraged to set the value explicitly.
    parser = html_parser.HTMLParser()
  /home/okhatavk/Satellite/robottelo/env/lib/python3.4/site-packages/widgetastic/widget/select.py:133: DeprecationWarning: The unescape method is deprecated and will be removed in 3.5, use html.unescape() instead.
    for option

============================== 1 passed, 19 deselected, 6 warnings in 234.81 seconds ==============================

```

### DIFF
--- a/airgun/entities/login.py
+++ b/airgun/entities/login.py
@@ -1,5 +1,6 @@
 from airgun.entities.base import BaseEntity
 from airgun.navigation import NavigateStep, navigator
+from airgun.views.common import BaseLoggedInView
 from airgun.views.login import LoginView
 
 
@@ -11,7 +12,10 @@ class LoginEntity(BaseEntity):
         view.submit.click()
 
     def logout(self):
-        pass
+        view = BaseLoggedInView(self.browser)
+        view.taxonomies.select_logout()
+        view.flash.assert_no_error()
+        view.flash.dismiss()
 
 
 @navigator.register(LoginEntity)
--- a/airgun/widgets.py
+++ b/airgun/widgets.py
@@ -839,6 +839,8 @@ class ContextSelector(Widget):
     CURRENT_LOC = '//li[@id="location-dropdown"]/a'
     ORG_LOCATOR = '//li[@id="organization-dropdown"]/ul/li/a[contains(.,{})]'
     LOC_LOCATOR = '//li[@id="location-dropdown"]/ul/li/a[contains(.,{})]'
+    ACCOUNT_MENU = Text("//a[@id='account_menu']")
+    LOGOUT = Text("//a[@href='/users/logout']")
 
     def select_org(self, org_name):
         self.logger.info('Selecting Organization %r' % org_name)
@@ -871,6 +873,10 @@ class ContextSelector(Widget):
     @property
     def current_loc(self):
         return self.browser.text(self.CURRENT_LOC)
+
+    def select_logout(self):
+        self.ACCOUNT_MENU.click()
+        self.LOGOUT.click()
 
     def read(self):
         """As reading organization and location is not atomic operation: needs


mirzal:
This is method that will be run for every single UI test, so before merging this in, I would like to allow for discussion about costs and benefits.

1. What is the impact on tests performance? This is only clicking two things in UI, but times 420 test cases that we have for UI...
2. What are the benefits? For over a year, we have not logged out user on test teardown. Was it a root cause of us missing any issue?

We are already after 6.6 beta and getting closer to GA. Since this might potentially impact all test cases, perhaps this is something we should merge only after branching out 6.6, on our way towards 6.7?

omkarkhatavkar:
> This is a method that will be run for every single UI test, so before merging this in, I would like to allow for discussion about costs and benefits.

I don't have an intention to run this every UI test. As per my view, we should have support for logout functionality. This is one of the key functionality every software (I wonder, why we don't have a test case for this). This is something we can not test using API or CLI.     

> What is the impact on tests performance? This is only clicking two things in UI, but times 420 test cases that we have for UI...

If we used it would not take more than 1 sec to get executed. which means 420 secs ~ 7 min. I would it is still ok to have over each test case, I do agree to not to call every time. But I think we should have at least one test case around this. Even currently it was saying logout in Airgun Session, under the hood we are just saying pass 

Another point I want to raise here is that we had this functionality previously https://github.com/SatelliteQE/robottelo/commit/2147680c7675b6a6b46a5169e75864eef50b0277#diff-fb33006804184a4b4fa6fd5ff257b5beL30. (earlier framework)

 > What are the benefits? For over a year, we have not logged out a user on test teardown. Was it a root cause of us missing any issue?

I started working on this, as this was an open issue in Airgun https://github.com/SatelliteQE/airgun/issues/349. I thought it is good to have this functionality. I don't have any immediate issue which can be resolved with this functionality. I have added this on account of adding coverage. There is one scenario where we could require to log in more than 1 user in a single test, currently, this is handled by creating a new session or new browser. e.g. https://github.com/SatelliteQE/robottelo/blob/4c46ce2e55cae5599ce6de222539238052407cbb/tests/foreman/ui/test_ldap_authentication.py#L294       

> We are already after 6.6 beta and getting closer to GA. Since this might potentially impact all test cases, perhaps this is something we should merge only after branching out 6.6, on our way towards 6.7?

I do agree here we can merge this after 6.6 GA. We can use the below condition to avoid impact by multiple tests. As 'self.login.logout()' earlier doing only pass 
``` python 
        try:
            if passed:
                pass
            else:
                self.take_screenshot()
```

mirekdlugosz:
@omkarkhatavkar I think we can remove "DO NOT MERGE" label and merge it in. What do you think?

Will you be able to follow up on performance change after we get first snap with this change? I'm thinking about two metrics:
1. Change in entire UI automation time
2. Change in average time per test case (paired sample t-test)
We should compare first snap with this change with last snap without this change; assuming we merge soon: first 6.7 snap with last 6.6 snap. Of course we will ignore failures in 6.7 for now, as we may expect some tests to fail due to reasons unrelated to this change.

PR #398, opened 2019-09-18
Added support for RH Recommended Repo

vijay8451:
output:
```
(Pdb) dir(session.redhatrepository)
['HELPER_CLASS', '__class__', '__delattr__', '__dict__', '__dir__', '__doc__', '__eq__', '__format__', '__ge__', '__getattribute__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__le__', '__lt__', '__module__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__str__', '__subclasshook__', '__weakref__', '_helper', 'browser', 'create_bookmark', 'disable', 'enable', 'helper', 'navigate_to', 'read', 'recommended_repos', 'search', 'session']
(Pdb) 

(Pdb) rrepos = session.redhatrepository.recommended_repos()
NAVIGATE: Checking if already at All
NAVIGATE: I'm not at All
NAVIGATE: Heading to destination All
(Pdb) rrepos
[{'name': 'Red Hat Ansible Engine 2.8 RPMs for Red Hat Enterprise Linux 7 Server', 'label': 'rhel-7-server-ansible-2.8-rpms'}, {'name': 'Red Hat Enterprise Linux 5 Server - Extended Life Cycle Support (RPMs)', 'label': 'rhel-5-server-els-rpms'}, {'name': 'Red Hat Enterprise Linux 6 Server (RPMs)', 'label': 'rhel-6-server-rpms'}, {'name': 'Red Hat Enterprise Linux 7 Server - Extended Update Support (RPMs)', 'label': 'rhel-7-server-eus-rpms'}, {'name': 'Red Hat Enterprise Linux 7 Server - Extras (RPMs)', 'label': 'rhel-7-server-extras-rpms'}, {'name': 'Red Hat Enterprise Linux 7 Server - Optional (RPMs)', 'label': 'rhel-7-server-optional-rpms'}, {'name': 'Red Hat Enterprise Linux 7 Server (RPMs)', 'label': 'rhel-7-server-rpms'}, {'name': 'Red Hat Enterprise Linux 8 for x86_64 - AppStream (RPMs)', 'label': 'rhel-8-for-x86_64-appstream-rpms'}, {'name': 'Red Hat Enterprise Linux 8 for x86_64 - BaseOS (RPMs)', 'label': 'rhel-8-for-x86_64-baseos-rpms'}, {'name': 'Red Hat Satellite Capsule 6.6 (for RHEL 7 Server) (RPMs)', 'label': 'rhel-7-server-satellite-capsule-6.6-rpms'}, {'name': 'Red Hat Satellite Maintenance 6 (for RHEL 7 Server) (RPMs)', 'label': 'rhel-7-server-satellite-maintenance-6-rpms'}, {'name': 'Red Hat Satellite Tools 6.6 (for RHEL 5 Server - ELS) (RPMs)', 'label': 'rhel-5-server-els-satellite-tools-6.6-rpms'}, {'name': 'Red Hat Satellite Tools 6.6 (for RHEL 6 Server) (RPMs)', 'label': 'rhel-6-server-satellite-tools-6.6-rpms'}, {'name': 'Red Hat Satellite Tools 6.6 (for RHEL 7 Server) (RPMs)', 'label': 'rhel-7-server-satellite-tools-6.6-rpms'}, {'name': 'Red Hat Satellite Tools 6.6 for RHEL 8 x86_64 (RPMs)', 'label': 'satellite-tools-6.6-for-rhel-8-x86_64-rpms'}, {'name': 'Red Hat Software Collections RPMs for Red Hat Enterprise Linux 7 Server', 'label': 'rhel-server-rhscl-7-rpms'}]
(Pdb) 
```
- Will write a test around same ... but may take sometime.

### DIFF
--- a/airgun/entities/redhat_repository.py
+++ b/airgun/entities/redhat_repository.py
@@ -29,6 +29,14 @@ class RedHatRepositoryEntity(BaseEntity):
             entity_data['items'] = view.available.items(name=entity_name)[0].read_items()
         return entity_data
 
+    def recommended_repos(self):
+        """Click on 'Recommended Repositories'
+        and read RH Repositories values.
+        """
+        view = self.navigate_to(self, 'All')
+        view.recommended_repos.click()
+        return view.available.read()
+
     def enable(self, entity_name, arch, version=None):
         """Enable a redhat repository.
 
--- a/airgun/entities/redhat_repository.py
+++ b/airgun/entities/redhat_repository.py
@@ -17,25 +17,26 @@ class RedHatRepositoryEntity(BaseEntity):
         view = self.navigate_to(self, 'All')
         return view.search(value, category=category, types=types)
 
-    def read(self, entity_name, category='Available', widget_names=None):
+    def read(self, entity_name=None, category='Available', recommended_repo=None):
         """Read RH Repositories values.
 
         :param entity_name: The repository name
         :param category: The repository category to search, options: Available, Enabled
+        :param recommended_repo: on/off RH recommended repositories
         """
         view = self.navigate_to(self, 'All')
+
+        if recommended_repo:
+            current_value = self.browser.get_attribute('class',
+                                                       locator=view.recommended_repos.locator)
+            if recommended_repo not in current_value:
+                view.recommended_repos.click()
+            return view.available.read()
+
         entity_data = view.search('name = "{0}"'.format(entity_name), category=category)[0]
         if category == 'Available':
             entity_data['items'] = view.available.items(name=entity_name)[0].read_items()
         return entity_data
-
-    def recommended_repos(self):
-        """Click on 'Recommended Repositories'
-        and read RH Repositories values.
-        """
-        view = self.navigate_to(self, 'All')
-        view.recommended_repos.click()
-        return view.available.read()
 
     def enable(self, entity_name, arch, version=None):
         """Enable a redhat repository.
--- a/airgun/views/redhat_repository.py
+++ b/airgun/views/redhat_repository.py
@@ -211,6 +211,7 @@ class RedHatRepositoriesView(BaseLoggedInView):
     search_box = TextInput(id='downshift-0-input')
     search_button = Text(".//button[text() = 'Search']")
     search_types = RepositorySearchTypes(".//div[button[@data-id='formControlsSelectMultiple']]")
+    recommended_repos = Text(".//span[@class='bootstrap-switch-label']")
 
     @View.nested
     class available(RepositoryCategoryView):
--- a/airgun/views/redhat_repository.py
+++ b/airgun/views/redhat_repository.py
@@ -211,7 +211,7 @@ class RedHatRepositoriesView(BaseLoggedInView):
     search_box = TextInput(id='downshift-0-input')
     search_button = Text(".//button[text() = 'Search']")
     search_types = RepositorySearchTypes(".//div[button[@data-id='formControlsSelectMultiple']]")
-    recommended_repos = Text(".//span[@class='bootstrap-switch-label']")
+    recommended_repos = Text(".//div[contains(@class, 'bootstrap-switch wrapper')]")
 
     @View.nested
     class available(RepositoryCategoryView):


airgun/entities/redhat_repository.py
```diff
@@ -29,6 +29,14 @@ def read(self, entity_name, category='Available', widget_names=None):
             entity_data['items'] = view.available.items(name=entity_name)[0].read_items()
         return entity_data
 
+    def recommended_repos(self):
+        """Click on 'Recommended Repositories'
+        and read RH Repositories values.
+        """
+        view = self.navigate_to(self, 'All')
+        view.recommended_repos.click()
+        return view.available.read()
```
omkarkhatavkar: image

After successful iteration of the first attempt of reading recommended repos, Consider case here that, if someone wants to read this second time. On the second time, it will click on view.recommended_repos.click() unknowingly recognising it is in recommend the view. 

you can '$x("//div[contains(@class, 'bootstrap-switch wrapper')]")' locator get the current status e.g.

```suggestion
        view.recommended_repos.click()
        return view.available.read()
```
     if 'off' in view.recommended_repos.get_attribute("class")
          view.recommended_repos.click()
     return view.available.read()
.     

```

File: airgun/entities/redhat_repository.py [hunk 2]
```diff
@@ -29,6 +29,14 @@ def read(self, entity_name, category='Available', widget_names=None):
             entity_data['items'] = view.available.items(name=entity_name)[0].read_items()
         return entity_data
 
+    def recommended_repos(self):
```
omkarkhatavkar: Can we change the name to read_recommended_repos(), As all other methods start with action? what do you think?

airgun/entities/redhat_repository.py
mirzal: Also, what if I try to call `read()` after calling this new method? If repo that I try to read is not recommended, method will fail to provide information, as knob is turned on.

I think that this new method should be more generic. Instead of reading recommended repos, it should focus on reading repos list. Additional parameters would control if it should read Available repos, Enabled repos, or both. In case of Available, another parameter should control if "Recommended repos" should be turned on or off. As far as I can tell, `RedHatRepositoriesView` already has infrastructure for reading all items in the view (through `RepositoryCategoryView.items()`), but we always filter that list in some way.

Moreover, `read()` method should be extended to check state of "Recommended repos" and turn it off in case it is turned on (or just set it to "off" and let widget figure out if it has to do anything).

vijay8451: Thank you @omkarkhatavkar and @mirzal ...
I have refactor it by using your ideas and some mine too :) and now it will as below : 

- It would be part of `.read()` now.
- By-default behaviour of `.read()` will remain same.
- `recommended_repo` will work if specified specifically i.e.
```
(Pdb) on_rrepos = session.redhatrepository.read(recommended_repo='on')
(Pdb) off_rrepos = session.redhatrepository.read(recommended_repo='off')
```
- As `Recommended  Repositories` only works for `available` repositories hence worth to retrun the `list of available repos` instead of `enabled` or something else.

omkarkhatavkar: @vijay8451 new logic looks okay to me.

airgun/entities/redhat_repository.py
```diff
@@ -17,13 +17,22 @@ def search(self, value, category='Available', types=None):
...
+        :param recommended_repo: on/off RH recommended repositories
         """
         view = self.navigate_to(self, 'All')
+
+        if recommended_repo:
+            current_value = self.browser.get_attribute('class',
+                                                       locator=view.recommended_repos.locator)
+            if recommended_repo not in current_value:
+                view.recommended_repos.click()
+            return view.available.read()
```
omkarkhatavkar: @vijay8451 Sorry, I have one more question. 

Do we want to make sure that the switcher should be always in OFF state? Just to ensure that if someone calls read method with the only entity_name it should give the correct result. (there could be chances where recommended_repo filter got applied in the previous test case)  
e.g. Scenarios1 :

- Test1 -> uses ->.read(recommended_repo='on')  => Result in recommended switch On
- Test2->uses ->.read(entity_name='$repo_name not exist in recommended repo') => Result not showing that repo.

I would recommend making recommended switcher always to Off State, anyone wants to result from ON state they can get by applying or we read with some logic if the search occurs with entity name it should check in off state as well.

vijay8451: @omkarkhatavkar good question .. the switch always remain in `off` state and whenever a test enable that then 
- It keeps enable until page/browser refresh or session close.
- It will only keep on for that specific test not others.

omkarkhatavkar: cool then everything works in that case.

airgun/entities/redhat_repository.py
```diff
@@ -17,13 +17,22 @@ def search(self, value, category='Available', types=None):
...
         :param entity_name: The repository name
         :param category: The repository category to search, options: Available, Enabled
+        :param recommended_repo: on/off RH recommended repositories
         """
         view = self.navigate_to(self, 'All')
+
+        if recommended_repo:
+            current_value = self.browser.get_attribute('class',
+                                                       locator=view.recommended_repos.locator)
+            if recommended_repo not in current_value:
```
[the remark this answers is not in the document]
vijay8451: I thought to create a separate however found a single use in entire UI , thus used the above way.

vijay8451:
@mirzal 
>> Now read() can return details of single repo, or list of repos, depending on parameters passed. This may be confusing.

`recommended_repo` a special sceanrio and has use case around same then a normal repo read/enable hence does impact much to retrun list in it.

>> Current code may leave "recommended repos" on, causing next call to read() or enable() to fail. Think of test that would do
>> Note that this is distinct to case that Omkar considered in his comment. Omkar was worried about knob state 
across different tests, I'm thinking about knob state in one test.

As per my understanding this would be a rare scenario and if still a hard requirement to use same then might use in below way:

```
recommended_repos = session.redhatrepository.read(recommended_repo='on')
assert "whatever" in recommended_repos
session.brower.refresh()      # To refresh the current page and then `recommended_repo` would be automatice in `off`
whatever = session.redhatrepository.read("whatever")
session.redhatrepository.enable("whatever", "x86_64")
```

omkarkhatavkar:
> Two concerns:
> 1. Now `read()` can return details of the single repo, or list of repos, depending on parameters passed. This may be confusing.
 
I think the first solution was better to read the Recommended Repos in a separate function, I guess @vijay8451 changed this because of this comment https://github.com/SatelliteQE/airgun/pull/398#discussion_r325654934.     

> 2. Current code may leave "recommended repos" on, causing next call to `read()` or `enable()` to fail. Think of a test that would do: 
>    Note that this is distinct to the case that Omkar considered in his comment. Omkar was worried about knob state across different tests, I'm thinking about knob state in one test.
> I understand that practical impact of these concerns is not very big, so I would like to hear @omkarkhatavkar opinion on them - are they enough to block PR, or should we merge as-is and maybe fix these if/when they become the problem?

I think this would be a rare scenario where the user wants to search after recommending a search. because we have very less list of recommend a list of repos (can be seen in a single page for user). As @vijay8451 suggested we can solve this while test case design by applying browser refresh or just making that search as the first call before calling recommended repos. For me, it looks okay (anyof either solution).

PR #399, opened 2019-09-20
add tasks charts support

xdragu01:
Adding view for new tasks dashboard.

### DIFF
--- a/airgun/entities/task.py
+++ b/airgun/entities/task.py
@@ -10,15 +10,31 @@ class TaskEntity(BaseEntity):
         view = self.navigate_to(self, 'All')
         return view.search(value)
 
-    def read_all(self):
-        """Read all tasks widgets values from the title page"""
+    def read_all(self, widget_names=None):
+        """Read all tasks widgets values from the title page.
+        Or read specific widgets by adding 'widget_names' parameter
+        """
         view = self.navigate_to(self, 'All')
-        return view.read()
+        return view.read(widget_names=widget_names)
 
     def read(self, entity_name, widget_names=None):
         """Read specific task values from details page"""
         view = self.navigate_to(self, 'Details', entity_name=entity_name)
         return view.read(widget_names=widget_names)
+
+    def set_chart_filter(self, chart_name, index=None):
+        """Remove filter from searchbox and set filter from specific chart
+
+        :param index: index in 'StoppedChart' table,
+            dict with 'row' number and 'focus' as column name
+        """
+        view = self.navigate_to(self, 'All')
+        chart = getattr(view, chart_name)
+        view.searchbox.clear()
+        if chart_name == 'StoppedChart' and index:
+            chart.table[index['row']][index['focus']].click()
+        else:
+            chart.name.click()
 
 
 @navigator.register(TaskEntity, 'All')
--- a/airgun/views/task.py
+++ b/airgun/views/task.py
@@ -2,7 +2,15 @@ from widgetastic.widget import Text, View
 from widgetastic_patternfly import BreadCrumb
 
 from airgun.views.common import BaseLoggedInView, SearchableViewMixin, SatTab
-from airgun.widgets import ProgressBar, ReadOnlyEntry, SatTable
+from airgun.widgets import (
+    ProgressBar,
+    ReadOnlyEntry,
+    SatTable,
+    Table,
+    ActionsDropdown,
+    PieChart,
+    Pagination
+)
 
 
 class TaskReadOnlyEntry(ReadOnlyEntry):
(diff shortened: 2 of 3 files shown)


airgun/views/task.py
```diff
@@ -20,12 +20,38 @@ class TasksView(BaseLoggedInView, SearchableViewMixin):
...
         }
     )
+    focus = ActionsDropdown("//div[@class='dropdown btn-group']")
 
     @property
     def is_displayed(self):
         return self.browser.wait_for_element(
             self.title, exception=False) is not None
 
+    class RunningChart(View):
```
mirzal: I'm pretty sure you want these decorated with `@View.nested`.

airgun/views/task.py
```diff
@@ -20,12 +20,38 @@ class TasksView(BaseLoggedInView, SearchableViewMixin):
...
+
+    class PausedChart(View):
+        ROOT = ".//div[@id='paused-tasks-card']"
+        name = Text("./h2")
+        total = PieChart("./div[@class='card-pf-body']")
+
+    class StoppedChart(View):
+        ROOT = ".//div[@id='stopped-tasks-card']"
+        name = Text("./h2")
+        table = SatTable(
```
mirzal: Wouldn't `Table` suffice here?

Andrii and Alex marked `SatTable` as deprecated and wanted to remove it completely, but did not finish that task. While I don't find it particularly high priority, let's not make our lives harder and don't use `SatTable` where `Table` is good enough.

airgun/entities/task.py
```diff
@@ -20,6 +20,22 @@ def read(self, entity_name, widget_names=None):
         view = self.navigate_to(self, 'Details', entity_name=entity_name)
         return view.read(widget_names=widget_names)
 
+    def set_chart_filter(self, char_name, index=None):
```
mirzal: Can you rename this parameter to `chart_name`?

`char_name` suggests you mean name of character, which is not the case here.

airgun/entities/task.py
```diff
@@ -20,6 +20,22 @@ def read(self, entity_name, widget_names=None):
...
+        view = self.navigate_to(self, 'All')
+        if char_name not in view.widget_names:
+            raise ValueError('Provide correct chart name')
```
mirzal: getattr for attribute that does not exist will raise exception like:
```
AttributeError: type object 'TasksView' has no attribute 'nonexistingchart'
```
Don't you think it's readable enough? Why do you shadow it with your own exception?

airgun/entities/task.py
mirzal: Both of these new methods are missing docstrings. Please add it, along with description of expected parameters (especially for `index` here, which seems to be dictionary with specific items to navigate the table).

airgun/entities/task.py
```diff
@@ -20,6 +20,22 @@ def read(self, entity_name, widget_names=None):
...
+        if char_name == 'StoppedChart' and index:
+            getattr(view, char_name).table[
+                index['row']][index['focus']].click()
+        else:
+            getattr(view, char_name).name.click()
```
mirzal: Common code in both conditional paths is good sign that it can be taken out of conditional:

```python
chart = getattr(view, chart_name)
if index:
    chart.table[index['row']][index['focus']].click()
else:
    chart.name.click()

File: airgun/entities/task.py [hunk 6]
```diff
@@ -20,6 +22,35 @@ def read(self, entity_name, widget_names=None):
...
+        view = self.navigate_to(self, 'All')
+        chart = getattr(view, chart_name)
+        view.searchbox.clear()
+        if chart_name == 'StoppedChart' and index:
+            chart.table[index['row']][index['focus']].click()
+        else:
+            chart.name.click()
+
+    def read_widget(self, widget_name):
+        """Read specific widget"""
```
mirzal: I feel this goes against the framework, which is built around the idea that tests shouldn't care too much about implementation details behind actions.

Perhaps you could extend `read_all` method to take another argument `widget_names`, and then on test side you would do:
```python
search_term = session.task.read_all('searchbox')
```
? This would be similar to what `read()` does.

airgun/entities/task.py
```diff
@@ -20,6 +22,35 @@ def read(self, entity_name, widget_names=None):
...
+        view = self.navigate_to(self, 'All')
+        return getattr(view, widget_name).read()
+
+    def table_row_count(self):
```
mirzal: There is [`Pagination` widget](https://github.com/SatelliteQE/airgun/blob/master/airgun/widgets.py#L1487). Could you extend it to include `total_items` widget, and return it's content on `read()`? Then on test side you should be able to do:
```python
pagination = session.task.read_all('pagination')
assert pagination['total_items'] == some_oracle
```

airgun/views/task.py
```diff
@@ -14,18 +21,51 @@ class TaskReadOnlyEntry(ReadOnlyEntry):
...
+    focus = ActionsDropdown(
+        "//div[./button[@id='tasks-dashboard-time-period-dropdown']]"
+    )
     table = SatTable(
         ".//div[@id='tasks-table']/table",
         column_widgets={
             'Action': Text('./a'),
         }
     )
+    table_row_count = Text("//span[@class='pagination-pf-items-total']")
```
mirzal: If you extend Pagination widget, this should be:
```python
pagination = Pagination()
```

airgun/entities/task.py
```diff
@@ -20,6 +22,35 @@ def read(self, entity_name, widget_names=None):
         view = self.navigate_to(self, 'Details', entity_name=entity_name)
         return view.read(widget_names=widget_names)
 
+    def set_chart_filter(self, chart_name, index=None):
+        """Remove filter from searchbox ans set filter from specific chart
```
omkarkhatavkar: ans ? typo

airgun/views/task.py
```diff
@@ -14,18 +21,51 @@ class TaskReadOnlyEntry(ReadOnlyEntry):
...
+        total = PieChart("./div[@class='card-pf-body']")
+
+    @View.nested
+    class StoppedChart(View):
+        ROOT = ".//div[@id='stopped-tasks-card']"
+        name = Text("./h2")
+        table = Table(
+            locator='.//table',
+            column_widgets={
+                'Action': Text('./button'),
```
omkarkhatavkar: Is that column name is correct? I see column names as Total and 24h
Tasks

airgun/views/task.py
```diff
@@ -14,18 +21,51 @@ class TaskReadOnlyEntry(ReadOnlyEntry):
 
 class TasksView(BaseLoggedInView, SearchableViewMixin):
     title = Text("//h1[text()='Tasks']")
+    focus = ActionsDropdown(
+        "//div[./button[@id='tasks-dashboard-time-period-dropdown']]"
```
omkarkhatavkar: I think we should have just simplified to `//button[@id='tasks-dashboard-time-period-dropdown']`

mirzal: Oh no, current one means "div that contains button with id", yours means "button with id". They are not the same.
We could use this instead: `//button[@id='tasks-dashboard-time-period-dropdown']/parent::div`. I thought about it and decided it's not improvement, so I did not leave any comment.

xdragu01:
Thanks for your reviews. Test results and PR on robottelo has been updated.

mirekdlugosz:
@omkarkhatavkar can you ACK and merge?

PR #400, opened 2019-09-24
Pass the chromedriver v76+ undefined method 'left' js error

rplevka:
Some of the actions in our testcases fail with `chromedriver v77` due to the javascript error.

according the background of the issue
https://bugs.chromium.org/p/chromedriver/issues/detail?id=3044#c1

I'm just about to silently catch and pass the exception, which is (according the above comment) pretty much what older versions of the chromedriver did.

- tested on several test cases, this didn't break anything.

Please note, that the error occurs under several circumstances, and in our case it wasn't exactly the dropdown.

### DIFF
--- a/airgun/browser.py
+++ b/airgun/browser.py
@@ -670,7 +670,13 @@ class AirgunBrowser(Browser):
         el = self.element(locator, *args, **kwargs)
         self.execute_script(
             "arguments[0].scrollIntoView(false);", el, silent=True)
-        ActionChains(self.selenium).move_to_element(el).perform()
+        try:
+            ActionChains(self.selenium).move_to_element(el).perform()
+        except selenium.common.exceptions.JavascriptException as e:
+            if "javascript error: Cannot read property 'left' of undefined" in e.msg:
+                self.logger.debug("Ignoring 'Cannot read property left of undefined' exception")
+            else:
+                raise(e)
         return el
 
     def get_downloads_list(self):


vijay8451:
@rplevka have following query on it:
- AFAIK we ran last build with v74 ... thus is there any challenge we facing around that ?
- Second , can we let this issue fix and could move further accordingly with next version `v77/78` .. whichever would be more stable  instead `v76` this time?

rplevka:
@vijay8451 
I'd like to update to the latest chromedriver. The PR mentions 76 **+** as that is the version this issue got introduced.
This PR fixes/workarounds one of the issues that airgun has with newer chromedriver.

I'd like to move to the new selenium version (thus newer chromedriver) as the current selenium image coming with v74 contains a bug that causes the grid to get stucked.

airgun/browser.py
```diff
@@ -670,7 +670,13 @@ def move_to_element(self, locator, *args, **kwargs):
         el = self.element(locator, *args, **kwargs)
         self.execute_script(
             "arguments[0].scrollIntoView(false);", el, silent=True)
-        ActionChains(self.selenium).move_to_element(el).perform()
+        try:
+            ActionChains(self.selenium).move_to_element(el).perform()
+        except selenium.common.exceptions.JavascriptException as e:
+            if 'javascript error: Cannot read property \'left\' of undefined' in e.msg:
+                pass
```
mirzal: Can you add following line?
```python
self.logger.debug("Ignoring 'Cannot read property left of undefined' exception")
```

Ignoring for a while that we fail to capture these logs, I believe we want to monitor them, so eventually we might be able to fix the root cause. We should put required pieces already, even if we end up removing the method.

rplevka: sure, that's a good idea

omkarkhatavkar:
> @vijay8451
> I'd like to update to the latest chromedriver. The PR mentions 76 **+** as that is the version this issue got introduced.
> This PR fixes/workarounds one of the issues that airgun has with newer chromedriver.
> 
> I'd like to move to the new selenium version (thus newer chromedriver) as the current selenium image coming with v74 contains a bug that causes the grid to get stucked.

@rplevka Thanks for the quick fix on this. Appreciated :100:

rplevka:
@mirzal - 
> Reading comment to this method, I think we could just remove it altogether.
i reviewed the method i'm modifying and it really looks like it shouldn't be needed anymore. Shall we give it a try and remove it altogether?

mirzal:
@rplevka I'm OK with playing it safe and applying your patch. We can try method removal after 6.6 GA, when we will have a lot of time to deal with potential fallout.

Unless you can run half of UI suite on your local branch and get back to us with results? If there is no visible impact and there are no new failures, we can remove it right away.

airgun/browser.py
```diff
@@ -670,7 +670,14 @@ def move_to_element(self, locator, *args, **kwargs):
         el = self.element(locator, *args, **kwargs)
         self.execute_script(
             "arguments[0].scrollIntoView(false);", el, silent=True)
-        ActionChains(self.selenium).move_to_element(el).perform()
+        try:
+            ActionChains(self.selenium).move_to_element(el).perform()
+        except selenium.common.exceptions.JavascriptException as e:
+            if "javascript error: Cannot read property 'left' of undefined" in e.msg:
+                self.logger.debug("Ignoring 'Cannot read property left of undefined' exception")
+                pass
```
mirekdlugosz: `pass` is "no operation". since you added `logger.debug` call, you don't need it :)

rplevka: ah

rplevka:
> @rplevka let us now if you prefer to run more extensive testing and possibly remove that method, or should we move forward with merging that PR.

I don't have any plans to test the method removal any time soon.

PR #403, opened 2019-09-30
Fixes session navigator and NavigateStep.

ldjebran:
This should fix the basic session navigator copy exception and NavigateStep with newer version of navmazing as newer versions need logger in __init__ to work as expected that is called from navigator 

in earlier version navigator created navigation step without logger, but in newer version it require NavigationStep with logger. 

This should be tested with robottelo, but with navmazing version >=  1.1.6

in insight project created a work-arround for this , and seems working as expected, but robottelo use all the feature of widgetastic and navmazing and need an intensify testing as maybe other areas may be affected.

### DIFF
--- a/airgun/navigation.py
+++ b/airgun/navigation.py
@@ -13,11 +13,11 @@ class NavigateStep(navmazing.NavigateStep):
 
     VIEW = None
 
-    def __init__(self, obj, navigate_obj):
+    def __init__(self, obj, navigate_obj, logger=None):
         """Adding shortcut for navigate object to make easier calls to its
         navigate method
         """
-        super(NavigateStep, self).__init__(obj, navigate_obj)
+        super(NavigateStep, self).__init__(obj, navigate_obj, logger=logger)
         self.navigate_to = self.navigate_obj.navigate
 
     @cached_property
--- a/airgun/session.py
+++ b/airgun/session.py
@@ -1,5 +1,4 @@
 """Session controller which manages UI session"""
-import copy
 import logging
 import os
 import sys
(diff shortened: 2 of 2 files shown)


airgun/session.py
```diff
@@ -182,8 +181,8 @@ def __enter__(self):
             self._factory.post_init()
 
             # Navigator
-            self.navigator = copy.deepcopy(navigator)
-            self.navigator.browser = self.browser
+            self.navigator = Navigate(self.browser)
```
ldjebran: Please note: that we need to create a new Navigator with the session browser for tests that may need many UI sessions and each navigator browser have to bind to the relevant session browser:
example in robottelo many tests use this scenario:
```python
with Session(username=admin_user) as admin_session:
    ...
    with Session(username=non_admin_user) as user_session:
         ...
```

airgun/navigation.py
```diff
@@ -13,11 +13,11 @@ class NavigateStep(navmazing.NavigateStep):
 
     VIEW = None
 
-    def __init__(self, obj, navigate_obj):
+    def __init__(self, obj, navigate_obj, logger=None):
```
ldjebran: @abalakh this will not work with old navmazing versions min 1.1.6

mirekdlugosz:
Adding DO NOT MERGE to keep in sync with https://github.com/SatelliteQE/robottelo/pull/7603 . Next snap is likely to be beta, and proposal is to play it safe and merge these changes after beta.

mirekdlugosz:
Removing DO NOT MERGE, as we already have beta snap. We can resume operating as usual.

This should be merged at around the same time as https://github.com/SatelliteQE/robottelo/pull/7603

PR #404, opened 2019-10-02
airgun bulk action now waits till task is successfully finished

tstrych:
Test `test_positive_bulk_delete_host` does not wait till delete action ends. Delete action was not done but navigation already continues with reading all hosts where still are some (not deleted yet). Therefore assert on empty host table fails.

 [Test reference](https://github.com/SatelliteQE/robottelo/blob/8f45cb9d4f94d0388a651a404bfb528779c7317d/tests/foreman/ui/test_host.py#L1502) 

```
collected 28 items / 27 deselected / 1 selected
=================== 1 passed, 27 deselected in 249.63 seconds ==================
```


omkarkhatavkar:
> But I think maybe we should introduce more generic mechanism already. After all, other actions might wait for the condition to become true, too. I'm thinking of something like:

I agree with @mirekdlugosz, some of the examples are related to you can see here as well. https://github.com/SatelliteQE/airgun/blob/ca6a7f0eec9b7eb4040183273130a779cc5e6fab/airgun/entities/subscription.py#L16

tstrych:
Sorry it takes me a lot of time to get into airgun once again. 

I found out there are only 3 actions that require wait for action, which are different from the others
`Delete Hosts` `Schedule Remote Job` and `Play Ansible roles`. As `Schedule Remote Job` and `Play Ansible roles` were already implemented by jhutar, I made similar change with `Tasks` for `Delete Hosts` action. 

I also verify it with the referenced PR. 
please review once again @mirekdlugosz  @omkarkhatavkar

airgun/views/task.py
```diff
@@ -95,3 +97,15 @@ class task(SatTab):
         progressbar = ProgressBar()
         output = TaskReadOnlyEntry(name='Output')
         errors = TaskReadOnlyEntry(name='Errors')
+
+    def wait_for_result(self, timeout=60, delay=1):
+        """Wait for invocation job to finish"""
+        wait_for(
+            lambda: (self.is_displayed and self.task.progressbar.is_displayed
+                     and self.task.result.read() == 'success'
+                     and ('0 fail' in self.task.output.read())
```
omkarkhatavkar: I don't know do we really need these checks?

airgun/entities/host.py
```diff
@@ -175,6 +171,17 @@ def play_ansible_roles(self, entities_list, timeout=60, wait_for_results=True):
             status_view.wait_for_result(timeout=timeout)
         return status_view.read()
 
+    def delete_hosts(self, entities_list, timeout=60, wait_for_results=True):
+        """Delete all hosts from entities list"""
+        view = self._select_action('Delete Hosts', entities_list)
+        view.submit.click()
+        view.flash.assert_no_error()
+        view.flash.dismiss()
+        status_view = HostsDeleteTaskDetailsView(self.browser)
```
omkarkhatavkar: you could have directly imported TaskDetailsView here? any reason to wrap with HostsDeleteTaskDetailsView.

tstrych: As you are going from host view to task details view, I think it's good connect these views together by `HostsDeleteTaskDetailsView`.  But I do not know the reason and I did it because two other exemples were defined same way. If we want to change it I can do that :) 

Maybe @ldjebran will know as he implemented these classes: `class HostsJobInvocationCreateView(JobInvocationCreateView):`
why it's important to use new HostViewAndAnotherClassView. 

link to host view entity [here](https://github.com/SatelliteQE/airgun/blob/master/airgun/views/host.py#L619)

airgun/views/task.py
tstrych: No we do not need it. :+1:

airgun/entities/host.py
omkarkhatavkar: Okay, I missed looking at those examples. Ok fine

tstrych:
DONE :)

omkarkhatavkar:
@tstrych you need another ack for this.

PR #415, opened 2019-10-28
add function to get the name of virt-who options

hkx303:
Add function to get the name of virt-who options on Overview Page to support coverage for [BZ1649928](https://bugzilla.redhat.com/show_bug.cgi?id=1649928)

### DIFF
--- a/airgun/views/virtwho_configure.py
+++ b/airgun/views/virtwho_configure.py
@@ -219,6 +219,25 @@ class VirtwhoConfigureDetailsView(BaseLoggedInView):
         )
 
     @View.nested
+    class options(View):
+        status = Text("//strong[text()='Status']")
+        hypervisor_type = Text("//strong[text()='Hypervisor Type']")
+        hypervisor_server = Text("//strong[text()='Hypervisor Server']")
+        hypervisor_username = Text("//strong[text()='Hypervisor Username']")
+        interval = Text("//strong[text()='Interval']")
+        satellite_url = Text("//strong[text()='Satellite server FQDN']")
+        hypervisor_id = Text("//strong[text()='Hypervisor ID']")
+        filtering = Text("//strong[text()='Filtering']")
+        filter_hosts = Text("//strong[text()='Filter Hosts']")
+        filter_host_parents = Text("//strong[text()='Filter Host Parents']")
+        exclude_hosts = Text("//strong[text()='Exclude Hosts']")
+        exclude_host_parents = Text("//strong[text()='Exclude Host Parents']")
+        debug = Text("//strong[text()='Enable debugging output?']")
+        proxy = Text("//strong[text()='HTTP Proxy']")
+        no_proxy = Text("//strong[text()='Ignore Proxy']")
+        kubeconfig_path = Text("//strong[text()='Kubeconfig Path']")
+
+    @View.nested
     class overview(SatTab):
         status = VirtwhoConfigureStatus('.')
         debug = VirtwhoConfiguresDebug()
--- a/airgun/views/virtwho_configure.py
+++ b/airgun/views/virtwho_configure.py
@@ -237,20 +237,25 @@ class VirtwhoConfigureDetailsView(BaseLoggedInView):
         no_proxy = Text('.//span[contains(@class,"config-no_proxy")]')
         kubeconfig_path = Text('.//span[contains(@class,"config-kubeconfig_path")]')
 
-        __label_template = "//span[contains(@class, '{class_name}')]/../preceding-sibling::div/strong"
+        __label_template = "//span[contains(@class, '{class_name}')]" \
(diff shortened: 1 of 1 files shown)


hkx303:
Usage:
```
    test_positive_overview_options_name
    ......
    with session:
        session.virtwho_configure.create(form_data)
        results = session.virtwho_configure.read(name)
        fields = {'status': 'Status',
                 'hypervisor_type': 'Hypervisor Type',
                 'hypervisor_server': 'Hypervisor Server',
                 'hypervisor_username': 'Hypervisor Username',
                 'interval': 'Interval',
                 'satellite_url': 'Satellite server FQDN',
                 'hypervisor_id': 'Hypervisor ID',
                 'filtering': 'Filtering',
                 'debug': 'Enable debugging output?'}
        for key, value in fields.items():
            assert results['options'][key] == value
        session.virtwho_configure.delete(name)
      ......
```

Test Result:
```
(venv) [hkx303@kuhuang virtwho]$ pytest test_ui_virtwho.py::test_positive_overview_options_name -s
2019-10-28 10:09:10 - conftest - DEBUG - Registering custom pytest_configure

==================================== test session starts =====================================
platform linux -- Python 3.7.3, pytest-4.0.2, py-1.8.0, pluggy-0.12.0
rootdir: /home/hkx303/Documents/CI/robottelo/robottelo, inifile:
plugins: services-1.3.1, forked-1.0.2, mock-1.10.4, cov-2.7.1, xdist-1.29.0
collecting ... 2019-10-28 10:09:10 - conftest - DEBUG - BZ deselect is disabled in settings

collected 1 item  

=========================== 1 passed, 3 warnings in 80.43 seconds ============================
(venv) [hkx303@kuhuang virtwho]$ 
```

airgun/views/virtwho_configure.py
```diff
@@ -218,6 +218,25 @@ def is_displayed(self):
                 and self.breadcrumb.read() != 'Create Config'
         )
 
+    @View.nested
+    class options(View):
+        status = Text("//strong[text()='Status']")
```
mirekdlugosz: I find this approach troublesome from theoretical point of view.

You want to confirm that labels are what we expect, without "Satellite Virt Who Configure/Config" part. But you select element for comparison based on label value. If value is wrong (bug exists), this approach will not find element at all. When using provided test, you will get confusing exception about KeyError instead of simple to understand string mismatch, along with current value and expected value. Moreover, if label ever change, you will have to modify Airgun instead of test - while when [oracle](https://glossary.istqb.org/en/search/oracle) gets outdated, we expect to update test instead of go hunting in frameworks code.

Following XPath is longer, but will match exact element you are looking for - no matter what label it currently has. Unless element disappeared (which would be another bug), label will be always present in `read()` results:
```python
status_label = Text("//span[contains(@class, 'config-hypervisor_id')]/../preceding-sibling::div/strong")
```
You can store it as template and `.format()` for each label to save yourself some typing:
```python
__label_template = "//span[contains(@class, '{class_name}')]/../preceding-sibling::div/strong"
status_label = Text(__label_template.format(class_name="config-hypervisor_id")
```

hkx303:
> This shouldn't be modeled as new "options" nested view, as there is no "Options" tab on Virt Who config details page. These items should be added to `overview` class. I suggest using `label` in variable names, like:
> 
> ```python
>     @View.nested
>     class overview(SatTab):
>         status = VirtwhoConfigureStatus('.')
>         debug = VirtwhoConfiguresDebug()
>         # skipped some items
>         status_label = Text("//strong[text()='Status']")
>         hypervisor_type_label = Text("//strong[text()='Hypervisor Type']")
> ```
> 
> Personally I would have one block for fields and labels below, but you can place label next to field.
> 
> I also left one more comment, but I don't consider that one blocking.

yeah, I knew we only have two tabs: Overview and Deploy but ignored the nested view model as I wanted to set all labels alone as one dict(not along with existing values), will modify this with next commit, thanks

airgun/views/virtwho_configure.py
hkx303: oh, this method of the relative path is awesome, I knew mine is not good but could not come up with a better one, thanks.

hkx303:
> ACK. Thanks for incorporating all change requests.
> 
> Please note that this is being merged to master, which is work towards 6.7. If you want these changes while testing Satellite 6.6, submit another PR with cherry-pick of this commit to `6.6.z` branch.



> ACK. Thanks for incorporating all change requests.
> 
> Please note that this is being merged to master, which is work towards 6.7. If you want these changes while testing Satellite 6.6, submit another PR with cherry-pick of this commit to `6.6.z` branch.

ok, thanks for the reminder

hkx303:
@omkarkhatavkar @vijay8451 Hi,could you help review this, thanks :)

PR #419, opened 2019-11-07
Remove sauceclient

### LINKED ISSUE #418: sauceclient call to post test result is error prone
https://github.com/SatelliteQE/airgun/blob/f4bc16c370545968db48835cc452829e3d317e47/airgun/browser.py#L321-L341

From my understanding from researching this is only so that the test result in Saucelabs it self will be shown as Pass/Fail. Without it, it would be just shown as "Complete" in saucelabs.

https://wiki.saucelabs.com/display/DOCS/Setting+Test+Status+to+Pass+or+Fail

On my local setup I have had many issues with this (SSL, and some others). Also, if you forget to install `sauceclient` (in `requirements-optional.txt`) you get a error. This code seems to be the only user of the `sauceclient` package

I was going to try to investigate more and fix the issues, but I started wondering what, if any value this feature has in the first place. I have never needed it. Also, I am not sure we want to be in the business of trying to hit a rest API (that may or may not be available) everytime a test ends. If it does fail, pytest marks the test as failed. 

Does anyone mind if I remove this?

pgagne:
Fixes #418

```
(venv) redacted$  PYTHONPATH="../airgun/:$PYTHONPATH" pytest tests/foreman/ui/test_debug.py 
========================================================================================================================================== test session starts ===========================================================================================================================================
platform darwin -- Python 3.6.5, pytest-4.6.3, py-1.8.0, pluggy-0.12.0
rootdir: /redacted/robottelo
plugins: services-1.3.1, xdist-1.30.0, forked-1.0.2, mock-1.10.4, cov-2.8.1
collecting ... 2019-11-07 18:42:47 - conftest - DEBUG - Collected 1 test cases

                                                                                                                                                                                                                                                                                        

tests/foreman/ui/test_debug.py .                                                                                                                                                                                                                                                                   [100%]

======================================================================================================================================= 1 passed in 181.25 seconds =======================================================================================================================================
(venv) ovpn-125-135:robottelo pgagne$ 

```

### DIFF
--- a/airgun/browser.py
+++ b/airgun/browser.py
@@ -23,14 +23,6 @@ except ImportError:
     # Let it fail later if not installed
     docker = None
 
-try:
-    import sauceclient
-except ImportError:
-    # Optional requirement, airgun will report results back to saucelabs if
-    # installed
-    sauceclient = None
-
-
 LOGGER = logging.getLogger(__name__)
 
 
@@ -147,20 +139,14 @@ class SeleniumBrowserFactory(object):
                 self.browser == 'chrome'):
             self._webdriver.maximize_window()
 
-    def finalize(self, passed=True):
-        """Finalize browser - close browser window, report results to saucelabs
-        or close docker container if needed.
+    def finalize(self):
+        """Finalize browser - close browser window or close docker container if needed.
 
-        :param bool passed: Boolean value indicating whether test passed
-            or not. Is only used for ``saucelabs`` provider.
         :return: None
         """
-        if self.provider == 'selenium' or self.provider == 'remote':
+        if self.provider in ['selenium', 'remote', 'saucelabs']:
             self._webdriver.quit()
             return
-        elif self.provider == 'saucelabs':
-            self._webdriver.quit()
-            return self._finalize_saucelabs_browser(passed)
         elif self.provider == 'docker':
             return self._finalize_docker_browser()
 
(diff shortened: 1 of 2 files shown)


pgagne:
@lhellebr Sent. Have you also had issues with airgun calling sauceclient? If so could you please add them to #418

airgun/browser.py
```diff
@@ -147,20 +139,14 @@ def post_init(self):
                 self.browser == 'chrome'):
             self._webdriver.maximize_window()
 
-    def finalize(self, passed=True):
-        """Finalize browser - close browser window, report results to saucelabs
-        or close docker container if needed.
+    def finalize(self):
```
mirekdlugosz: As you change the signature, can you also update docstring in line 77?

airgun/browser.py
```diff
@@ -147,20 +139,14 @@ def post_init(self):
...
         """
-        if self.provider == 'selenium' or self.provider == 'remote':
+        if self.provider in ['selenium', 'remote', 'saucelabs']:
             self._webdriver.quit()
             return
-        elif self.provider == 'saucelabs':
-            self._webdriver.quit()
-            return self._finalize_saucelabs_browser(passed)
         elif self.provider == 'docker':
             return self._finalize_docker_browser()
```
mirekdlugosz: At this point, this whole thing will be cleaner when written:
```python
if self.provider == 'docker':
    self._finalize_docker_browser()
else:
    self._webdriver.quit()
```
or
```python
if self.provider == 'docker':
    return self._finalize_docker_browser()
self._webdriver.quit()
```

ntkathole:
Just putting here for visibility, I also agree with https://github.com/SatelliteQE/airgun/issues/418#issuecomment-552621375 . so please do consider before going ahead.

lpramuk:
Close ?

mshriver:
Having read #418 and the relevant comments thus far, I'm going to close the PR at this time.  We, SatelliteQE are not currently using saucelabs, but anyone using saucelabs with airgun is welcome to submit changes to the logic concerning browser finalization.

Ultimately, I would like to utilize an external library for browser management, like https://github.com/RonnyPfannschmidt/webdriver_kaifuku. Support for saucelabs client webdrivers should be added to webdriver_kaifuku, composed on top of BrowserFactory, like WharfFactory.

PR #422, opened 2019-11-14
Use "All" page of requested entity as starting url for session

mirekdlugosz:
This is Airgun-only re-implementation of idea presented by @omkarkhatavkar in https://github.com/SatelliteQE/robottelo/pull/7442

The main goal here is to use entity "All items" URL as starting point for session. This way list of all entities will be visible right after logging in (instead of displaying dashboard, which we hardly ever need).

This is implemented by deferring execution of browser initialization code until entity is requested. It has two effects:
1. We can let each entity define URL fragment on their own. Web-specific data is kept within Airgun, which is in line of our design principles. As a result, this change is completely transparent to Robottelo - no test must be changed to benefit from this PR.
2. If you happen to use `Session` context manager, but never actually interact with it, browser will not be initialized. This will save time during test development and allows us to put more code inside session context (as there is no performance penalty for that).

Test case:
```bash
$ pytest -s --durations=0 --count 3 -k end_to_end tests/foreman/ui/test_lifecycleenvironment.py
```

Before change:
```
=================================================================================== slowest test durations ====================================================================================
73.52s call     tests/foreman/ui/test_lifecycleenvironment.py::test_positive_end_to_end[1-3]
71.91s call     tests/foreman/ui/test_lifecycleenvironment.py::test_positive_end_to_end[2-3]
70.26s call     tests/foreman/ui/test_lifecycleenvironment.py::test_positive_end_to_end[3-3]
6.75s setup    tests/foreman/ui/test_lifecycleenvironment.py::test_positive_end_to_end[1-3]
0.24s teardown tests/foreman/ui/test_lifecycleenvironment.py::test_positive_end_to_end[3-3]
```

After change:
```
=================================================================================== slowest test durations ====================================================================================
61.88s call     tests/foreman/ui/test_lifecycleenvironment.py::test_positive_end_to_end[3-3]
61.47s call     tests/foreman/ui/test_lifecycleenvironment.py::test_positive_end_to_end[1-3]
60.53s call     tests/foreman/ui/test_lifecycleenvironment.py::test_positive_end_to_end[2-3]
4.36s setup    tests/foreman/ui/test_lifecycleenvironment.py::test_positive_end_to_end[1-3]
0.26s teardown tests/foreman/ui/test_lifecycleenvironment.py::test_positive_end_to_end[3-3]
```

So, with that PR, we can expect individual test execution time to be reduced by **8 to 13 seconds**.

I am also running entire UI Robottelo suite, but this takes forever.

### DIFF
--- a/airgun/session.py
+++ b/airgun/session.py
@@ -167,38 +167,14 @@ class Session(object):
         self._password = password or settings.satellite.password
         self._session_cookie = session_cookie
         self._factory = None
+        self.navigator = None
         self.browser = None
 
     def __enter__(self):
-        """Starts the browser, navigates to satellite, performs post-init
-        browser tweaks, initializes navigator and UI entities, and logs in to
-        satellite.
+        """Just a shim to make it compatible with context manager
+        protocol. The real work is done by _open the first time
+        any entity is requested.
         """
-        if self._session_cookie:
-            LOGGER.info(u'Starting UI session id: %r from a session cookie',
-                        self._session_cookie.cookies.get_dict()['_session_id'])
-        else:
-            LOGGER.info(u'Starting UI session %r for user %r', self.name, self._user)
-        self._factory = SeleniumBrowserFactory(
-            test_name=self.name,
-            session_cookie=self._session_cookie
-        )
-        try:
-            selenium_browser = self._factory.get_browser()
-            self.browser = AirgunBrowser(selenium_browser, self)
-
-            self.browser.url = 'https://' + settings.satellite.hostname
-            self._factory.post_init()
-
-            # Navigator
-            self.navigator = copy.deepcopy(navigator)
-            self.navigator.browser = self.browser
-            if self._session_cookie is None:
-                self.login.login({
-                    'username': self._user, 'password': self._password})
-        except Exception as exception:
-            self.__exit__(*sys.exc_info())
-            raise exception
         return self
 
     def __exit__(self, exc_type, exc_value, traceback):
@@ -220,6 +196,49 @@ class Session(object):
             LOGGER.exception(err)
         finally:
             self._factory.finalize(passed)
+
+    def _open(self, entity):
+        """Initializes requested entity. If this is first time session
+        requests an entity, also initialize and prepare browser.
+        """
+        if self.browser is None:
+            endpoint = getattr(entity, 'endpoint_path', '/')
+            full_url = f"https://{settings.satellite.hostname}{endpoint}"
+            self._prepare_browser(full_url)
+
+        return entity(self.browser)
+
+    def _prepare_browser(self, url):
+        """Starts the browser, navigates to satellite, performs post-init
+        browser tweaks, initializes navigator and UI entities, and logs in to
+        satellite.
+        """
+        if self._session_cookie:
+            LOGGER.info('Starting UI session id: %r from a session cookie',
+                        self._session_cookie.cookies.get_dict()['_session_id'])
+        else:
+            LOGGER.info('Starting UI session %r for user %r', self.name, self._user)
+        self._factory = SeleniumBrowserFactory(
+            test_name=self.name,
+            session_cookie=self._session_cookie
+        )
+        try:
+            selenium_browser = self._factory.get_browser()
+            self.browser = AirgunBrowser(selenium_browser, self)
+
+            self.browser.url = url
+
+            self._factory.post_init()
+
+            # Navigator
+            self.navigator = copy.deepcopy(navigator)
+            self.navigator.browser = self.browser
+            if self._session_cookie is None:
+                self.login.login({
+                    'username': self._user, 'password': self._password})
+        except Exception as exception:
+            self.__exit__(*sys.exc_info())
+            raise exception
 
     def take_screenshot(self):
         """Take screen shot from the current browser window.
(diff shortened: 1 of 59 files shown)


airgun/session.py
```diff
@@ -253,318 +272,318 @@ def take_screenshot(self):
...
     def usergroup(self):
         """Instance of User Group entity."""
-        return UserGroupEntity(self.browser)
+        return self._open(UserGroupEntity)
 
     @cached_property
     def virtwho_configure(self):
         """Instance of Virtwho Configure entity."""
-        return VirtwhoConfigureEntity(self.browser)
+        return self._open(VirtwhoConfigureEntity)
```
rochacbruno: This comment is not a blocker, the PR LGTM. 

I only have one comment/suggestion about how it is possible to make it DRY, all that `@cached_property` are similiar and too much code repetition.

```python
@lru_cache  # lru_cache can replace @cached_property
def get_open_entity(self, name):
    # This dict can actually be automatically built from `airgun.entities` members.
    entities = {
        'virtwho_configure': VirtwhoConfigureEntity,
       'usergroup': UserGroupEntity,
      ...
    }
    return self._open(entities[name])

def __getattr__(self, name):
    """Called when an attribute lookup has not found"""
    try:
        return self.get_open_entity(name)
    except KeyError:
        return super().__getattr__(name)
```

That is just an idea to be tested/evaluated and maybe implemented in another PR  because with this it is possible to replace all the repetition, 248 lines becomes 72 lines. (or less)

Any time a `session.foo` is called, the `foo` is going to be looked up in the entity list before it raises `AttributeError` if not found.

omkarkhatavkar: @rochacbruno looks like to be a good idea not sure here about consequences, unrelated to this PR. can you please create an issue in Airgun with mentioning details.

mirekdlugosz:
With bd1653b019b9f3d74e0c77e2be67d366f7d3bdd6, both of following idioms are supported:
```python
with session:
   # do something - starting URL will be read from requested entity
```
```python
with session(url="url")
   # do something - starting URL will be $HOSTNAME/url, i.e. provided by test and not entity
```
We initially discussed `session('url')` format, but I decided to generalize this to other arguments supported by `__init__`. So you can also use it to login as specific user (`session(user_name, user_pass)`), or pass session cookie obtained in test (`session(session_cookie=cookie))`).

I am very much interested on what you think about this API.

omkarkhatavkar:
@mirekdlugosz overall looks nice, we are adding many ways access session makes more sense to me. I have one doubt here If we use plain `with session:` it will still navigate entity directly instead navigating from the menubar. For this, we can have the first test from each entity to be using plain session.   

 In other cases with `with session('entity/create')` will navigate to any required page.

airgun/session.py
```diff
@@ -161,24 +162,88 @@ def __init__(self, session_name=None, user=None, password=None, session_cookie=N
...
+    def _open(self, entity):
+        """Initializes requested entity. If this is first time session
+        requests an entity, also initialize and prepare browser.
+        """
+        if self.browser is None:
+            if self._url:
+                endpoint = self._url
+            else:
+                endpoint = getattr(entity, 'endpoint_path', '/')
+            full_url = f"https://{settings.satellite.hostname}{endpoint}"
```
omkarkhatavkar: @mirekdlugosz do you think we should add the logger to know which path url being used?

mirekdlugosz:
> I have one doubt here If we use plain `with session:` it will still navigate entity directly instead navigating from the menubar. 

With plain `with session:`, it will be:
```
All page for entity --(use button)-> New entity page
```

I have checked on `test_positive_create_with_cv` from `tests/foreman/ui/test_activationkey.py`:

```
# with session:
$ pytest -v tests/foreman/ui/test_activationkey.py -k 'test_positive_create_with_cv[1]'
==================================================================== 1 passed, 42 deselected, 1 warnings in 97.66 seconds =====================================================================
```

```
# with session(url='/activation_keys/new')
$ pytest -v tests/foreman/ui/test_activationkey.py -k 'test_positive_create_with_cv[1]'
==================================================================== 1 passed, 42 deselected, 1 warnings in 95.05 seconds =====================================================================
```

Of course these are only single runs for each.

So yes, it will reduce execution time a little further if we change robottelo to set initial URL to that of "new entity" page. But that's for another PR, as it must be done in robottelo and not airgun.

omkarkhatavkar:
> So yes, it will reduce execution time a little further if we change robottelo to set initial URL to that of "new entity" page. But that's for another PR, as it must be done in robottelo and not airgun.

Absolutely, thanks for examples

airgun/session.py
mirekdlugosz: Sure, why not. Added in 2c0665beaf409c8d7a02527760f65e9da66643a5

airgun/session.py
```diff
@@ -161,33 +162,99 @@ def __init__(self, session_name=None, user=None, password=None, session_cookie=N
...
             to bypass login
+        :param str optional url: URL path to open when starting session (without protocol
+            and hostname)
         """
         self.name = session_name or gen_string('alphanumeric')
         self._user = user or settings.satellite.username
         self._password = password or settings.satellite.password
         self._session_cookie = session_cookie
         self._factory = None
+        self._url = None
```
rochacbruno: Shouldn't it take the `url` passed as an argument instead of `None`?
```suggestion
        self._url = url
```

mirekdlugosz: Right, of course. Thanks for catching this.
Changed in 140334a39bd280d575a6bf2d212766ecc0d54d96 . Yes, I do expect these to be squashed before merge :)

airgun/session.py
rochacbruno: https://github.com/SatelliteQE/airgun/issues/426

PR #427, opened 2019-11-22
Input textbox fix

latran:
Fixing the small input textbox and large input text box in settings.  Currently, automation does not recognize the large text area textbox and cannot update values in there.  This fix corrects this issue.

### DIFF
--- a/airgun/widgets.py
+++ b/airgun/widgets.py
@@ -2143,7 +2143,7 @@ class Link(Text):
 
 
 class PopOverWidget(Widget):
-    """Popover-content UI widget which contains header, drop_down or input_box and
+    """Popover-content UI widget which contains header, drop_down, input_box, or textarea and
     submit button. This is associated within a table.
 
     Example html representation::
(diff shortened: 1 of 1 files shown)


latran:
Test results can be seen in [#7502](https://github.com/SatelliteQE/robottelo/pull/7502)

airgun/widgets.py
```diff
@@ -2180,8 +2180,10 @@ def fill(self, item):
         if self.header.is_displayed:
             if self.drop_down.is_displayed:
                 self.drop_down.fill(item)
-            else:
+            elif self.input_box.is_displayed:
                 self.input_box.fill(item)
+            else:
+                self.textarea.fill(item)
```
mirekdlugosz: Hmm... Wonder if we shouldn't refactor it into something like:
```python
for widget_name in ('input_box', 'textarea', 'drop_down'):
    widget = getattr(self, widget_name)
    if widget.is_displayed:
        widget.fill(item)
        break
else:
    raise ReadOnlyWidgetError('Setting FooBar is read-only')
```
This should make it easier to add new widget types in the future, in case they appear.

It also correctly handles settings like Authentication -> OAuth consumer key, which are read-only, but which we have treated as simple inputs in the past.

latran: I'll try your suggestion as it's more clean and update accordingly if it works well.

mirekdlugosz: I think I was a little too fast with that "correctly handles read-only settings". Exception should be raised inside else to `if self.header.is_displayed:`, not to `for widget_name in ():`.

airgun/widgets.py
```diff
@@ -2180,8 +2180,10 @@ def fill(self, item):
         if self.header.is_displayed:
             if self.drop_down.is_displayed:
                 self.drop_down.fill(item)
-            else:
+            elif self.input_box.is_displayed:
```
lhellebr: Minor nitpick: Please update docstring to also mention textarea both for class and fill function

latran: Updated both as you suggested.

airgun/widgets.py
latran: @mirekdlugosz, tried your implementation and it works fine.  Hopefully the code is up to standard.

mirekdlugosz: Yes, it's very good. Thank you.

airgun/widgets.py
```diff
@@ -2143,7 +2143,7 @@ def fill(self, value):
 
 
 class PopOverWidget(Widget):
-    """Popover-content UI widget which contains header, drop_down or input_box and
+    """Popover-content UI widget which contains header, drop_down, input_box, textarea and
```
lhellebr: It only contains one of the three and a submit button, right? Not all three and a submit button.
```suggestion
    """Popover-content UI widget which contains header, drop_down, input_box or textarea and
```

latran: @lhellebr, It contains one of the three and a submit button.  I'll update to say "header, drop_down, input_box, or textarea and submit button."  Would that ok?

Issue #425, opened 2019-11-18
Job template edit view is changed in Sat 6.7


peterdragun:
Editor field in template edit view has lot of changes. Looks like change needs to be done [here](https://github.com/SatelliteQE/airgun/blob/master/airgun/views/common.py#L377). And there is also new field 'Description' below editor field.
old:
![image](https://user-images.githubusercontent.com/43444182/69059182-54b59a00-0a15-11ea-892a-047b8c6ecb22.png)

new:
![image](https://user-images.githubusercontent.com/43444182/69059147-42d3f700-0a15-11ea-9347-6a3d12f4beb0.png)

mirekdlugosz:
Code pointers:
* In `airgun/views/common.py` there is `TemplateEditor` class that models editor and associated widgets
* editor itself is `ACEEditor` from `airgun/widgets.py`
* "Description" field should be added as normal field in `airgun/views/job_template.py` (it's not really part of UI associated with editor)

PR #428, opened 2019-11-25
update view for new template editor

peterdragun:
Resolving #425

### DIFF
--- a/airgun/views/common.py
+++ b/airgun/views/common.py
@@ -16,6 +16,7 @@ from airgun.widgets import (
     ACEEditor,
     ContextSelector,
     GenericRemovableWidgetItem,
+    ItemsList,
     LCESelector,
     ProgressBar,
     ReadOnlyEntry,
@@ -25,8 +26,6 @@ from airgun.widgets import (
     SatTable,
     SatVerticalNavigation,
     Search,
-    Select,
-    ToggleRadioGroup,
     ValidationErrors,
 )
 
@@ -385,12 +384,10 @@ class TemplateEditor(View):
         editor = View.nested(TemplateEditor)
 
     """
-    ROOT = ".//div[@class='editor-container']"
-    rendering_options = ToggleRadioGroup(".//div[@class='btn-group']")
-    import_template = Text(".//a[normalize-space(.)='Import']")
-    fullscreen = Text(".//a[normalize-space(.)='Fullscreen']")
-    syntax_type = Select(id='mode')
-    key_binding = Select(id='keybinding')
+    ROOT = ".//div[@id='editor-container']"
+    rendering_options = ItemsList(".//div[contains(@class,'navbar-editor')]/ul")
+    import_template = Button(id='import-btn')
+    fullscreen = Button(id='fullscreen-btn')
     editor = ACEEditor()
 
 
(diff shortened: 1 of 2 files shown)


airgun/views/common.py
```diff
@@ -385,12 +384,10 @@ class TemplateEditor(View):
...
-    ROOT = ".//div[@class='editor-container']"
-    rendering_options = ToggleRadioGroup(".//div[@class='btn-group']")
-    import_template = Text(".//a[normalize-space(.)='Import']")
-    fullscreen = Text(".//a[normalize-space(.)='Fullscreen']")
-    syntax_type = Select(id='mode')
-    key_binding = Select(id='keybinding')
+    ROOT = ".//div[@id='editor-container']"
+    rendering_options = ItemsList(".//div[contains(@class,'navbar-editor')]/ul")
+    import_template = Text(".//button[@id='import-btn']")
+    fullscreen = Text(".//button[@id='fullscreen-btn']")
```
mirekdlugosz: Wonder why we model these buttons using `Text` instead of `Button`. with `Button`, you would be able to do `Button(id='import-btn')`.

airgun/views/common.py
```diff
@@ -385,12 +384,10 @@ class TemplateEditor(View):
         editor = View.nested(TemplateEditor)
 
     """
-    ROOT = ".//div[@class='editor-container']"
-    rendering_options = ToggleRadioGroup(".//div[@class='btn-group']")
-    import_template = Text(".//a[normalize-space(.)='Import']")
-    fullscreen = Text(".//a[normalize-space(.)='Fullscreen']")
-    syntax_type = Select(id='mode')
-    key_binding = Select(id='keybinding')
```
mirekdlugosz: These two are still available, but in popover that appears after clicking cog button.
From what I see, nothing in Robottelo refers to these items. We can add support for them in Arigun sometime later.

peterdragun: This will be a bit more difficult to add. Considering that these two (+ one new - theme) won't be used in automation I think we don't need these options. Spending time on something we don't even use is not so effective.

airgun/views/common.py
peterdragun: Nice catch. I just rewrote the id and didn't noticed that there was `Text` not `Button`, but I tested it and it works :smile: . I will update this.

airgun/views/common.py
```diff
@@ -385,12 +384,10 @@ class TemplateEditor(View):
...
+    ROOT = ".//div[@id='editor-container']"
+    rendering_options = ItemsList(".//div[contains(@class,'navbar-editor')]/ul")
+    import_template = Button(".//button[@id='import-btn']")
+    fullscreen = Button(".//button[@id='fullscreen-btn']")
```
mirekdlugosz: Why `Button(".//button[@id='fullscreen-btn']")` and not `Button(id='fullscreen-btn')`? Second one doesn't work?

peterdragun:
@omkarkhatavkar this change require also changes in robottelo I will add it with test results in PR. I am already working on it. This is caused by renaming tabs in Satellite from 'Input' to 'Editor'.

peterdragun:
I was finally able to create PR in robottelo that implements these changes. There are also results from tests that was affected by this change.

mirekdlugosz:
Merging based on test results from https://github.com/SatelliteQE/robottelo/pull/7536

PR #434, opened 2019-12-19
updated xpath for task details check

pondrejk:
required by test_positive_bulk_delete_host
```
test tests/foreman/ui/test_host.py -k test_positive_bulk_delete_host
=================================================== test session starts ====================================================
                                                                          

tests/foreman/ui/test_host.py .                                                                                      [100%]


================================== 1 passed, 28 deselected, 3 warnings in 118.70 seconds ===================================
```

### DIFF
--- a/airgun/views/task.py
+++ b/airgun/views/task.py
@@ -17,8 +17,8 @@ from airgun.widgets import (
 
 class TaskReadOnlyEntry(ReadOnlyEntry):
     BASE_LOCATOR = (
-        "//span[contains(., '{}') and contains(@class, 'param-name')]"
-        "/following-sibling::span"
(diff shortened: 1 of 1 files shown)


airgun/views/task.py
```diff
@@ -17,8 +17,8 @@
 
 class TaskReadOnlyEntry(ReadOnlyEntry):
     BASE_LOCATOR = (
-        "//span[contains(., '{}') and contains(@class, 'param-name')]"
-        "/following-sibling::span"
+        "//span[contains(., '{}') and contains(@class, 'list-group-item-heading')]//parent::div"
```
mirekdlugosz: Just to let you know, double slash before `parent::div` is not necessary. Single slash would be enough. I don't think it has any practical implications, so not a blocker.

PR #437, opened 2020-01-06
selecting all hosts using checkbox from table header

### LINKED ISSUE #435: add a function to select all hosts by checking the select all
Currently tests that select all created hosts (e.g.  test_positive_bulk_delete_host) select host by selecting them one by one in the table. Checking the select all box in the table will speed things up

pondrejk:
fixes #435

Previously, airgun selected host one by one in the table based on the entity list. This PR adds an option to select all hosts via the table header checkbox

Confirmation run:
```
pytest tests/foreman/ui/test_host.py -k test_positive_bulk_delete_host

collected 29 items / 28 deselected / 1 selected                                                                                  

tests/foreman/ui/test_host.py .                                                                                            [100%]

===================================== 1 passed, 28 deselected, 3 warnings in 106.81 seconds ======================================
```

### DIFF
--- a/airgun/entities/host.py
+++ b/airgun/entities/host.py
@@ -160,6 +160,8 @@ class HostEntity(BaseEntity):
 
     def play_ansible_roles(self, entities_list, timeout=60, wait_for_results=True):
         """Play Ansible Roles on hosts names in entities_list
+           If keyword 'All' is supplied instead of list, all hosts are selected
+           using the checkbox from table header
 
         :param entities_list: The host names to play the ansible roles on.
         :param timeout: The time to wait for the job to finish.
@@ -173,7 +175,10 @@ class HostEntity(BaseEntity):
         return status_view.read()
 
     def delete_hosts(self, entities_list, timeout=60, wait_for_results=True):
-        """Delete all hosts from entities list"""
+        """Delete all hosts from entities list
+           If keyword 'All' is supplied instead of list, all hosts are selected
+           using the checkbox from table header
+        """
         view = self._select_action('Delete Hosts', entities_list)
         view.submit.click()
         view.flash.assert_no_error()
@@ -295,6 +300,9 @@ class HostsSelectAction(NavigateStep):
             raise ValueError('Please provide a valid action name.'
                              ' action_name: "{0}" not found.')
         entities_list = kwargs.get('entities_list')
-        for entity in entities_list:
-            self.parent.table.row(name=entity)[0].widget.fill(True)
+        if entities_list == "All":
+            self.parent.select_all.fill(True)
+        else:
+            for entity in entities_list:
+                self.parent.table.row(name=entity)[0].widget.fill(True)
         self.parent.actions.fill(action_name)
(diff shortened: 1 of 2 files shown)


airgun/views/host.py
```diff
@@ -203,7 +203,7 @@ class HostsView(BaseLoggedInView, SearchableViewMixin):
         './/table',
         column_widgets={
             0: Checkbox(
-                locator=".//input[@class='host_select_boxes']"),
+                locator="//input[@id='check_all']"),
```
mirekdlugosz: I'm not sure about this change. Have you tried running `session.host.delete_hosts(['host1', 'host2'])`? Seems that instead of selecting these specific hosts, it will click `<input id="check_all">` multiple times.

airgun/entities/host.py
```diff
@@ -160,6 +160,8 @@ def schedule_remote_job(self, entities_list, values, timeout=60, wait_for_result
 
     def play_ansible_roles(self, entities_list, timeout=60, wait_for_results=True):
         """Play Ansible Roles on hosts names in entities_list
+           If keyword 'All' is supplied instead of list, or no list is supplied,
```
mirekdlugosz: After review, this comment get outdated - if you don't pass `entities_list`, you will get `TypeError` about missing positional argument. Same in `delete_hosts`.

airgun/entities/host.py
```diff
@@ -295,6 +300,9 @@ def step(self, *args, **kwargs):
             raise ValueError('Please provide a valid action name.'
                              ' action_name: "{0}" not found.')
         entities_list = kwargs.get('entities_list')
-        for entity in entities_list:
-            self.parent.table.row(name=entity)[0].widget.fill(True)
+        if entities_list == "All":
+            self.parent.select_all.click()
```
mirekdlugosz: Nitpick - please follow widgetastic conventions:
```suggestion
            self.parent.select_all.fill(True)
```

airgun/entities/host.py
```diff
@@ -160,6 +160,8 @@ def schedule_remote_job(self, entities_list, values, timeout=60, wait_for_result
 
     def play_ansible_roles(self, entities_list, timeout=60, wait_for_results=True):
         """Play Ansible Roles on hosts names in entities_list
+           If keyword 'All' is supplied instead of list, all hosts are selected
+           using the checkbox from table header
```
omkarkhatavkar: Is this statement still valid? I don't see conditional statement for 'All' in this play_ansible_roles method

airgun/views/host.py
```diff
@@ -199,6 +199,7 @@ class HostsView(BaseLoggedInView, SearchableViewMixin):
     export = Text(
         ".//a[contains(@class, 'btn')][contains(@href, 'hosts.csv')]")
     new = Text("//a[contains(@href, '/hosts/new')]")
+    select_all = Checkbox(locator="//input[@id='check_all']")
```
omkarkhatavkar: I thought we should have a function that works on the table to select all option or select_all. I don't see that option in widgetastic :(

airgun/entities/host.py
mirekdlugosz: Support is in `HostsSelectAction.step()`, which does most of the work for many of these actions. Note `self._select_action('Play Ansible roles', entities_list)` in method body.

omkarkhatavkar: okay make sense

PR #438, opened 2020-01-10
Do not use javascript, download the file directly.

lhellebr:
This fixes issue with xmlhttpresponse not being able to get file:// urls.

### DIFF
--- a/airgun/browser.py
+++ b/airgun/browser.py
@@ -1,7 +1,6 @@
 """Tools to help getting selenium and widgetastic browser instance to run UI
 tests.
 """
-import base64
 import logging
 import os
 import selenium
@@ -723,31 +722,8 @@ class AirgunBrowser(Browser):
         """
         if settings.selenium.webdriver != 'chrome':
             raise NotImplementedError('Currently only chrome is supported')
-        result = self.selenium.execute_async_script("""
-            function _arrayBufferToBase64(buffer) {
-                var binary = '';
-                var bytes = new Uint8Array(buffer);
-                var len = bytes.byteLength;
-                for (var i = 0; i < len; i++) {
-                    binary += String.fromCharCode(bytes[i]);
-                }
-                return window.btoa(binary);
-            }
-            var uri = arguments[0];
-            var callback = arguments[1];
-            var xhr = new XMLHttpRequest();
-            xhr.responseType = 'arraybuffer';
-            xhr.onload = function(){
-                callback(_arrayBufferToBase64(xhr.response))
-            };
-            xhr.onerror = function(){ callback(xhr.status) };
-            xhr.open('GET', uri);
-            xhr.send();
-        """, uri)
-        if isinstance(result, int):
-            raise Exception(
-                'Failed to get file content. Status code {}'.format(result))
-        return base64.b64decode(result)
+        self.url = uri
+        return bytes(self.selenium.find_elements_by_tag_name('pre')[0].text, "utf-8")
 
     def save_downloaded_file(self, file_uri=None, save_path=None):
         """Save local or remote browser's automatically downloaded file to
@@ -792,6 +768,7 @@ class AirgunBrowser(Browser):
             with open(os.path.join(save_path, filename), 'wb') as f:
                 f.write(content)
             file_path = os.path.join(save_path, filename)
+        self.selenium.back()
         self.selenium.back()
         self.plugin.ensure_page_safe()
         return file_path


lhellebr:
@rplevka Please take a look at this.
Anyway, do not merge yet, I want to test this once more with Zalenium.

mirekdlugosz:
Looks like we never call `get_file_content()` directly, but there are couple of references to `save_downloaded_file()`. Did this ever work? I think it did. Any idea when it started failing?

Also, what exact error do you get?

This might be little academic, but proposed approach will work only for text files that are opened by browser itself. Previous approach should have worked for everything, including PDFs or image files.

lhellebr:
Tested with Zalenium with Chromedriver 74 and 76, both works. If we are ok with the limitations, I think this is OK to get ACKed and merged.

mirekdlugosz:
Actually, there seems to be a way to retain old method.

1. Chrome must be opened with flag `--allow-file-access-from-file`.
2. Open URL `file:///` (or any other known `file://`, for that matter)
3. In JavaScript, execute old method of `XMLHttpRequest`ing file from `file:///` protocol (i.e. whatever we obtained from `get_downloads_list()`).

Seems that in the days of old, Chrome had no problem of accessing `file://` from `chrome://` page. This must have changed recently and now `file://` can be accessed only from `file://` page. But for this to work, additional flag must be passed when opening browser.

@lhellebr Can you check if approach above will work on our existing tests?

lhellebr:
Chrome is already being run with --disable-web-security which contains behavior of --allow-file-access-from-file. I also tried --allow-file-access-from-file explicitly. It didn't help. I can't explain why but it didn't.

mirekdlugosz:
You are right. `--allow-file-access-from-files` did work for me locally (opened new browser with that arg and executed JS in console), but on Zalenium it complained that cross-origin request was blocked by CORS, saying my origin is 'null'. That was on both `file:///` and `chrome://downloads/`.

I did some search and found https://stackoverflow.com/a/47164044/3552063 . This looks very hacky, but it seems to work.

Basically, it creates new input of type "file" on page and sets it's value to local path. Value must be set by Selenium (`send_keys()`), as browsers refuse to set file input value from JavaScript (to prevent data leak). Then `FileReader` API is used to read content of that file and send it back to Selenium.

Curiously enough, old version of this exact answer looks very similar to code that we have right now...

As far as I can tell, there are only two tests in robottelo that try to download file, and both of them pass:
```
$ pytest -v tests/foreman/ui/test_reporttemplates.py::test_positive_end_to_end tests/foreman/ui/test_host.py::test_positive_export
tests/foreman/ui/test_reporttemplates.py::test_positive_end_to_end PASSED                                                                                                               [ 50%]
tests/foreman/ui/test_host.py::test_positive_export PASSED                                                                                                                              [100%]
```

I created #439 with these changes. Can you give them a try?
Let's continue discussion here.

lhellebr:
I encountered the very same post during my research. I didn't like it. As you said, it seems very hacky. Its behavior can change anytime. Plus, it still uses Javascript which seems unnecessary to me.
On the other hand, it supports non-text files. But do we ever need to download those? What will our automation do with such a file?
Personally, I prefer my solution (and it's also much more readable) but it's not a strong preference.

mirekdlugosz:
As for support of non-text files - ideally I'd like to keep that option open, but it's true we do not need it now and there is high chance we will not need it in future.

@omkarkhatavkar @vijay8451 @latran Can you read discussion above and share your opinion?

omkarkhatavkar:
@lhellebr I like your idea and implementation which is short and sweet. I do no like injecting javascript into the application for our need. If we did not have the solution in normal selenium API's then it makes sense to use the hacky javascript solution.

@lhellebr Do you have the test result? Looks like you tested with Zalenium 74, 76 make sense for me.

lhellebr:
There's an issue with downloading .csv files, Chrome refuses to open them as text, it rather downloads them again. I think we need this to work for .csv files but I haven't been able to fix it yet.

lhellebr:
Guys, I don't think I can come with any similarly simple solution that also works with .csv - Chrome limitations are too tight. If you don't have any better idea, I think we will need to use Miroslaw's alternative.

omkarkhatavkar:
> Guys, I don't think I can come with any similarly simple solution that also works with .csv - Chrome limitations are too tight. If you don't have any better idea, I think we will need to use Miroslaw's alternative.

In this case, can you please close this PR and we can review @mirekdlugosz PR.

PR #441, opened 2020-01-22
Added is_displayed for confirm_dialog click

vijay8451:
- Multiple tests are false failed due to a slight delay to load dialog window/related element and `self.confirm_dialog.click()` action failed.
- Added a `wait_for` to avoid false fail and check `is_displayed`.
- Tests results:
```
PASSED2020-01-22 13:15:31 - robottelo - DEBUG - Finished Test: test_repository.py/test_positive_resynchronize_rpm_repo
tests/foreman/ui/test_repository.py::test_positive_resynchronize_puppet_repo PASSED                                                                                   [100%]
tests/foreman/ui/test_repository.py::test_positive_end_to_end_custom_module_streams_crud PASSED                                                                       [100%]
```

### DIFF
--- a/airgun/widgets.py
+++ b/airgun/widgets.py
@@ -1085,12 +1085,23 @@ class ConfirmationDialog(Widget):
     discard_dialog = Text(
         ".//button[contains(@ng-click, 'cancel') and @class='close']")
 
+    def _check_is_displayed(self, elem):
+        """ This is to check if dialog is displayed """
+        wait_for(
+            lambda: elem.is_displayed,
+            timeout=10,
+            delay=1,
+            logger=self.logger
+        )
+
     def confirm(self):
         """Clicks on the positive outcome button like 'Remove', 'Ok', 'Yes'"""
+        self._check_is_displayed(self.confirm_dialog)
         self.confirm_dialog.click()
 
     def cancel(self):
         """Clicks on the negative outcome button like 'Cancel' or 'No'"""
+        self._check_is_displayed(self.cancel_dialog)
         self.cancel_dialog.click()
 
     def read(self):


vijay8451:
> ACK
> I would suggest to move new code into private method and call that method in both `confirm()` and `cancel()`. Next time someone tries to cancel dialog and gets the same error that you had this time.

yeah we could add ... I have pushed the same latest commit now.

airgun/widgets.py
```diff
@@ -1085,12 +1085,24 @@ class ConfirmationDialog(Widget):
     discard_dialog = Text(
         ".//button[contains(@ng-click, 'cancel') and @class='close']")
 
+    def _check_is_displayed(self, value=None):
+        """ This is to check if dialog is displayed """
+        action = self.cancel_dialog.is_displayed if value else self.confirm_dialog.is_displayed
```
mirekdlugosz: I believe you can get away with checking visibility of same element in all cases.
If you want this parametrized, I recommend passing in reference to element:
```python
def _check_is_displayed(self, elem):
    wait_for(
        lambda: elem.is_displayed,
        timeout=30,
        delay=1,
        logger=self.logger
    )

def confirm(self):
    self._check_is_displayed(self.confirm_dialog)
```

vijay8451: I think it's good in both way .. let me replace with yours one.

vijay8451: it's there now ..

airgun/widgets.py
```diff
@@ -1085,12 +1085,23 @@ class ConfirmationDialog(Widget):
     discard_dialog = Text(
         ".//button[contains(@ng-click, 'cancel') and @class='close']")
 
+    def _check_is_displayed(self, elem):
+        """ This is to check if dialog is displayed """
+        wait_for(
+            lambda: elem.is_displayed,
+            timeout=30,
```
omkarkhatavkar: please reduce the timeout to 10

PR #443, opened 2020-01-23
Remove move_to_element override

mirekdlugosz:
There's no header menu in Satellite anymore, and this was source of consistent failures on one test.

Test results and more context in https://github.com/SatelliteQE/robottelo/pull/7603

### DIFF
--- a/airgun/browser.py
+++ b/airgun/browser.py
@@ -11,7 +11,6 @@ import urllib
 from datetime import datetime
 from fauxfactory import gen_string
 from selenium import webdriver
-from selenium.webdriver.common.action_chains import ActionChains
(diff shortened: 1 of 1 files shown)


mirekdlugosz:
Adding DO NOT MERGE to keep in sync with https://github.com/SatelliteQE/robottelo/pull/7603 . Next snap is likely to be beta, and proposal is to play it safe and merge these changes after beta.

mirekdlugosz:
Removing DO NOT MERGE, as we already have beta snap. We can resume operating as usual.

This should be merged at around the same time as https://github.com/SatelliteQE/robottelo/pull/7603

PR #447, opened 2020-02-07
Fixes #446 - Fix Oscap Tailoring file view and entity.

### LINKED ISSUE #446: Fix Oscap Tailoring file view and entity.
Test "tests/foreman/ui/test_oscaptailoringfile.py::test_positive_end_to_end" is failing .

jameerpathan111:
Fixes #446

### DIFF
--- a/airgun/views/oscaptailoringfile.py
+++ b/airgun/views/oscaptailoringfile.py
@@ -14,17 +14,17 @@ from airgun.views.common import (
 from airgun.widgets import (
     ActionsDropdown,
     MultiSelect,
-    SatTable,
+    Table,
 )
 
 
 class SCAPTailoringFilesView(BaseLoggedInView, SearchableViewMixin):
     title = Text("//h1[text()='Tailoring Files']")
     new = Text("//a[contains(@href, 'tailoring_files/new')]")
-    table = SatTable(
+    table = Table(
         './/table',
         column_widgets={
-            'Title': Text('./a'),
+            'Name': Text("./a[contains(@href, '/compliance/tailoring_files')]"),
             'Actions': ActionsDropdown("./div[contains(@class, 'btn-group')]"),
         }
     )


airgun/views/oscaptailoringfile.py
```diff
@@ -14,17 +14,19 @@
...
 class SCAPTailoringFilesView(BaseLoggedInView, SearchableViewMixin):
     title = Text("//h1[text()='Tailoring Files']")
     new = Text("//a[contains(@href, 'tailoring_files/new')]")
-    table = SatTable(
+    table = Table(
         './/table',
         column_widgets={
-            'Title': Text('./a'),
+            'Name': Text("./a"),
+            'Filename': Text("./a"),
```
omkarkhatavkar: I'm just curious will this work anytime? because "./a" point to the same locator in a table. Not sure I'm missing anything. you can use examples mentioned here https://github.com/SatelliteQE/airgun/blob/e5ed58aac970094ebd71923cac1451571a9ecced/airgun/views/puppet_environment.py#L35

airgun/entities/oscaptailoringfile.py
```diff
@@ -100,3 +100,21 @@ def step(self, *args, **kwargs):
         entity_name = kwargs.get('entity_name')
         self.parent.search(entity_name)
         self.parent.table.row(name=entity_name)['Name'].widget.click()
+
+
+@navigator.register(OSCAPTailoringFileEntity, 'Delete')
+class DeleteSCAPTailoringFile(NavigateStep):
+    """To get data from SCAPPolicyDetail view
```
omkarkhatavkar: is the delete view is different from edit view? if not just add a method instead of writing a new view e.g. https://github.com/SatelliteQE/airgun/blob/389b9d1286f0de4f9ca40f02b1c9d4da2d457fc1/airgun/entities/hostgroup.py#L38

airgun/views/oscaptailoringfile.py
mirekdlugosz: Widget definition in `column_widgets` is created relative to table cell (value in row under column with the same name as `column_widgets` dict key), so `Text("./a")` for Name is different element than `Text("./a")` for Filename.

Of course it's good idea to make widget little more specific where possible, to avoid misunderstandings like this one.

airgun/views/oscaptailoringfile.py
```diff
@@ -14,17 +14,19 @@
...
+            'Created at': Text("./a"),
```
mirekdlugosz: Changing `Title` to `Name` is good.

The other two are incorrect. Widget definition is created relative to cell element in table body, not in header, and as far as I can see, these cells contain plain text, not links. So these widgets won't match anything.

Also, `column_widgets` only has to contain definitions for cells with widgets that widgetastic can't figure on it's own. You can, and should, skip columns that contain only text, or with which you don't plan to interact in any away.
Case in point: in current master, `view.table.read()` returns following list on my toy Satellite:
```python
[
    {'Name': 'whatever', 'Filename': 'ssg-rhel7-ds-tailoring.xml', 'Created at': 'about 1 hour ago', 'Actions': ['Download']},
    {'Name': 'tKNrMxynYL', 'Filename': 'ssg-rhel7-ds-tailoring.xml', 'Created at': 'less than a minute ago', 'Actions': ['Download']}
]
```
As you can see, `Filename` and `Created at` fields are both present, despite not being explicitly mentioned in table definition on Airgun side.

airgun/views/oscaptailoringfile.py
jameerpathan111: Yeah this works but will update this to avoid confusion. Thanks.

airgun/entities/oscaptailoringfile.py
```diff
@@ -100,3 +100,21 @@ def step(self, *args, **kwargs):
         entity_name = kwargs.get('entity_name')
         self.parent.search(entity_name)
         self.parent.table.row(name=entity_name)['Name'].widget.click()
+
+
+@navigator.register(OSCAPTailoringFileEntity, 'Delete')
+class DeleteSCAPTailoringFile(NavigateStep):
```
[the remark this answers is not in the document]
jameerpathan111: Yeah thanks for noticing this, will update PR accordingly.

airgun/entities/oscaptailoringfile.py
jameerpathan111: Thanks, will update PR.

airgun/views/oscaptailoringfile.py
jameerpathan111: Thanks for explaining, I didn't knew that :)
will update PR accordingly.

jameerpathan111:
@mirekdlugosz 
Test results:
```
$ pytest tests/foreman/ui/test_oscaptailoringfile.py -k test_positive_end_to_end
collected 9 items / 8 deselected / 1 selected

tests/foreman/ui/test_oscaptailoringfile.py 

tests/foreman/ui/test_oscaptailoringfile.py  .                                                                                                                                                 [100%]

======================================================================== 1 passed, 2 warnings in 123.78 seconds =========================================================================
```

PR #449, opened 2020-02-12
Fixes #448 - Add support for Product Select Action

### LINKED ISSUE #448: Add bulk action support for managing Http proxy for products.
On products page we have Select action drop down through which we can perform some bulk action on products, for example 'Manage HTTP Proxy', 'Sync Selected' etc.
Currently we don't have support for performing this bulk action using Airgun.

jameerpathan111:
Fixes #448

### DIFF
--- a/airgun/entities/product.py
+++ b/airgun/entities/product.py
@@ -10,7 +10,6 @@ from airgun.views.product import (
     ProductsTableView,
     ProductSyncPlanView,
     ProductManageHttpProxy,
-    ProductSyncSelected
 )
 
 
@@ -77,28 +76,6 @@ class ProductEntity(BaseEntity):
         view.progressbar.wait_for_result()
         return view.read()
 
-    def _select_action(self, action_name, entities_list, values=None):
-        """Navigate to all entities, select the entities, and returns the view
-        of the selected action name from main entity select action dropdown.
-        """
-        return self.navigate_to(
-            self,
-            'Select Action',
-            action_name=action_name,
-            entities_list=entities_list,
-            values=values
-        )
-
-    def apply_action(self, action_name, entities_list, values=None):
-        """Apply action name for product/products"""
-        view = self._select_action(action_name, entities_list, values=None)
-        if values['http_proxy_policy'] == "Global Default":
-            values['http_proxy_policy'] = view.http_proxy_policy.all_options[0][0]
-        view.fill(values)
-        view.update.click()
-        view.flash.assert_no_error()
-        view.flash.dismiss()
-
     def manage_http_proxy(self, entities_list, values):
         """Manage HTTP Proxy for product/products
 
@@ -107,20 +84,16 @@ class ProductEntity(BaseEntity):
             eg: {'http_proxy_policy': 'No HTTP Proxy'}, {'http_proxy_policy': 'Global Default'},
             {'http_proxy_policy': 'Use specific HTTP Proxy', 'http_proxy': 'proxy_name'}
         """
-        view = self._select_action('Manage HTTP Proxy', entities_list, values=None)
+        view = self.navigate_to(
+            self,
+            'Select Action',
+            action_name='Manage HTTP Proxy',
+            entities_list=entities_list,
+        )
         if values['http_proxy_policy'] == "Global Default":
             values['http_proxy_policy'] = view.http_proxy_policy.all_options[0][0]
         view.fill(values)
         view.update.click()
-        view.flash.assert_no_error()
-        view.flash.dismiss()
-
-    def sync_selected(self, entities_list):
-        """Apply Sync Selected action on products listed in entities_list
-
-        :param entities_list: The product names to perform sync operation.
-        """
-        view = self._select_action('Sync Selected', entities_list)
         view.flash.assert_no_error()
         view.flash.dismiss()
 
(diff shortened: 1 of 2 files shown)


airgun/entities/product.py
```diff
@@ -75,6 +77,53 @@ def synchronize(self, entity_name):
...
+        """
+        view = self._select_action('Manage HTTP Proxy', entities_list, values=None)
+        if values['http_proxy_policy'] == "Global Default":
+            values['http_proxy_policy'] = view.http_proxy_policy.all_options[0][0]
+        view.fill(values)
+        view.update.click()
+        view.flash.assert_no_error()
+        view.flash.dismiss()
+
+    def sync_selected(self, entities_list):
```
mirekdlugosz: Do we need that option option at this moment? We already have `synchronize` method that syncs single product. How often do you plan to sync multiple products at once, and how many products?

airgun/entities/product.py
```diff
@@ -124,3 +173,35 @@ def prerequisite(self, *args, **kwargs):
...
+        self.VIEW = self.ACTIONS_VIEWS.get(action_name)
+        if not self.VIEW:
+            raise ValueError('Please provide a valid action name.'
+                             ' action_name: "{0}" not found.')
+        entities_list = kwargs.get('entities_list')
+        if entities_list == "All":
+            self.parent.select_all.click()
+        else:
+            for entity in entities_list:
+                self.parent.table.row(name=entity)[0].click()
```
mirekdlugosz: You can't do that, because `ProductsTableView` class doesn't have `select_all` element. If we call `session.product.manage_http_proxy('All', ...)`, we will get exception here.

You can take a look at https://github.com/SatelliteQE/airgun/pull/437 for all the details, or just drop support for `'All'` and only leave `for` loop here.

airgun/views/product.py
```diff
@@ -238,3 +238,19 @@ class ProductSyncPlanView(SyncPlanCreateView):
     def is_displayed(self):
         return self.browser.wait_for_element(
             self.title, exception=False) is not None
+
+
+class ProductManageHttpProxy(BaseLoggedInView):
+    """Represents Http Proxy Management page for Products."""
+    title = Text("//h4[text()='Assign Compliance Policy]")
```
mirekdlugosz: That won't match anything, title is "Http Proxy Management". And you have unclosed single quote.

Please take another look at `airgun/views/host.py`. Specific modal views have `title` set, because they all inherit from `HostsActionCommonDialog`, that will provide `is_displayed` that waits until title is displayed (assumes "title is displayed" means "modal is fully loaded").

I think you should add `is_displayed` here, too, or make this inherit from `ProductsActionCommonDialog` that will define `is_displayed`. Either way works for me.

airgun/entities/product.py
```diff
@@ -75,6 +77,53 @@ def synchronize(self, entity_name):
...
+    def _select_action(self, action_name, entities_list, values=None):
+        """Navigate to all entities, select the entities, and returns the view
+        of the selected action name from main entity select action dropdown.
+        """
+        return self.navigate_to(
+            self,
+            'Select Action',
+            action_name=action_name,
+            entities_list=entities_list,
+            values=values
```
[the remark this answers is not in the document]
jameerpathan111: Updated PR.

airgun/entities/product.py
```diff
@@ -75,6 +77,53 @@ def synchronize(self, entity_name):
...
+        )
+
+    def apply_action(self, action_name, entities_list, values=None):
+        """Apply action name for product/products"""
+        view = self._select_action(action_name, entities_list, values=None)
+        if values['http_proxy_policy'] == "Global Default":
+            values['http_proxy_policy'] = view.http_proxy_policy.all_options[0][0]
```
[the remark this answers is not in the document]
jameerpathan111: updated PR.

airgun/entities/product.py
```diff
@@ -75,6 +77,53 @@ def synchronize(self, entity_name):
...
+        view.fill(values)
+        view.update.click()
```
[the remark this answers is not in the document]
jameerpathan111: I have now dropped the `apply_action` method and kept only `manage_http_proxy`.

airgun/entities/product.py
jameerpathan111: Nope, we don't need it right now. Removed `sync_selected` method.

airgun/entities/product.py
jameerpathan111: Dropped support for `'All'`

airgun/views/product.py
jameerpathan111: Added `is_displayed` and fixed `title`.

airgun/views/product.py
```diff
@@ -238,3 +238,20 @@ class ProductSyncPlanView(SyncPlanCreateView):
     def is_displayed(self):
         return self.browser.wait_for_element(
             self.title, exception=False) is not None
+
+
+class ProductManageHttpProxy(BaseLoggedInView):
+    """Represents Http Proxy Management page for Products."""
+    title = Text("//h4[text()='Assign Compliance Policy']")
```
mirekdlugosz: will that work? Shouldn't it be
```suggestion
    title = Text("//h4[text()='Http Proxy Management']")
```

jameerpathan111: yep

jameerpathan111: update PR, thanks.

jameerpathan111:
@omkarkhatavkar I haven't written any tests yet, but I have tested this code and it works as expected.
I will write tests for it soon and will raise PR on robottelo.

PR #450, opened 2020-02-12
add script for autogen apidoc

### LINKED ISSUE #354: Configure Sphinx to generate thorough API documentation automatically
Current [API documentation](https://airgun.readthedocs.io/en/latest/api/) is preliminary at best. It contains only things placed manually in relevant file and is missing classes and methods docstrings.

There is probably some setting we need to tune to let sphinx do the work for us.

peterdragun:
fixes #354 
PR is resolving:

- problem with updating rst files: ~~I have created script similar to [one that robottelo uses](SatelliteQE/robottelo/blob/master/scripts/gen_api_docs.sh). This needs to be run every time you add or remove file. (this solution is not ideal but haven't found better one) You just need to make docs and script will be triggered.~~ After discussion: I am using auto-api generator (as Mirek suggested), which runs every time when docs are built. Now we don't have to store rst files in repo, they are autogenereted and then deleted.

- ~~another issue with doc-strings: can be fixed by installing airgun to readthedocs, this needs to be done in readthedocs.org->airgun->Admin->Advanced Settings by checking "Install Project" using user with admin rights. I found that we are missing selenium in setup.py and this wouldn't work without it.~~ This is resolved by using autoapi

I have tested this on my repo, results are available [here](https://airgun-peter.readthedocs.io/en/latest/autoapi/index.html).


airgun/widgets.py
```diff
@@ -643,6 +643,8 @@ def fill(self, item):
 
 
 class Search(Widget):
+    """
+    """
```
mirekdlugosz: Please don't add empty docstrings just to make tool happy.

It's better to add this widget to list of ignored references in `docs/conf.py`:
```
nitpick_ignore = [
    ('py:class', 'airgun.widgets.Search'),
]
```

Better yet, write a sentence or two on what this widget is doing :) .

peterdragun:
Thanks for your review.
1. Agree with that, I will update it and also resolve your comment in code.
2. I have tried this before. When you create api docs with this generator the structure of docs is a bit different to that we are using now. Personally I don't like it. It creates 'Submodules' and 'Subpackages' directories, see [auto-api generated docs](https://airgun-peter.readthedocs.io/en/docs-test1/) (I changed just rst files to auto-gen so there are some issues) and compare it to my [docs](https://airgun-peter.readthedocs.io/en/latest/api/index.html). Maybe this could be adjusted, but I haven't found out how. It was easier and faster to just create my own script. This script is needed to run just when you create(rename) file or change structure of directories. I think this is not happening often so I don't think this will be big problem. But if you prefer this method I can update this PR to use auto-api.

mirekdlugosz:
> 2. I have tried this before. When you create api docs with this generator the structure of docs is a bit different to that we are using now. Personally I don't like it. It creates 'Submodules' and 'Subpackages' directories, which is shown in navigation menu, see [auto-api generated docs](https://airgun-peter.readthedocs.io/en/docs-test1/autoapi/index.html) (I changed just rst files to auto-gen so there are some issues) and compare it to my [docs](https://airgun-peter.readthedocs.io/en/latest/api/index.html).

I see what you mean. From where I stand, that's a small price to pay for fully automated docs build (no need to run some command in specific situations) and avoiding committing autogenerated code in git. Index pages are almost indistinguishable and both show all files at once, which is good enough for me.

airgun/entities/puppet_environment.py
```diff
@@ -3,7 +3,7 @@
 from airgun.entities.base import BaseEntity
 from airgun.navigation import NavigateStep
 from airgun.navigation import navigator
-from airgun.views.puppet_environment import ImportPuppetEnvironmentView
+from airgun.views.puppet_environment import ChangedPuppetEnvironmentView
```
mirekdlugosz: This name is hard to understand outside of context of this PR. Could you rename that to `PuppetEnvironmentImportView`?

So:
* `airgun/entities/puppet_environment.py` will define `ImportPuppetEnvironmentView` with `navigator.register` decorator
* `airgun/views/puppet_environment.py` will define `PuppetEnvironmentImportView`

This will be consistent with other class names around.

docs/index.rst
```diff
@@ -51,7 +51,7 @@ web UI) to these defined by Widgetastic (like generic `Text`). Widgets allows
 you to interact with various UI elements and are used in views.
 
 And finally tests. Tests are not part of AirGun repository. You can contribute
-tests to the Robottelo, ``tests/foreman/`` `ui_airgun`_ folder (see "Quick
+tests to the Robottelo, ``tests/foreman/`` `ui`_ folder (see "Quick
```
mirekdlugosz: Good catch! Thanks for fixing these as well

lhellebr:
I don't see any obvious issues. How can I test this PR?

mirekdlugosz:
@lhellebr 
```bash
$ git fetch upstream pull/450/head  # `upstream` here is name of remote pointing to this repo, not your fork
$ git checkout FETCH_HEAD
$ make docs-html
```
Then explore content of `docs/_build/html/`, or:
```bash
$ python -m http.server -d docs/_build/html/
```
And open <http://localhost:8000/> in your browser. Theme will be wrong (compared to readthedocs), but content is there.

peterdragun:
Or if you want just see docs - https://airgun-peter.readthedocs.io/en/latest/autoapi/index.html
These docs are built from same branch as this PR, also there is correct theme.

PR #454, opened 2020-02-18
Add pre commit

mshriver:
Same idea as 

https://github.com/SatelliteQE/nailgun/pull/700

https://github.com/SatelliteQE/robottelo/pull/7641


.travis.yml
```diff
@@ -1,11 +1,14 @@
 language: python
 python:
-  - "3.6"
+  - 3.6
```
jawatts: Plans to move this up to 3.9 soon?

mirekdlugosz: No. Latest discussion around Python version that we had was in https://github.com/SatelliteQE/robottelo/issues/7636

We could add support for newer versions along 3.6, but given that we can assume them to be backwards-compatible and our automation won't use them anyway, it's not clear what benefit that would bring.

jawatts: :+1: Thanks @mirekdlugosz

PR #459, opened 2020-02-24
added suppot for rhsso/keycloak login

omkarkhatavkar:
added support for PR SatelliteQE/robottelo#7657

### DIFF
--- a/airgun/session.py
+++ b/airgun/session.py
@@ -150,7 +150,7 @@ class Session(object):
     """
 
     def __init__(self, session_name=None, user=None, password=None,
-                 session_cookie=None, url=None):
+                 session_cookie=None, url=None, login=None):
         """Stores provided values, doesn't perform any actions.
 
         :param str optional session_name: string representing session name.
@@ -175,8 +175,9 @@ class Session(object):
         self._url = url
         self.navigator = None
         self.browser = None
+        self._login = True
 
-    def __call__(self, user=None, password=None, session_cookie=None, url=None):
+    def __call__(self, user=None, password=None, session_cookie=None, url=None, login=None):
         """Stores provided values. This allows tests to provide additional
         value when Session object is returned from fixture and used as
         context manager. Arguments are the same as when initializing
@@ -190,6 +191,8 @@ class Session(object):
             self._session_cookie = session_cookie
         if url is not None:
             self._url = url
+        if login is not None:
+            self._login = login
         return self
 
     def __enter__(self):
@@ -232,14 +235,11 @@ class Session(object):
             else:
                 endpoint = getattr(entity, 'endpoint_path', '/')
             full_url = f"https://{settings.satellite.hostname}{endpoint}"
-            if entity.__name__ == "RHSSOLoginEntity":
-                self._prepare_browser(full_url, login_type="RHSSO")
-            else:
-                self._prepare_browser(full_url)
+            self._prepare_browser(full_url)
 
         return entity(self.browser)
 
-    def _prepare_browser(self, url, login_type=None):
+    def _prepare_browser(self, url):
         """Starts the browser, navigates to satellite, performs post-init
         browser tweaks, initializes navigator and UI entities, and logs in to
         satellite.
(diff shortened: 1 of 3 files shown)


airgun/session.py
```diff
@@ -615,3 +619,8 @@ def usergroup(self):
     def virtwho_configure(self):
         """Instance of Virtwho Configure entity."""
         return self._open(VirtwhoConfigureEntity)
+
+    @cached_property
+    def rhsso_login(self):
```
mirekdlugosz: We try to have these listed in alphabetical order.

airgun/session.py
```diff
@@ -231,11 +232,14 @@ def _open(self, entity):
             else:
                 endpoint = getattr(entity, 'endpoint_path', '/')
             full_url = f"https://{settings.satellite.hostname}{endpoint}"
-            self._prepare_browser(full_url)
+            if entity.__name__ == "RHSSOLoginEntity":
+                self._prepare_browser(full_url, login_type="RHSSO")
+            else:
+                self._prepare_browser(full_url)
```
mirekdlugosz: For consistency, it's better to add that to object instance as attribute instead of passing around as parameter.

Also, how many login types do we plan to support in near future? That's the difference between naming attribute `login_type` (which suggests at least three different methods, one of which must be used exclusively) and `rhsso_login` (which would be boolean: we either use RHSSO or we don't).

airgun/session.py
```diff
@@ -261,7 +265,7 @@ def _prepare_browser(self, url):
             # Navigator
             self.navigator = Navigate(self.browser)
             self.navigator.dest_dict = navigator.dest_dict.copy()
-            if self._session_cookie is None:
+            if self._session_cookie is None and login_type is None:
                 self.login.login({
                     'username': self._user, 'password': self._password})
```
mirekdlugosz: The problem I have with that is that we pass `login_type` parameter with specific value, but we never care about this value. We over-expose information - we tell Airgun something, but we don't even expect it to act upon it.

I guess it really breaks down to how many tests with alternative login flow you plan, and what will be their goal.

Option 1
----------

If we want to tell Airgun about login method we will use, it naturally follows that we expect Airgun to **use** that login method.

On Airgun side, it would mean two changes:

1. Change Session `__init__` and `__call__` to support new parameter, say `login_type`
2. Let Airgun select login provider based on parameter presence/value:
```python
if self._session_cookie is not None:
    return  # session cookie logged us in already, we are done

login_provider = self.login
if self._login_type == "RHSSO":
    login_provider = self.rhsso_login

login_provider.login({'username': self._user, 'password': self._password})
```
(You could store username and password dict in variable instead, if that's more to your liking)

Then, in your test you would do:
```
with session(login_type="RHSSO"):
    session.user.search('')  # at this point, we are already logged in
```

The obvious drawback of that solution is that logging in is a little hidden from test perspective, while using alternative provider for logging in is really the core of what this test is about.

Option 2
---------------

If we don't want Airgun to perform logging in for us (because we want that explicitly in test), I suggest we don't over-complicate it by telling Airgun specific logging method. Just teach Airgun a way to communicate our real intent, which is "I will perform login myself, don't do that for me".

Then, new session parameter should be something like `log_in`, defaulting to `True`. When it is `False`, `self.login.login()` line is not executed.

------------

Again, the choice between these two breaks down to number of tests you plan to have and what will be their purpose. For small number of tests whose main idea is "this login method works", I would recommend option 2. If you already plan tests where alternative login method is just a step towards larger goals, I would recommend option 1.

Let me know if you want to discuss that outside of github.

omkarkhatavkar: The login is an important part of authentication, but there will be tests which could have authentication and authorization. 
e.g. new rhsso user can log in? Is that user that right permissions not planned right now but in future definitely.
I like both the ideas, but I want to test should explicitly mention the login type/method. Because any new user would see and come to know what's happening instead of looking into airgun code. User should get at least come a level of understanding. Let's discuss in meeting I have some more doubts

airgun/session.py
```diff
@@ -174,8 +175,9 @@ def __init__(self, session_name=None, user=None, password=None,
         self._url = url
         self.navigator = None
         self.browser = None
+        self._login = True
```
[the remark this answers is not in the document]
omkarkhatavkar: okay

PR #461, opened 2020-02-28
Add support for assigning http proxy to repository.

jameerpathan111:


jameerpathan111:
Robottelo PR: [https://github.com/SatelliteQE/robottelo/pull/7666](https://github.com/SatelliteQE/robottelo/pull/7666)

airgun/views/repository.py
```diff
@@ -172,13 +201,17 @@ class YumRepository(View):
...
         gpg_key = EditableEntrySelect(name='GPG Key')
         download_policy = EditableEntrySelect(name='Download Policy')
         upload_content = FileInput(name='content[]')
         upload = Text("//button[contains(., 'Upload')]")
+        http_proxy_policy = EditableEntrySelect(name='HTTP Proxy')
+        proxy_policy = ConditionalSwitchableView(reference='http_proxy_policy')
+
+        @proxy_policy.register(True, default=True)
+        class NoSpecificHttpProxy(View):
+            pass
```
omkarkhatavkar: Please mentioned the comment, why we are just doing pass here?

airgun/entities/repository.py
```diff
@@ -14,8 +15,21 @@
 
 class RepositoryEntity(BaseEntity):
 
+    def _find_default_http_proxy(self, values):
```
omkarkhatavkar: Add doc string

airgun/views/repository.py
mirekdlugosz: You can find the same pattern used in `airgun/views/host.py`, `airgun/views/job_invocation.py` or `airgun/views/virtwho_configure.py`.

The thing is - `ConditionalSwitchableView` expects to find a subclass matching current value of reference element. If there is nothing matching, subclass registered with `default=True` will be used. So there must be one like that.

That subclass is empty, because new fields appear only for one value of reference field. Two other possible values do not result in any fields appearing.

I agree this pattern might be surprising, as `ConditionalSwitchableView` is complex and not used often. But since other usages of that pattern are not commented, I don't think we need comment here.

airgun/entities/repository.py
jameerpathan111: Added doc string.

airgun/views/repository.py
omkarkhatavkar: @mirekdlugosz thanks for mentioning the reason.

airgun/entities/repository.py
```diff
@@ -14,8 +15,22 @@
...
+        """ Finds the global proxy name and returns updated values accordingly."""
+        proxy = SettingsEntity(self.browser)
+        result = proxy.read(property_name='name = content_default_http_proxy')['table'][0]['Value']
+        global_proxy_name = result.split()[0]
+        if result.split()[0] == 'no':
+            values["repo_content.http_proxy_policy"] = 'Global Default (None)'
+        else:
+            values["repo_content.http_proxy_policy"] = 'Global Default ({})'.format(
+                global_proxy_name)
+        return values
```
mirekdlugosz: It really should be `if global_proxy_name == 'no':` in first line.

Also, I personally would find it more readable:
```python
if global_proxy_name == 'no':
    global_proxy_name = 'None'
values["repo_content.http_proxy_policy"] = 'Global Default ({})'.format(
            global_proxy_name)
return values
```

Use whichever you like more, nothing's really wrong with current approach.

airgun/entities/repository.py
```diff
@@ -14,8 +15,22 @@
...
+
     def create(self, product_name, values):
         """Create new repository for product"""
+        if values["repo_content.http_proxy_policy"] == "Global Default":
+            values = self._find_default_http_proxy(values)
```
mirekdlugosz: There are a lot of existing tests that don't provide `repo_content.http_proxy_policy`, and all would fail with KeyError exception now.
You need to first check if that key was provided at all:
```python
if values.get("repo_content.http_proxy_policy") == "Global Default":
    values = self._find_default_http_proxy(values)
```

airgun/entities/repository.py
```diff
@@ -35,6 +50,8 @@ def read(self, product_name, entity_name, widget_names=None):
 
     def update(self, product_name, entity_name, values):
         """Update product repository values"""
+        if values["repo_content.http_proxy_policy"] == "Global Default":
+            values = self._find_default_http_proxy(values)
```
mirekdlugosz: Same with checking for `repo_content.http_proxy_policy` here.

airgun/views/repository.py
```diff
@@ -60,16 +60,26 @@ class DockerRepository(View):
         verify_ssl = Checkbox(id='verify_ssl_on_sync')
         upstream_username = TextInput(id='upstream_username')
         upstream_password = TextInput(id='upstream_password')
-        ignore_global_proxy = TextInput(id='ignore_global_proxy')
+        http_proxy_policy = Select(id="http_proxy_policy")
+        proxy_policy = ConditionalSwitchableView(reference='http_proxy_policy')
+
+        @proxy_policy.register('Use specific HTTP Proxy')
+        class HttpProxyForm(View):
+            http_proxy = Select(id="http_proxy")
```
mirekdlugosz: That doesn't seem right. Looks like you have support for "Use specific HTTP Proxy" only when creating repository. So there is no support for setting specific proxy when editing repo, and no support for anything except specific proxy when creating repo.

I imagine you should have something like below in all create and edit subclasses:
```python
        @proxy_policy.register(True, default=True)
        class NoSpecificHttpProxy(View):
            pass

        @proxy_policy.register('Use specific HTTP Proxy')
        class HttpProxyForm(View):
            http_proxy = Select(id="http_proxy")
```
Can you try with some existing test cases, or with new small cases (like "create docker repo without setting proxy", "edit docker repo to set specific proxy"; we don't need to merge these in robottelo, they would be just for demonstration)?

airgun/entities/repository.py
jameerpathan111: @mirekdlugosz this looks much better, will use this one.

airgun/entities/repository.py
jameerpathan111: @mirekdlugosz updated PR.

airgun/entities/repository.py
jameerpathan111: updated PR.

airgun/views/repository.py
jameerpathan111: @mirekdlugosz As per our offline discussion I have filed new issue https://github.com/SatelliteQE/airgun/issues/464 for this.

mirekdlugosz: Thanks, Jameer.

Side note: I see there are still no `NoSpecificHttpProxy` blocks for create repositories. I was sure this will cause some of existing tests to fail, but it does not - at least not two I have tried. Apparently, my model on how `ConditionalSwitchableView` works is not correct. I should probably get back to this one day.

PR #466, opened 2020-03-05
cockpit test changes

tstrych:
Moving cockpit test more into airgun. 

I was also thinking to use view
`webconsole_view = WebConsoleView(self.browser)`

```
class WebConsoleView(View):
    ROOT = "//iframe"
    hostname = Text("//a[@id='system_information_hostname_button']")

    @property
    def is_displayed(self):
            self.hostname, exception=False) is not None
```

I was unable to get into the frame for hostname. 

I thought, first view to get me into frame second view to get me inside the frame. I was unable to create that. 

I also find issue on widgetastic not sure if the frame I mean is the same as they mentioned in the [issue](https://github.com/RedHatQE/widgetastic.core/issues/98).

I think this approach, not all logic in the test and not all in the airgun is somewhere in the middle and I think it's just not that important to spend more time with different approaches on this. 

But I would be happy for your feedback here, 
there is still open #7554 PR in robottelo, if you want to see conversation about this before.

### DIFF
--- a/airgun/entities/host.py
+++ b/airgun/entities/host.py
@@ -228,7 +228,9 @@ class HostEntity(BaseEntity):
             handle_exception=True, timeout=10,
             logger=view.logger
         )
-        return hostname_button_view
+        hostname = hostname_button_view[0].text
+        self.browser.switch_to_main_frame()
+        return hostname
 
 
 @navigator.register(HostEntity, 'All')
--- a/airgun/entities/host.py
+++ b/airgun/entities/host.py
@@ -210,14 +210,25 @@ class HostEntity(BaseEntity):
         view.flash.assert_no_error()
         view.flash.dismiss()
 
-    def open_webconsole(self, entity_name):
-        """Navigate to host's webconsole
+    def has_working_webconsole(self, entity_name):
+        """Navigate to host's webconsole and return the hostname from the cockpit page
 
         :param str entity_name: The host name for which to set the parameter value.
         """
         view = self.navigate_to(self, 'Details', entity_name=entity_name)
         view.webconsole.click()
         view.validations.assert_no_errors()
+        # the remote host content is loaded in an iframe, let's switch to it
+        self.browser.selenium.switch_to.frame(0)
+        webconsole_view = self.browser.selenium.find_elements_by_id(
+            'system_information_hostname_button'
+        )
+        wait_for(
+            lambda: webconsole_view[0].text != '',
+            handle_exception=True, timeout=10,
+            logger=view.logger
+        )
+        return webconsole_view
 
 
 @navigator.register(HostEntity, 'All')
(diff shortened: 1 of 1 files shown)


airgun/entities/host.py
```diff
@@ -210,14 +210,25 @@ def set_puppet_class_parameter_value(self, entity_name, name, value):
         view.flash.assert_no_error()
         view.flash.dismiss()
 
-    def open_webconsole(self, entity_name):
-        """Navigate to host's webconsole
+    def has_working_webconsole(self, entity_name):
+        """Navigate to host's webconsole and return the hostname from the cockpit page
```
omkarkhatavkar: Instead changing the whole use case of this function, I would have added an optional parameter to return the view. @mirekdlugosz, @tstrych  what do you think about it?

mirekdlugosz: Yes, function with that name should return boolean value.
No, I don't think optional parameter to return view would be appropriate. So far we have managed without exposing views to robottelo, I would like to try to keep it this way.

This should be named something like `get_webconsole_content()`, and do:
1. navigate to specific page
2. switch frames
3. get content of `webconsole_view` (`wait_for` should return it on success, just wrapped in some structure)
4. switch frames back
5. Return content from step 3

airgun/entities/host.py
```diff
@@ -210,14 +210,25 @@ def set_puppet_class_parameter_value(self, entity_name, name, value):
...
 
         :param str entity_name: The host name for which to set the parameter value.
         """
         view = self.navigate_to(self, 'Details', entity_name=entity_name)
         view.webconsole.click()
         view.validations.assert_no_errors()
+        # the remote host content is loaded in an iframe, let's switch to it
+        self.browser.selenium.switch_to.frame(0)
+        webconsole_view = self.browser.selenium.find_elements_by_id(
```
mirekdlugosz: It's not really `webconsole_view`, but one button on that page, isn't it? I think this variable name is little misleading.

tstrych:
What about now ?

tstrych:
@mirekdlugosz 
I was unable to change frame and then return the value.

Maybe other way can be to get 
```
hostname_text = hostname_button_view[0].text
switch view back
return hostname_text
```

mirekdlugosz:
> Maybe other way can be to get
> 
> ```
> hostname_text = hostname_button_view[0].text
> switch view back
> return hostname_text
> ```

Yes, I think that's exactly how it should be handled - take the value, assign to variable, switch frame back, return variable.

If `self.browser.switch_to_main_frame()` doesn't work as advertised, then we can deal with that later. I assumed it does what it was supposed to do.

tstrych:
@mirekdlugosz thanks for your cooperation :) , yes  `self.browser.switch_to_main_frame()` works perfectly. 

and I though I can use `self.browser.switch_to_frame()` but it's not worth of time.

PR #467, opened 2020-03-16
Add --ignore-certificate-errors flag for Chrome 80+

mirekdlugosz:
On Chrome 80 and above, opening Satellite results in `NET::ERR_CERT_AUTHORITY_INVALID` error page displayed ("Your connection is not private"). Starting Chrome with `--ignore-certificate-errors` makes this error go away and allows UI test to work.

`--ignore-certificate-errors` has been available in Chrome for at least 10 years, and Chrome team has been trying to get rid of it since at least 2014. It's not publicly advertised and is probable to stop working sometime soon, but it does get the job done now.

[This StackOverflow question](https://stackoverflow.com/q/7580508/3552063) deals with similar issue and provides multiple solutions applied at various levels. I'm open to discussion about better approach than one in this PR.

Verified on Chrome 80 and Chrome 81 (beta, to be released tomorrow).

### DIFF
--- a/airgun/browser.py
+++ b/airgun/browser.py
@@ -194,6 +194,7 @@ class SeleniumBrowserFactory(object):
             prefs = {'download.prompt_for_download': False}
             options.add_experimental_option("prefs", prefs)
             options.add_argument('disable-web-security')
+            options.add_argument('ignore-certificate-errors')
             if browseroptions:
                 for opt in browseroptions.split(';'):
                     options.add_argument(opt)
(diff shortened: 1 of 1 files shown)


mirekdlugosz:
> @mirekdlugosz is this tested with chrome 80 with Zalenium locally?

No, I tested this with our shared Zalenium (on whatever version of Chrome that thing uses), and locally with chrome 80 and chrome 81.
You think running this through Zalenium will make any difference?

omkarkhatavkar:
> You think running this through Zalenium will make any difference

That will be okay, Just wanted to ensure this tested with Zalenium.

PR #468, opened 2020-03-17
added ext logout support code

omkarkhatavkar:
### PR Description 
Adding support to Rhsso ext_logout scenarios. Which means any user logged in into Satellite as RHSSO User when they sign out there are seeing https://satellite_host/users/extlogout page. This Pr adds that support so we can validate scenarios based on the sign out in Satellite.

### DIFF
--- a/airgun/entities/rhsso_login.py
+++ b/airgun/entities/rhsso_login.py
@@ -2,28 +2,27 @@ from airgun.entities.base import BaseEntity
 from airgun.navigation import NavigateStep
 from airgun.navigation import navigator
 from airgun.views.common import BaseLoggedInView
-from airgun.views.rhsso_login import RhssoExtLogoutView
+from airgun.views.rhsso_login import RhssoExternalLogoutView
 from airgun.views.rhsso_login import RhssoLoginView
 
 
 class RHSSOLoginEntity(BaseEntity):
 
-    def login(self, values):
-        view = self.navigate_to(self, 'NavigateToLogin')
-        view.fill(values)
-        view.submit.click()
-
-    def login_again(self):
-        view = RhssoExtLogoutView(self.browser)
-        view.login_again.click()
-        return view.read()
+    def login(self, values, external_login=False):
+        if external_login:
+            view = RhssoExternalLogoutView(self.browser)
+            view.login_again.click()
+        else:
+            view = self.navigate_to(self, 'NavigateToLogin')
+            view.fill(values)
+            view.submit.click()
 
     def logout(self):
         view = BaseLoggedInView(self.browser)
-        view.taxonomies.select_logout()
+        view.select_logout()
         view.flash.assert_no_error()
         view.flash.dismiss()
-        view = RhssoExtLogoutView(self.browser)
+        view = RhssoExternalLogoutView(self.browser)
         return view.read()
 
 
--- a/airgun/entities/rhsso_login.py
+++ b/airgun/entities/rhsso_login.py
@@ -1,7 +1,9 @@
 from airgun.entities.base import BaseEntity
 from airgun.navigation import NavigateStep
 from airgun.navigation import navigator
-from airgun.views.rhsso_login import LoginView
+from airgun.views.rhsso_login import RhssoLoginView
+from airgun.views.rhsso_login import RhssoExtLogoutView
+from airgun.views.common import BaseLoggedInView
 
 
 class RHSSOLoginEntity(BaseEntity):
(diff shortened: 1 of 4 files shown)


airgun/entities/rhsso_login.py
```diff
@@ -11,10 +13,23 @@ def login(self, values):
         view.fill(values)
         view.submit.click()
 
+    def login_again(self):
+        view = RhssoExtLogoutView(self.browser)
+        view.login_again.click()
+        return view.read()
```
mirekdlugosz: I feel there should be better way to approach that, but I'm not 100% sure without seeing how things work on Robottelo side.

As an idea, could we move logic to `NavigateToLogin` class? So you would have single `RHSSOLoginEntity.login` method that uses `self.navigate_to(self, 'NavigateToLogin')`. But `navigate_to()` would be smart to look where it is and take action based on that. If it is on `RhssoLoginView`, it would do one thing. If it is on `RhssoExtLogoutView`, it would first click `login_again` button.

Take a look at [navmazing documentation](https://github.com/RedHatQE/navmazing) to see if there is any method you could hook into.

As discussed off GitHub, navmazing is not really designed for this kind of circular navigations, so I'll understand if there is nothing we could do. But I think we should at least explore other possibilities.

omkarkhatavkar: @mirekdlugosz  below code snippet makes sense? 

``` python
    def login(self, values, ext_login=False):
        if not ext_login:
            view = RhssoExtLogoutView(self.browser)
            view.login_again.click()
            return view.read()
        else:
            view = self.navigate_to(self, 'NavigateToLogin')
            view.fill(values)
            view.submit.click()
```

airgun/views/rhsso_login.py
```diff
@@ -13,3 +13,13 @@ class LoginView(View, ClickableMixin):
     def is_displayed(self):
         return self.browser.wait_for_element(
             self.username, exception=False) is not None
+
+
+class RhssoExtLogoutView(View, ClickableMixin):
```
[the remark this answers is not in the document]
omkarkhatavkar: I think it is better to keep the same, I do not know what could be the full word for this. If you look at the locators as well they are kept like extlogin and extlogout. I do not want to give name as Welcome Page or something which make confusions for people reading this code later. If you have a better name please suggest me

airgun/widgets.py
```diff
@@ -638,7 +638,6 @@ def fill(self, item):
 
 
 class Search(Widget):
-    """Searchbar for table filtering"""
```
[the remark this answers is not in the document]
omkarkhatavkar: will remove that

airgun/widgets.py
```diff
@@ -835,6 +834,8 @@ class ContextSelector(Widget):
     CURRENT_LOC = '//li[@id="location-dropdown"]/a'
     ORG_LOCATOR = '//li[@id="organization-dropdown"]/ul/li/a[contains(.,{})]'
     LOC_LOCATOR = '//li[@id="location-dropdown"]/ul/li/a[contains(.,{})]'
+    ACCOUNT_MENU = Text("//a[@id='account_menu']")
+    LOGOUT = Text("//a[@href='/users/logout']")
```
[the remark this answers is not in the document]
omkarkhatavkar: @mirekdlugosz Looks like too many things we are talking here. Let me see what extent of work I can incorporate for this PR. Whatever you had in your mind are great suggestions, let me try out that if those work in the first attempt I will invest more time, If not we can create the airgun issue.

airgun/views/rhsso_login.py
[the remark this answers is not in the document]
omkarkhatavkar: I gotcha you, I just remembered it should external

omkarkhatavkar:
> 1. Please move logging out to base class.
> 2. I would like to see if we can do without `login_again()` method. It seems awkward, but I'll understand if there is no better way.
> 
> I need to see Robottelo changes before I give ACK, so I have better idea what is going on.

Robottelo PR `https://github.com/SatelliteQE/robottelo/pull/7698`

airgun/views/rhsso_login.py
```diff
@@ -13,3 +13,13 @@ class LoginView(View, ClickableMixin):
     def is_displayed(self):
         return self.browser.wait_for_element(
             self.username, exception=False) is not None
+
+
+class RhssoExternalLogoutView(View, ClickableMixin):
+    login_again = Text('//a[@href="/users/extlogin"]')
+    logo = Text('//a[@alt="logo')
```
mirekdlugosz: Does that selector match anything? As far as I know, `alt` is not legal attribute of `a`. Perhaps you meant `img`?

airgun/views/common.py
```diff
@@ -38,6 +38,13 @@ class BaseLoggedInView(View):
         locator='//div[@class="toast-notifications-list-pf"]')
     validations = ValidationErrors()
     current_user = Text("//a[@id='account_menu']")
+    ACCOUNT_MENU = Text("//a[@id='account_menu']")
+    LOGOUT = Text("//a[@href='/users/logout']")
```
mirekdlugosz: Is there a reason for these two to be in all-caps?

omkarkhatavkar: copy paste :D

airgun/views/rhsso_login.py
omkarkhatavkar: copy + paste

PR #469, opened 2020-03-25
Fix task table locator

swadeley:
Hello

This change works in Firefox Dev Edition's console. But still not able to read the table in a test.

### DIFF
--- a/airgun/views/task.py
+++ b/airgun/views/task.py
@@ -28,7 +28,7 @@ class TasksView(BaseLoggedInView, SearchableViewMixin):
         "//div[./button[@id='tasks-dashboard-time-period-dropdown']]"
     )
     table = SatTable(
-        ".//div[@id='tasks-table']/table",
+        ".//div[@class='tasks-table']//table",
         column_widgets={
             'Action': Text('./a'),
         }


airgun/views/task.py
```diff
@@ -28,7 +28,7 @@ class TasksView(BaseLoggedInView, SearchableViewMixin):
         "//div[./button[@id='tasks-dashboard-time-period-dropdown']]"
     )
     table = SatTable(
-        ".//div[@id='tasks-table']/table",
+        ".//div[@class='tasks-table']",
```
mirekdlugosz: I'm pretty sure `Table` widget should match `<table>` element, so you want to retain this `/table` part.

Also, seems in latest snap entire table is wrapped in `<div>`? I think that wasn't there before and looks like a small bug. So make that `//table`.

swadeley: OK, // table works, /table did not.
The extra <div> is there in previous snap too.
Thank you

swadeley: @mirekdlugosz thanks a lot, that fixed the reading of the table in my testing.

swadeley:
Before I got empty dict, like so []

No it works:

In [2]: session.task.search('Update package for g5vni1949pbkhpui.lab.eng.rdu2.redhat.com walrus')                 
Out[2]: 
[{'Select all rows': 'Select row',
  'Action': 'Update package for g5vni1949pbkhpui.lab.eng.rdu2.redhat.com walrus',
  'State': 'stopped',
  'Result': 'warning',
  'Started at': 'March 25, 2020, 11:41:27 AM',
  'Duration': '21 seconds',
  'Operation': 'Cancel'}]

In [3]:

swadeley:
thank you all

PR #470, opened 2020-03-26
Added host view for AzureRm form

vijay8451:


airgun/views/host.py
```diff
@@ -315,6 +316,33 @@ class storage(RemovableWidgetsItemsListView):
...
+            TAB_NAME = 'Virtual Machine'
+            resource_group = FilteredDropdown(id='s2id_azure_rm_rg')
+            vm_size = FilteredDropdown(id='s2id_azure_rm_size')
+            platform = FilteredDropdown(id='s2id_host_compute_attributes_platform')
+            username = TextInput(id='host_compute_attributes_username')
+            password = TextInput(id='host_compute_attributes_password')
+            ssh_key = TextInput(id='host_compute_attributes_ssh_key_data')
+            premium_os_disk = Checkbox(id='host_compute_attributes_premium_os_disk')
+            os_disk_caching = FilteredDropdown(
+                locator="//div[@id='compute_resource']/div[9]//b[@role='presentation']")
```
vijay8451: let me know if you see a better locator then above one..

airgun/views/host.py
```diff
@@ -240,6 +240,7 @@ class host(SatTab):
         inherit_deploy_option = ToggleButton(
             locator=".//div[label[@for='compute_resource_id']]//button")
         deploy = FilteredDropdown(id='host_compute_resource')
+        compute_profile = FilteredDropdown(id='s2id_host_compute_profile_id')
```
mirekdlugosz: I'm on snap 17 and I don't see "Compute Profile" between "Deploy On" and "Lifecycle Environment".
Is there anything specific that must be done for this one to appear?

vijay8451: It only comes, when you select `Deploy` on a Compute Resource

airgun/views/host.py
```diff
@@ -315,6 +316,32 @@ class storage(RemovableWidgetsItemsListView):
...
+            os_disk_caching = FilteredDropdown(id="s2id_host_compute_attributes_os_disk_caching")
+            custom_script_command = TextInput(id="host_compute_attributes_script_command")
+            file_uris = TextInput(id="host_compute_attributes_script_uris")
+
+        @View.nested
+        class operating_system(SatTab):
+            TAB_NAME = 'Operating System'
+
+            architecture = FilteredDropdown(id='host_architecture')
+            operating_system = FilteredDropdown(id='host_operatingsystem')
```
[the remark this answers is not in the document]
vijay8451: I too the reference from old code ... it good now :+1:

airgun/views/host.py
mirekdlugosz: OK, figured that out. When you go to Compute Profiles page (there are some by default), you will have list of compute resources, and ability to change VM attributes of that compute profile in context of specific compute resource (at least for Azure these are VM attributes). 

Once you have compute profile with VM attribute for compute resource, then when you select this compute resource when creating host, new option will appear.

I'm not sure if we have a way in Airgun to specific complex visibility logic like that, so let's just add this here and hope for the best.

PR #471, opened 2020-03-30
Sync Templates Support

jyejare:
Support Sync (Import / Export) of templates from UI.

### DIFF
--- a/airgun/entities/sync_templates.py
+++ b/airgun/entities/sync_templates.py
@@ -1,0 +1,47 @@
+from wait_for import wait_for
+
+from airgun.entities.base import BaseEntity
+from airgun.navigation import NavigateStep
+from airgun.navigation import navigator
+from airgun.views.sync_templates import SyncTemplatesView
+from airgun.views.sync_templates import TemplatesReportView
+
+
+class SyncTemplatesEntity(BaseEntity):
+    endpoint_path = '/template_syncs'
+
+    def sync(self, values):
+        """Import Export Switch Action Entity"""
+        view = self.navigate_to(self, 'Sync')
+        view.fill(values)
+        view.submit.click()
+        self.browser.plugin.ensure_page_safe()
+        if view.validations.messages:
+            raise AssertionError(
+                f'Validation Errors are present on Page. Messages are {view.validations.messages}')
+        reports_view = TemplatesReportView(self.browser)
+        wait_for(
+            lambda: reports_view.is_displayed is True,
+            timeout=60,
+            delay=1,
+            logger=reports_view.logger
+        )
+        return reports_view.title.read()
+
+
+@navigator.register(SyncTemplatesEntity, 'Main')
+class SyncMainPageNavigation(NavigateStep):
+    """Navigate to Import/Export Templates page"""
+    VIEW = SyncTemplatesView
+
+    def step(self, *args, **kwargs):
+        self.view.menu.select('Hosts', 'Sync Templates')
+
+
+@navigator.register(SyncTemplatesEntity, 'Sync')
+class SyncTemplatesActionNavigation(NavigateStep):
+    """Navigate to Import/Export Templates page"""
+    VIEW = SyncTemplatesView
+
+    def prerequisite(self, *args, **kwargs):
+        return self.navigate_to(self.obj, 'Main')
--- a/airgun/session.py
+++ b/airgun/session.py
@@ -69,6 +69,7 @@ from airgun.entities.smart_variable import SmartVariableEntity
 from airgun.entities.subnet import SubnetEntity
 from airgun.entities.subscription import SubscriptionEntity
 from airgun.entities.sync_status import SyncStatusEntity
+from airgun.entities.sync_templates import SyncTemplatesEntity
 from airgun.entities.syncplan import SyncPlanEntity
 from airgun.entities.task import TaskEntity
 from airgun.entities.trend import TrendEntity
@@ -599,6 +600,11 @@ class Session(object):
     def sync_status(self):
         """Instance of Sync Status entity"""
         return self._open(SyncStatusEntity)
+
+    @cached_property
+    def sync_template(self):
+        """Instance of Sync Templates entity"""
+        return self._open(SyncTemplatesEntity)
 
     @cached_property
     def task(self):
--- a/airgun/widgets.py
+++ b/airgun/widgets.py
@@ -792,7 +792,8 @@ class ValidationErrors(Widget):
         "not(contains(@style,'display:none'))]"
     )
     ERROR_MESSAGES = (
-        ".//*[(contains(@class,'error-msg-block') "
+        ".//*[(contains(@class, 'alert base in fade alert-danger')"
+        "or contains(@class,'error-msg-block')"
         "or contains(@class,'error-message') "
         "or contains(@class,'editable-error-block')) "
         "and not(contains(@style,'display:none'))]"
@@ -2192,7 +2193,7 @@ class PopOverWidget(Widget):
                     widget.fill(item)
                     break
         else:
-            raise ReadOnlyWidgetError(f'This field setting is read-only')
+            raise ReadOnlyWidgetError('This field setting is read-only')
         self.submit.click()
 
     def read(self):
--- a/airgun/views/sync_templates.py
+++ b/airgun/views/sync_templates.py
@@ -1,0 +1,70 @@
+from wait_for import wait_for
+from widgetastic.widget import Checkbox
+from widgetastic.widget import ConditionalSwitchableView
+from widgetastic.widget import Select
+from widgetastic.widget import Text
+from widgetastic.widget import TextInput
+from widgetastic.widget import View
+from widgetastic_patternfly import BreadCrumb
+
+from airgun.views.common import BaseLoggedInView
+from airgun.widgets import RadioGroup
+
+
+class SyncTemplatesView(BaseLoggedInView):
+    breadcrumb = BreadCrumb()
+    title = Text("//h2[contains(., 'Import or Export Templates')]")
+    sync_type = RadioGroup("//div[label[contains(., 'Action type')]]")
+    submit = Text("//button[@type='submit']")
+
+    template = ConditionalSwitchableView(reference='sync_type')
+
+    @property
+    def is_displayed(self):
+        breadcrumb_loaded = self.browser.wait_for_element(
+            self.breadcrumb, exception=False)
+        return (
+                breadcrumb_loaded
+                and self.browser.wait_for_element(self.title, exception=False) is not None)
+
+    def before_fill(self, values):
+        """Wait for Sync Type Radio Button to be displayed"""
+        wait_for(
+            lambda: self.sync_type.is_displayed,
+            timeout=10,
+            delay=1,
+            logger=self.logger
+        )
+
+    @template.register('Import')
+    class ImportTemplates(View):
+        associate = Select(name='associate')
+        branch = TextInput(name='branch')
+        dirname = TextInput(name='dirname')
+        filter = TextInput(name='filter')
+        force_import = Checkbox(name='force')
+        lock = Checkbox(name='lock')
+        negate = Checkbox(name='negate')
+        prefix = TextInput(name='prefix')
+        repo = TextInput(name='repo')
+
+    @template.register('Export')
+    class ExportTemplates(View):
+        branch = TextInput(name='branch')
+        dirname = TextInput(name='dirname')
+        filter = TextInput(name='filter')
+        metadata_export_mode = Select(name='metadata_export_mode')
+        negate = Checkbox(name='negate')
+        repo = TextInput(name='repo')
+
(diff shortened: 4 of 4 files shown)


airgun/entities/sync_templates.py
```diff
@@ -0,0 +1,53 @@
...
+@navigator.register(SyncTemplatesEntity, 'All')
+class ShowMainPageActionNavigation(NavigateStep):
+    """Navigate to the page thats main Import Export Templates page"""
+    VIEW = SyncTemplatesView
+
+    def step(self, *args, **kwargs):
+        self.view.menu.select('Hosts', 'Sync Templates')
+
+
+@navigator.register(SyncTemplatesEntity, 'import')
```
jameerpathan111:
```suggestion
@navigator.register(SyncTemplatesEntity, 'Import')
```
No idea why we have to do it, but rest of navigators in Airgun following the sae approach.
It would be good if we keep it that way :P

airgun/entities/sync_templates.py
```diff
@@ -0,0 +1,53 @@
...
+    """Navigate to Import Templates page"""
+    VIEW = SyncTemplatesView.ImportTemplates
+
+    prerequisite = NavigateToSibling('All')
+
+    def step(self, *args, **kwargs):
+        self.parent.import_template.click()
+
+
+@navigator.register(SyncTemplatesEntity, 'export')
```
jameerpathan111:
```suggestion
@navigator.register(SyncTemplatesEntity, 'Export')
```

airgun/views/template.py
```diff
@@ -0,0 +1,42 @@
...
+
+from airgun.views.common import BaseLoggedInView
+from airgun.widgets import ActionsDropdown
+from airgun.widgets import RadioGroup
+
+
+class SyncTemplatesView(BaseLoggedInView):
+    title = Text("//h2[contains(., 'Import or Export Templates')]")
+    import_export = RadioGroup("//div[@class='col-md-6']")
+    submit = Text("//button[@type='submit']")
```
jameerpathan111:
```suggestion
    submit = Text("//button[@type='submit']")
    cancel = Text('//a[text()="Cancel"]')
```

airgun/entities/sync_templates.py
mirekdlugosz: Strictly speaking, we don't. Navigation endpoints could be anything, `fdh874rfyj` is as good as any. Capitalization is just a convention that we follow.

But yes, being consistent and following established conventions is definitely desired.

airgun/entities/sync_templates.py
```diff
@@ -0,0 +1,53 @@
...
+        view.submit.click()
+
+    def export_templates(self, values):
+        """Export Templates entity"""
+        view = self.navigate_to(self, 'export')
+        view.fill(values)
+        view.submit.click()
+
+
+@navigator.register(SyncTemplatesEntity, 'All')
```
mirekdlugosz: While it's true that almost all of our entities register "All" endpoint, it's not strictly required. I always understood "All" as "list of all entities", and this description doesn't really fit this page. I would rather name it "Main", or "ImportExport", or "Sync".

airgun/entities/sync_templates.py
mirekdlugosz: I'm not fan of these navigation endpoints. These are not two distinct pages, this is one page that shows or hides some fields depending on which radio value is selected in first field. This is what `ConditionalSwitchableView` widget is for, and that should be single navigation endpoint.

airgun/views/template.py
```diff
@@ -0,0 +1,42 @@
...
+
+    @property
+    def is_displayed(self):
+        return self.browser.wait_for_element(self.title, exception=False) and \
+               self.browser.wait_for_element(self.import_export) is not None
+
+    @View.nested
+    class ImportTemplates(View):
```
mirekdlugosz: As mentioned, these should use `ConditionalSwitchableView`. You can find many examples in Airgun, but it would go something like this:
```python
class SyncTemplatesView(BaseLoggedInView):
    import_export = RadioGroup("//div[@class='col-md-6']")
    action_type = ConditionalSwitchableView(reference='import_export')

    @action_type.register('Import')
    class ImportTemplates(View):
        associate = ActionsDropdown("//select[@name='associate']")
        # Others skipped for brevity

    @action_type.register('Export')
    class ExportTemplates(View):
        branch = TextInput("//input[@name='branch']")
        # Others skipped for brevity
```

Written like that, on robottelo you will need to do something like:
```
session.sync_template.import({'import_export': 'Import', 'associate': 'some value'})
```
Since `import()` action will always select `Import`, and `export()` will always select `Export`, it makes sense to relieve user from specifying that option manually. Just add this key with specific value to `values` in `SyncTemplatesEntity`.

Or re-write this entity to expose single action, say `submit`, and require user to specify this key:
```
session.sync_template.sync({'import_export': 'Import', 'associate': 'some value'})
```

airgun/views/template.py
```diff
@@ -0,0 +1,42 @@
...
+        import_template = Text("//input[@value='import']")
+        associate = ActionsDropdown("//select[@name='associate']")
+        branch = TextInput("//input[@name='branch']")
```
mirekdlugosz: `TextInput` supports simplified specification, where you don't have to pass full location, but can use named argument `name` or `id`:
```python
branch = TextInput(name="branch")
```

The same applies to `Checkbox`. 

Up to, full locators are not wrong, strictly speaking. But simplified version is easier to read.

airgun/views/template.py
```diff
@@ -0,0 +1,42 @@
...
+from widgetastic.widget import View
+
+from airgun.views.common import BaseLoggedInView
+from airgun.widgets import ActionsDropdown
+from airgun.widgets import RadioGroup
+
+
+class SyncTemplatesView(BaseLoggedInView):
+    title = Text("//h2[contains(., 'Import or Export Templates')]")
+    import_export = RadioGroup("//div[@class='col-md-6']")
```
omkarkhatavkar: I do not have any issues with this locator, good way to add like e.g. https://github.com/SatelliteQE/airgun/blob/ab272c64a23d485c0932ea5b5bafd624042028cf/airgun/views/host.py#L353
`action_type = RadioGroup(locator="//div[label[contains(., 'Action type')]]") `

airgun/views/template.py
omkarkhatavkar: I think this needs to be added as there is some UI flow out there, which navigates to 
Import from https   github com theforeman community-templates git

I have one conflict here to mention that in Satellite we are having the all entities view by default on navigation from the menu bar, but in this case it actually creates entity page on the navigation. (this really different than the overall design we have in Satellite). This is unrelated to this PR by the way, I'm just expressing my thoughts.

jyejare: This is a reports page basically which is not necessarily to be implemented as the verification can be done from templates pages directly. But I understood that this has to be implemented to test UI reports of export/import sometimes.

airgun/session.py
```diff
@@ -600,6 +601,11 @@ def sync_status(self):
         """Instance of Sync Status entity"""
         return self._open(SyncStatusEntity)
 
+    @cached_property
+    def sync_templates(self):
```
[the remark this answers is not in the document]
jyejare: The Satellite page itself needs to be navigates https://<sat_fqdn>/sync_templates that means that the purpose of the feature is to sync multiple templates but still I would follow the rule of Airgun.

airgun/entities/sync_templates.py
```diff
@@ -0,0 +1,53 @@
...
+from airgun.entities.base import BaseEntity
+from airgun.navigation import NavigateStep
+from airgun.navigation import navigator
+from airgun.views.template import SyncTemplatesView
+
+
+class SyncTemplatesEntity(BaseEntity):
+    endpoint_path = '/template_syncs'
+
+    def import_templates(self, values):
```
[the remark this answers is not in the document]
jyejare: import is a reserved keyword in python and hence it doesnt allows me to use again :)

airgun/entities/sync_templates.py
```diff
@@ -0,0 +1,53 @@
...
+        """Import Templates entity"""
+        view = self.navigate_to(self, 'import')
+        view.fill(values)
+        view.submit.click()
```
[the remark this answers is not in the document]
jyejare: No, There won't be any task for export/import of templates, you will directly be redirected to the reports page. So I though its not required here. But I think I should be able to wait for the reports page to be displayed. Good point.

airgun/entities/sync_templates.py
jyejare: TBH, I was planning to remove this altogether cause, the endpoint_path = '/template_syncs' directly takes me to that page. :)

airgun/entities/sync_templates.py
jyejare: Yeah , You are correct and thats the reason I had :

```
import_export = RadioGroup("//div[@class='col-md-6']")
```
in my code to implement ```ConditionalSwitchableView``` but not sure why I change my mind :)

jyejare:
@mirekdlugosz @omkarkhatavkar Thanks for all your valuable feedback.

I know in airgun things can be implemented in different ways and I choose one of that. But thanks all for your comments that could help improve things for a newbie like me of Airgun :)

airgun/entities/sync_templates.py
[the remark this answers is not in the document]
jyejare: Done !

jyejare:
@omkarkhatavkar @mirekdlugosz @jameerpathan111 , Currently the reporting of syncing templates is missing search field (just like audit has, and I have raised a bug for it https://bugzilla.redhat.com/show_bug.cgi?id=1839108) and hence it's not feasible to read all the templates data including expanding field for each template. This all reports reading is going to take so much time for a test approx 5 minutes or so.

Though its not difficult but its not worth implementing that tricky Views in Airgun. Instead we can assert the template import from satellite (Hosts -> Provisioning Templates, Host -> Job Templates etc by simply searching by name) and export to the local directory through os operations/library. So that, the templates reporting implementation isnt a blocker to test now.

With all those considerations, we should be good to merge this PR. Let me know if you have some more comments.

omkarkhatavkar:
@mirekdlugosz @jameerpathan111 I had a discussion with Jitendra, he already has raised Bugzilla for the same, In case the if he the fix we anyway can add the search functionality. @jyejare please add the airgun issue for that or Jira just to track that we should not miss in future. From my view, I'm not going to block this PR as we need the test coverage. Adding those changes can be done in future if we already have the issue for that. 

ACK

jyejare:
@omkarkhatavkar @jameerpathan111 ,

The results are here : https://github.com/SatelliteQE/robottelo/pull/7800

jyejare:
@omkarkhatavkar @jameerpathan111 Mind merging this PR ? :)

jameerpathan111:
> @omkarkhatavkar @jameerpathan111 Mind merging this PR ? :)

@jyejare I am not an airgun reviewer yet. :P

omkarkhatavkar:
@jyejare waiting for @mirekdlugosz @lhellebr. If they agree we should merge the PR. I'm good with the PR approval

lhellebr:
I'll leave this PR for Miroslaw  to review since he's already been looking into it.

airgun/widgets.py
```diff
@@ -792,7 +792,8 @@ class ValidationErrors(Widget):
         "not(contains(@style,'display:none'))]"
     )
     ERROR_MESSAGES = (
-        ".//*[(contains(@class,'error-msg-block') "
+        ".//*[(contains(@class, 'alert base in fade alert-danger')"
```
mirekdlugosz, line 795: Since CSS classes are set, i.e. their order doesn't matter and isn't guaranteed, this looks rather brittle. Next time this alert has classes set to "base alert alert-danger fade in" and XPath won't match it.

I wasn't able to open this particular element in my quick test. How to access it? If I see it, I might have some idea how to better approach it.

airgun/entities/sync_templates.py
```diff
@@ -0,0 +1,47 @@
...
+    def sync(self, values):
+        """Import Export Switch Action Entity"""
+        view = self.navigate_to(self, 'Sync')
+        view.fill(values)
+        view.submit.click()
+        self.browser.plugin.ensure_page_safe(timeout='600s')
+        if view.validations.messages:
+            raise AssertionError(
+                f'Validation Errors are present on Page. Messages are {view.validations.messages}')
+        reports_view = TemplatesReportView(self.browser)
```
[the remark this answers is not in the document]
mshriver: @mirekdlugosz agreed on the approach, but I like this sync method actually returning the redirected view, making it that much easier on the caller to subsequently read the page.

I think that by default this method should initiate the sync and return - if we want the method to wait for the sync to complete before returning, that should be controlled by a `wait` (or similarly named) kwarg.
Thoughts?

jyejare:
@mshriver Thanks for peaching in. Thanks @mirekdlugosz for merging this PR.

PR #472, opened 2020-04-01
Added read title method in Settings

devendra104:
The intent of this method to get the title name of the settings page, It the user has admin right then settings page title would be "Settings" and If it has no admin right then the title would be "Permission denied".

### DIFF
--- a/airgun/entities/settings.py
+++ b/airgun/entities/settings.py
@@ -26,6 +26,10 @@ class SettingsEntity(BaseEntity):
         view.validations.assert_no_errors()
         view.wait_for_update()
 
+    def read_title(self):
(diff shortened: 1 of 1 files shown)


devendra104:
We are using the read_title method in https://github.com/SatelliteQE/robottelo/pull/7710

devendra104:
> Well, `Session` instance exposes browser property publicly, so you can just do this on Robottelo side:
> 
> ```python
> session.browser.title
> ```
> 
> I appreciate the effort, but I believe there is no change on Airgun side required. Agreed?

Yes, you are right, It is working(session.settings.session.browser.title).
So now, I am closing this PR

PR #474, opened 2020-04-02
Fixes 349 Added browser logout method

devendra104:
Raised this PR by taking the reference of #397 to fix one of the required login page footer text validation after logout https://github.com/SatelliteQE/robottelo/pull/7710

Test Result
```
$ pytest tests/foreman/ui/test_settings.py::test_positive_update_login_page_footer_text_with_long_string
=============================================================================================== test session starts ===============================================================================================
platform linux -- Python 3.6.6, pytest-4.6.3, py-1.8.1, pluggy-0.13.1
collecting ... 2020-04-02 12:59:37 - conftest - DEBUG - Collected 1 test cases
2020-04-02 18:29:40 - robottelo.ssh - DEBUG - Instantiated Paramiko client 0x7f3f7b3695c0
2020-04-02 18:29:40 - robottelo.ssh - INFO - >>> rpm -q satellite
2020-04-02 18:29:42 - robottelo.ssh - INFO - <<< stdout
satellite-6.7.0-7.el7sat.noarch

2020-04-02 18:29:42 - robottelo.ssh - DEBUG - Destroyed Paramiko client 0x7f3f7b3695c0
2020-04-02 18:29:42 - robottelo.host_info - DEBUG - Host Satellite version: 6.7
2020-04-02 18:29:42 - robottelo.helpers - DEBUG - Generated file bz_cache.json with BZ collect data
collected 1 item                                                                                                                                                                                                  

tests/foreman/ui/test_settings.py .                                                                                                                                                                         [100%]

============================================================================================ 1 passed in 56.60 seconds ============================================================================================

```

### DIFF
--- a/airgun/session.py
+++ b/airgun/session.py
@@ -216,9 +216,7 @@ class Session(object):
             u'Stopping UI session %r for user %r', self.name, self._user)
         passed = True if exc_type is None else False
         try:
-            if passed:
-                self.login.logout()
-            else:
+            if not passed:
                 self.take_screenshot()
         except Exception as err:
             LOGGER.exception(err)
--- a/airgun/views/login.py
+++ b/airgun/views/login.py
@@ -5,11 +5,12 @@ from widgetastic.widget import View
 
 
 class LoginView(View, ClickableMixin):
-    username = TextInput(id='login_login')
-    password = TextInput(id='login_password')
+    username = TextInput(id="login_login")
+    password = TextInput(id="login_password")
+    login_text = Text('//p[@id="login_text"]')
+    logo = Text('//img[@alt="logo"]')
     submit = Text('//button[@type="submit"]')
 
     @property
     def is_displayed(self):
-        return self.browser.wait_for_element(
-            self.username, exception=False) is not None
+        return self.browser.wait_for_element(self.username, exception=False) is not None
--- a/airgun/entities/login.py
+++ b/airgun/entities/login.py
@@ -1,18 +1,23 @@
 from airgun.entities.base import BaseEntity
 from airgun.navigation import NavigateStep
 from airgun.navigation import navigator
+from airgun.views.common import BaseLoggedInView
 from airgun.views.login import LoginView
 
 
 class LoginEntity(BaseEntity):
-
     def login(self, values):
-        view = self.navigate_to(self, 'NavigateToLogin')
+        view = self.navigate_to(self, "NavigateToLogin")
         view.fill(values)
         view.submit.click()
 
     def logout(self):
-        pass
+        view = BaseLoggedInView(self.browser)
+        view.select_logout()
+        view.flash.assert_no_error()
+        view.flash.dismiss()
+        view = LoginView(self.browser)
+        return view.read()
 
 
 @navigator.register(LoginEntity)


devendra104:
@mirekdlugosz please review it.

airgun/session.py
```diff
@@ -217,7 +217,7 @@ def __exit__(self, exc_type, exc_value, traceback):
         passed = True if exc_type is None else False
         try:
             if passed:
-                self.login.logout()
+                pass
             else:
                 self.take_screenshot()
```
mirekdlugosz, line 222:
```python
try:
    if not passed:
        self.take_screenshot()
except Exception as err:
```

Also, can you add issue with link to this change that says "we used to call `logout()` at the end of each test (which did nothing, as logout was not implemented), but since commit foobar we no longer do. See also discussion in #397 . Let's continue discussion here - should we call logout on end of each test?".

omkarkhatavkar: I think we should not call the logout for each test, but we should have log out functionality as in because there are test scenarios need to verify that.

airgun/views/login.py
```diff
@@ -5,11 +5,12 @@
 
 
 class LoginView(View, ClickableMixin):
-    username = TextInput(id='login_login')
-    password = TextInput(id='login_password')
+    username = TextInput(id="login_login")
+    password = TextInput(id="login_password")
+    login_text = Text('//p[@id="login_text"]')
+    logo = Text('//img[@alt="logo')
```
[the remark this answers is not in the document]
devendra104: Typo, Thanks Mirek.

airgun/views/login.py
```diff
@@ -5,11 +5,12 @@
 
 
 class LoginView(View, ClickableMixin):
-    username = TextInput(id='login_login')
-    password = TextInput(id='login_password')
+    username = TextInput(id="login_login")
+    password = TextInput(id="login_password")
+    login_text = Text('//p[@id="login_text"]')
```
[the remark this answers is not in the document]
devendra104: The problem will come only if the tester sets the page footer text "null", and try to access the value of login_text.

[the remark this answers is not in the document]
mirekdlugosz: null is default, and I'm thinking in context of calling `logout` at the end of each test.

Anyway, I just checked and it turns out that if element is not found, `login_text` will not be present in returned dict. Which is OK, I guess.

airgun/session.py
devendra104: @mirekdlugosz I agree with omkar, It is not required to call logout in each test case if we add it, the execution time has drastically changed and that will not be a good sign for us.
As we know the current logout method did nothing it just calls and passes the logout without performing the actual logout. 
I think we can call the logout method when it is actually required and this way, we can achieve what we want without impacting the performance.
What do you say?

mirekdlugosz: Guys, I have not asked on opinion on logout on end of each test. I have asked to create issue where such discussion could take place. So this way we capture the initial intent of Airgun authors and can come up with more data when timing is better. I do not want to block this PR on that discussion and decision.

I have now created that issue: #475

omkarkhatavkar: As you mentioned **Let's continue the discussion here - should we call logout on the end of each test?".** I personally thought you wanted to discuss on this same thread.

Issue #476, opened 2020-04-07
Add support for email functionality verification using SMTP/Sendmail delivery method.


devendra104:
Planning to add the support of email functionality verification for both delivery method 
- Sendmail
- SMTP

To provide the support of send email and email notification verification, we will require to update the settings entities and their view module.

**Locator**
locator for email_button
```
email_button = Text('#test_mail_button')
notification_locator = Text('//*[@id="toast-notifications-container"]/div/div')
```
**method:**
To read the contents of settingsview, we will require a separate method 
 ```
def contents(self):
        view = SettingsView(self.browser)
        return view

```

mshriver:
@devendra104 has this issue been resolved by the above linked PRs?

devendra104:
@mshriver yes it has been resolved.

PR #477, opened 2020-04-08
Added "Test Mail" support for Sendmail/SMTP delivery method.

devendra104:
#476

### DIFF
--- a/airgun/entities/settings.py
+++ b/airgun/entities/settings.py
@@ -26,6 +26,13 @@ class SettingsEntity(BaseEntity):
         view.validations.assert_no_errors()
         view.wait_for_update()
 
+    def send_test_mail(self, property_name):
+        """Send the mail to the recipient"""
+        view = self.navigate_to(self, 'All')
+        view.search(property_name)
+        view.Email.test_email_button.click()
+        return view.flash.read()
+
 
 @navigator.register(SettingsEntity, 'All')
 class ShowAllSettings(NavigateStep):
--- a/airgun/views/settings.py
+++ b/airgun/views/settings.py
@@ -1,8 +1,10 @@
 from wait_for import wait_for
 from widgetastic.widget import Table
 from widgetastic.widget import Text
+from widgetastic_patternfly import Button
 
 from airgun.views.common import BaseLoggedInView
+from airgun.views.common import SatTab
 from airgun.views.common import SearchableViewMixin
 from airgun.widgets import PopOverWidget
 
@@ -15,6 +17,10 @@ class SettingsView(BaseLoggedInView, SearchableViewMixin):
             'Value': PopOverWidget()
         },
     )
+
+    @SatTab.nested
+    class Email(SatTab):
+        test_email_button = Button(id="test_mail_button")
 
     @property
     def is_displayed(self):


airgun/views/settings.py
```diff
@@ -9,6 +9,8 @@
 
 class SettingsView(BaseLoggedInView, SearchableViewMixin):
     title = Text("//h1[text()='Settings']")
+    email_button = Text('#test_mail_button')
```
omkarkhatavkar, line 12: Use the Button widget, try to use XPath instead of CSS `#test_mail_button` for uniform purpose.

airgun/views/settings.py
```diff
@@ -9,6 +9,8 @@
...
+    notification_locator = Text('//*[@id="toast-notifications-container"]/div/div')
```
omkarkhatavkar, line 13: try not to use the * instead that use the HTML tag

airgun/entities/settings.py
```diff
@@ -26,6 +26,11 @@ def update(self, property_name, value):
         view.validations.assert_no_errors()
         view.wait_for_update()
 
+    def contents(self):
+        """ Read the contents of setting browser"""
+        view = SettingsView(self.browser)
+        return view
+
```
omkarkhatavkar, line 33: In Airgun we have a strategy that all lower-level activity is done in Airgun, In that case you could use the function as send_email instead return the view.

airgun/views/settings.py
devendra104: You are right, the Button widget is the right way to do this. I will prefer id instead of XPath here.

airgun/views/settings.py
devendra104: Fixed

airgun/entities/settings.py
devendra104: Agree, But to check the status of sent email notificaton, contents of view is required.
  ```
  def sendmail(self):
        """ Read the contents of setting browser"""
        view = SettingsView(self.browser)
        view.email_button.click()
        return view
```

Comment (omkarkhatavkar) [reply airgun/entities/settings.py:34]: that's fine

File: airgun/views/settings.py [hunk 4]
```diff
@@ -9,6 +10,8 @@
 
 class SettingsView(BaseLoggedInView, SearchableViewMixin):
     title = Text("//h1[text()='Settings']")
+    email_button = Button(id="test_mail_button")
```
mirekdlugosz, line 13: This button is available only under "Email" tab, so you can't put it in main class variable - the way you have it, it won't be accessible if any other tab is selected.

Something like that should do the trick:
```python
class SettingsView(BaseLoggedInView, SearchableViewMixin):

    @View.nested
    class email(SatTab):
        test_button = Button(id="test_mail_button")
```

In entity, you only need `view.email.test_button` - `SatTab` has magic to select tab before child widget is interacted with (see [widgetastic_patternfly](https://github.com/RedHatQE/widgetastic.patternfly/blob/master/src/widgetastic_patternfly/__init__.py#L602) for details).

devendra104: Fixed.

airgun/views/settings.py
```diff
@@ -9,6 +10,8 @@
...
+    notification_locator = Text('/html/body/div[2]/div/div/div[1]/div/div')
```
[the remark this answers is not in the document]
devendra104: Yes, the Xpath of flash alert is autogenerated, we have not required to mentioned it separately.

airgun/entities/settings.py
```diff
@@ -26,6 +26,12 @@ def update(self, property_name, value):
         view.validations.assert_no_errors()
         view.wait_for_update()
 
+    def sendmail(self):
```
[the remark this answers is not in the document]
devendra104: Fixed.

airgun/views/settings.py
```diff
@@ -1,6 +1,8 @@
 from wait_for import wait_for
 from widgetastic.widget import Table
 from widgetastic.widget import Text
+from widgetastic_patternfly import Button
+from widgetastic_patternfly import View
```
mirekdlugosz, line 5: This way it caused Travis to fail when building docs.
Import from `widgetastic.widget` directly:

```suggestion
from widgetastic.widget import View
```

airgun/views/settings.py
```diff
@@ -16,6 +18,10 @@ class SettingsView(BaseLoggedInView, SearchableViewMixin):
         },
     )
 
+    @View.nested
+    class EmailView(View):
```
mirekdlugosz, line 22: You need to inherit from SatTab to have this auto-selecting magic to work:
```suggestion
    class EmailView(SatTab):
```

Import from `from airgun.views.common`.

devendra104: Fixed

devendra104:
> Two small corrections required.
> What about navigation ([#477 (comment)](https://github.com/SatelliteQE/airgun/pull/477#discussion_r406814044))? All you need to do is adding `view = self.navigate_to(self, 'All')`.
> 
> Also, I will appreciate current test result once all changes here are done.

Thanks Mirek for your valuable comment, I incorporated your suggestion and executed the UI test case with that changes and it is working fine.

Test Results
```
$ pytest tests/foreman/ui/test_settings.py::test_positive_update_email_delivery_method_sendmail
=============================================================================================== test session starts ===============================================================================================
platform linux -- Python 3.6.6, pytest-4.6.3, py-1.8.1, pluggy-0.13.1
collected 1 item                                                                                                                                                                                                  

tests/foreman/ui/test_settings.py .                                                                                                                                                                         [100%]
===================================================================================== 1 passedn 157.32 seconds ======================================================================================

```

airgun/views/settings.py
```diff
@@ -16,6 +18,10 @@ class SettingsView(BaseLoggedInView, SearchableViewMixin):
         },
     )
 
+    @SatTab.nested
+    class Email(SatTab):
```
mirekdlugosz, line 22: I am genuinely puzzled by you resistance to implement the exact version I provided in https://github.com/SatelliteQE/airgun/pull/477#discussion_r406824270 .
Decorator is `@View.nested`. Class should inherit from `SatTab`.

I am going to let it slip this time. After all, SatTab does inherit from View (eventually), and this seems to work. But most importantly, we need to branch off 6.7.z and this is only PR here that I am waiting for.

But in the future, please, pay attention to details like that. Nothing bad will happen if you incorporate changes after hour instead of 15 minutes. But every time a change is requested, it adds a day to entire process.

PR #479, opened 2020-04-17
submit is too fast, causing bug in automation

tstrych:
Out automation is too fast. Zalenium record shows, that automation clicks on the right place, and also upgrades automation passes, just normal automation don't. I was unable to reproduce it, but this should solve the problem. 

Please let me try this, I know it is unusual approach, but if it will not work i can remove it after one snap. 

This is maybe a bug in satellite, but no one ever will be that fast, or will not mind to click submit once more.

Note and quetion: 
Not sure if I should use sleep ? 
And if this should go to z-stream ? I think if this patch will work it should

### DIFF
--- a/airgun/entities/host.py
+++ b/airgun/entities/host.py
@@ -1,3 +1,5 @@
+from time import sleep
+
 from navmazing import NavigateToSibling
 from wait_for import wait_for
 
@@ -179,6 +181,7 @@ class HostEntity(BaseEntity):
            using the checkbox from table header
         """
         view = self._select_action('Delete Hosts', entities_list)
+        sleep(1)
         view.submit.click()
         view.flash.assert_no_error()
         view.flash.dismiss()


tstrych:
@mirekdlugosz Yes I can commit myself. I will remove it after few snaps. 

But if this solution helps, the code will still smell, so it's a signed we should create smth for these kind of problems. (There is still possibility that this is a bug but I think no developer would mind to fix it).

vijay8451:
@tstrych I do not mind to ack/merge this change  ... however, I would like to give another insight here which would help to debug this issue as per your way and that is `satellite6-standalone-automation` Jenkins job ...
So did you give a try there (as you could pass branch as per your way robbottelo+airgun too)..

tstrych:
I will give a shot to what Vijay propsed here. Till there please do not merge :)

tstrych:
```
============= 2 failed, 7 passed, 54 deselected in 1311.64 seconds =============
```

tstrych:
I run only `bulk delete host ` in `satellite6-standalone-automation/1927/`
And I run all ui tests in: `satellite6-standalone-automation/1929/`

In both of these I was unable to reproduce an error.

If you want I can try all tier4 tests or all tests. But, I think we should merge it now.

mirekdlugosz:
@tstrych Have you had a chance to verify this workaround works and affected tests do not fail anymore?

tstrych:
Today I come back to this as I saw the problem still occurs with `6.7.z ` branch. Becuase this PR was never cherry-picked there. 

This is not failing any more in 6.8, but it's still failing in 6.7 as it was not cherry-picked.

Based on report portal data:
This issue occurs first time at `6.7.0-16.0` , run 198 in RP, (not quite sure about that, as before this date there was also error caused by another problem)

One think I am not sure about is: 
if this PR resolved the issue OR (the issue itself no longer appears or was resolved in 6.8).

PR #480, opened 2020-04-21
custom pagination for tasks page

pondrejk:
Custom paginaton needed due to changed button id. 

Will need a cherry-pick once acked

### DIFF
--- a/airgun/views/task.py
+++ b/airgun/views/task.py
@@ -22,6 +22,10 @@ class TaskReadOnlyEntry(ReadOnlyEntry):
     )
 
 
+class TaskPagination(Pagination):
+    PER_PAGE_BUTTON_DROPDOWN = ".//div[button[@id='tasks-table-dropdown']]"
+
+
 class TasksView(BaseLoggedInView, SearchableViewMixin):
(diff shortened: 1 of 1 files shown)


airgun/views/task.py
```diff
@@ -22,6 +22,10 @@ class TaskReadOnlyEntry(ReadOnlyEntry):
     )
 
 
+class CustomPagination(Pagination):
```
omkarkhatavkar, line 25: Instead of `CustomPagination` in, we can use TaskPagination as a name.

airgun/views/task.py
```diff
@@ -22,6 +22,10 @@ class TaskReadOnlyEntry(ReadOnlyEntry):
     )
 
 
+class TaskPagination(Pagination):
+    PER_PAGE_BUTTON_DROPDOWN = ".//div[button[@id='tasks-table-dropdown']]"
```
mirekdlugosz, line 26: Note to self: once we start seeing more of changes like that, we should see if we can't just change Pagination to:
```python
    PER_PAGE_BUTTON_DROPDOWN = ".//div[button[contains(@id, '-dropdown')]]"
```

PR #484, opened 2020-05-08
Add support for widgetastic.patternfly >=1.3.0 (fixes #456)

mirekdlugosz:
widgetastic.patternfly 1.3.0 broke backwards compatibility in flash messages. See #456 and https://github.com/SatelliteQE/robottelo/pull/7646#pullrequestreview-361973923 .

Main changes in widgetastic.patternfly:
* Class to use when instantiating single message is stored as `FlashMessages` class variable, which means we can drop override for `messages`. It also requires us to define `SatFlashMessage` before `SatFlashMessages`.
* `SatFlashMessages.messages` is no longer a property, but a method - must be called
* `SatFlashMessages.messages()` is a generator instead of returning list
* `SatFlashMessages` no longer takes locator as argument during instantiation

We will have to merge https://github.com/SatelliteQE/robottelo/pull/7678 around the same time as this one, as Airgun will no longer work with widgetastic.patternfly pre-1.3.0.

Tests (see https://github.com/SatelliteQE/airgun/issues/456#issuecomment-625869518 for reasoning):
```
$ pytest -v tests/foreman/ui/test_activationkey.py::test_positive_create_with_host_collection tests/foreman/ui/test_contentview.py::test_positive_publish_promote_with_custom_puppet_module tests/foreman/ui/test_settings.py::test_positive_update_email_delivery_method_sendmail tests/foreman/ui/test_subscription.py::test_positive_end_to_end
===================================================================================== test session starts =====================================================================================
platform linux -- Python 3.7.5, pytest-4.6.3, py-1.8.1, pluggy-0.13.1 -- /home/mzalewsk/.virtualenvs/robottelo/bin/python
cachedir: .pytest_cache
shared_function enabled - OFF - scope:  - storage: file
rootdir: /home/mzalewsk/sources/robottelo
plugins: xdist-1.31.0, cov-2.8.1, services-1.3.1, forked-1.1.3, mock-1.10.4
collected 4 items                                                                                                                                                                             

tests/foreman/ui/test_activationkey.py::test_positive_create_with_host_collection PASSED                                                                                                [ 25%]
tests/foreman/ui/test_contentview.py::test_positive_publish_promote_with_custom_puppet_module PASSED                                                                                    [ 50%]
tests/foreman/ui/test_settings.py::test_positive_update_email_delivery_method_sendmail PASSED                                                                                           [ 75%]
tests/foreman/ui/test_subscription.py::test_positive_end_to_end PASSED                                                                                                                  [100%]
[...]

PR #486, opened 2020-05-12
flake8 complained about f-string without placeholders

lhellebr:


mirekdlugosz:
That's silly and makes me wonder why travis did not complain in https://github.com/SatelliteQE/airgun/pull/484
ACK

PR #495, opened 2020-05-26
[6.7.z] Pin older flake8 to avoid travis failures

mirekdlugosz:
Our pre-commit config used unpinned flake8 dependency. flake8 recently introduced new version with updated style guide that reports syntax that used to pass without problems.

We could fix that error, but our strategy is to freeze older branches in time, so it makes more sense for me to pin flake8 dependency to known version.

Test results - see travis

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,6 +5,10 @@ repos:
   rev: v1.9.0
   hooks:
   - id: reorder-python-imports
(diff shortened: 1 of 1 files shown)


mirekdlugosz:
I imagine newer flake is more strict than older one, so code that passed in master (on newer flake) should also pass on 6.7.z (on older flake).
But I don't really know, I guess we will see.

PR #502, opened 2020-07-17
fix in job template creation

pondrejk:
Updated breadcrumb and added wait for dialog to appear.

### DIFF
--- a/airgun/entities/job_template.py
+++ b/airgun/entities/job_template.py
@@ -1,4 +1,5 @@
 from navmazing import NavigateToSibling
+from wait_for import wait_for
 
 from airgun.entities.base import BaseEntity
 from airgun.navigation import NavigateStep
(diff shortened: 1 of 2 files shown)


airgun/entities/job_template.py
```diff
@@ -14,6 +15,11 @@ class JobTemplateEntity(BaseEntity):
     def create(self, values):
         """Create new job template"""
         view = self.navigate_to(self, 'New')
+        wait_for(
+            lambda: JobTemplateCreateView(self.browser).is_displayed is True,
```
mirekdlugosz, line 19: Is this `wait_for` necessary?
Our implementation of `am_i_here` calls view `is_displayed` method, so calling it explicitly shouldn't be required in common cases.

pondrejk: This solved the problem in automation, where the test tried to fill in the form before it was displayed, not sure why it's happening in this case.

PR #503, opened 2020-08-05
added support for two-factor rh-sso auth login

omkarkhatavkar:
1. We need to have this url after taking hammer auth login --two- factor  
```

[root@dhcp-2-190 ~]# echo 'df64bf48-453f-4755-8eb1-ba4d5a6a13c7.f89cccec-f022-4ef5-b06b-c2eaa15f799e.77bad924-e252-4daa-a7c9-9697d05339eb' | hammer auth login oauth --oidc-token-endpoint https://rhss.example.com/auth/realms/satqe/protocol/openid-connect/token --oidc-authorization-endpoint https://rhssoexample.com/auth/realms/satqe/protocol/openid-connect/auth --oidc-client-id satelliteexample.com-foreman-openidc --oidc-redirect-uri 'urn:ietf:wg:oauth:2.0:oob' --two-factor 
Enter URL in browser: https://satellite.example.com/auth/realms/satqe/protocol/openid-connect/auth?response_type=code&client_id=satellite.example.com-foreman-openidc&redirect_uri=urn:ietf:wg:oauth:2.0:oob&scope=openid

```
RHSSO_Login_1

2. This is the view after the login needs to be collected the token for further automation in hammer cli.  
Selection_043

### DIFF
--- a/airgun/entities/rhsso_login.py
+++ b/airgun/entities/rhsso_login.py
@@ -4,6 +4,7 @@ from airgun.navigation import navigator
 from airgun.views.common import BaseLoggedInView
 from airgun.views.rhsso_login import RhssoExternalLogoutView
 from airgun.views.rhsso_login import RhssoLoginView
+from airgun.views.rhsso_login import RhssoTwoFactorSuccessView
 
 
 class RHSSOLoginEntity(BaseEntity):
@@ -23,6 +24,12 @@ class RHSSOLoginEntity(BaseEntity):
         view.flash.assert_no_error()
         view.flash.dismiss()
         view = RhssoExternalLogoutView(self.browser)
+        return view.read()
+
+    def get_two_factor_login_code(self, values, url):
+        self.browser.selenium.get(url)
+        self.login(values)
+        view = RhssoTwoFactorSuccessView(self.browser)
         return view.read()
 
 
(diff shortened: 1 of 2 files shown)


airgun/entities/rhsso_login.py
```diff
@@ -25,6 +26,12 @@ def logout(self):
         view = RhssoExternalLogoutView(self.browser)
         return view.read()
 
+    def get_two_factor_login_code(self, values, url):
+        self.browser.selenium.get(url)
+        self.login(values)
+        view = RhssoTwoFactorSuccessView(self.browser)
```
mirekdlugosz, line 32: Could you explain what exactly is happening here?

Why does test need to pass URL? What this URL is?
`self.login` will change view for us. And then we say we are on specific view, anyway?

omkarkhatavkar: I missed adding the description, Yes the URL different from satellite url, and there is no navigation to that url from the satellite (mentioned in the description). Yes, we have a new view after login.

mirekdlugosz: Thanks, makes sense now.

lhellebr (commented):
Is this actually used in some test? Can you show the results?

mirekdlugosz:
@lhellebr https://github.com/SatelliteQE/robottelo/pull/7925

lhellebr (approved):
LGTM

Issue #507, opened 2020-08-26
Sync Status page does not display correctly unless refreshed


swadeley:
Hello

Sync Status page does not display correctly unless refreshed. The clickable expander triangles aren't visible and test[1] fails when trying to open  them.

This issue affect 6.7.2 and 6.7.3 snap 1 but not 6.8 snap 12

![image](https://user-images.githubusercontent.com/11247237/91327003-e92b0500-e7bc-11ea-9b17-ac44bf7a8e26.png)

The test I am running[1] fails here:
  
    session.organization.select(org_name=module_org_with_manifest.name)
      >           results = session.sync_status.synchronize(repo_paths)

    tests/foreman/ui/test_sync.py:110:

underlying error is here:

       >               'Adding node to a non section one is prohibited')
     E           airgun.views.sync_status.ReservedToSectionOnlyError: Adding
     node to a non section one is prohibited

     ./airgun/views/sync_status.py:133:
     ReservedToSectionOnlyError

Adding view.browser.refresh() fixes the issue:

    airgun/entities/sync_status.py:

        view = self.navigate_to(self, 'All')
        view.browser.refresh()
        repo_nodes = [view.table.get_node_from_path(repo_path)
                      for repo_path in repository_paths]

Thanks to Tasos Papaioannou for the suggestion.

Using Chromium Version 84.0.4147.89 (Developer Build) Fedora Project (64-bit)

[1] https://github.com/SatelliteQE/robottelo/blob/abf74da078eaacbecc7a2e8e5e9a1e809eeaafc4/tests/foreman/ui/test_sync.py#L82

mirekdlugosz:
Isn't that regression that should be also fixed downstream?

I'm fine with workarounds being present until product fix their bugs, but we shouldn't keep them much longer than necessary. Having bz number in github issue and commit message will hopefully make it a little clearer this is something we can revert eventually.

swadeley:
> Isn't that regression that should be also fixed downstream?
> 
> I'm fine with workarounds being present until product fix their bugs, but we shouldn't keep them much longer than necessary. Having bz number in github issue and commit message will hopefully make it a little clearer this is something we can revert eventually.

Unfortunately we cannot reproduce the error except by running the test.  So there is no evidence a user would encounter this and a bug without a reproducer, other than our tests, is not good. I did ask devs to consider what might cause this. 

Now that we know Sat6.8 is not affected its seems to be the opposite of a regression to me.

Thank you

[...]

PR #509, opened 2020-08-26
Fix Sync Status page partial loading (#507)

swadeley:
Hello
For 6.7 branch

this fixes issue :
Sync Status page does not display correctly unless refreshed #507


mirekdlugosz (approved):
Thanks for explanations in #507 . ACK.

omkarkhatavkar (approved):
ACK, sorry for the late response from my end as I was PTO and enjoying some holidays

PR #510, opened 2020-08-28
update the logic for install package using katello-agent

omkarkhatavkar:
### PR Description

- When we install the package using Katello-Agent then Conte Host Task Details page populated, this view needs to be associated with action function. 
- Remove not required logic 

### Test Result 
Please see the PR https://github.com/SatelliteQE/robottelo/pull/7947

### DIFF
--- a/airgun/entities/contenthost.py
+++ b/airgun/entities/contenthost.py
@@ -124,13 +124,13 @@ class ContentHostEntity(BaseEntity):
             'rex': 'via remote execution',
             'rex_customize': 'via remote execution - customize first',
         }
-        if install_via is not None:
-            view.errata.apply_selected.fill(install_via_dict[install_via])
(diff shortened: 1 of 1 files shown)


mirekdlugosz (approved):
Seems with this test, we lose ability to click "Apply selected". Which would pop up a dialog and apply default method, I guess?
Since your test is the only one that actually uses this method, I'm not particularly concerned about this. 

ACK

omkarkhatavkar:
> I'm not particularly concerned about this.

You are correct are going to lose that functionality, seems there is not test which uses that yet. Hence Just to make things simpler I removed. We should add it in case any test using the 'apply selected'

omkarkhatavkar:
Incase anyone looking for Travis failure it is unrelated to PR, this is an existing issue.

lhellebr (approved):
LGTM

PR #512, opened 2020-09-02
Change wait_manifest to a more reliable check

### LINKED ISSUE #511: has_manifest relies on an inconsistent condition
The has_manifest function determines the organization has a manifest by seeing if the "Add Subscriptions" button is enabled. This breaks in a recent change where a cloned manifest is uploaded, but the Add button is disabled.

The function should instead navigate to "manage manifest" and then check to see if the "delete" button is enabled. Alternatively, it can check to see if a url is present, linking to the manifest on the portal.

https://github.com/SatelliteQE/airgun/blob/9543d1ca4b7f6ac15c6388ac31647b3955d0e486/airgun/entities/subscription.py#L53

JacobCallahan:
This change now checks to see if the delete button is enabled.
This should remain a valid check, regardless of manifest type.

Fixes #511

### DIFF
--- a/airgun/entities/subscription.py
+++ b/airgun/entities/subscription.py
@@ -33,13 +33,13 @@ class SubscriptionEntity(BaseEntity):
                 handle_exception=True, timeout=timeout,
                 logger=view.progressbar.logger
         )
+        view.flash.assert_no_error(ignore_messages=ignore_error_messages)
+        view.flash.dismiss()
         wait_for(
                 lambda: self.has_manifest == has_manifest,
                 handle_exception=True, timeout=10,
                 logger=view.logger
         )
-        view.flash.assert_no_error(ignore_messages=ignore_error_messages)
-        view.flash.dismiss()
 
     @property
     def has_manifest(self):
@@ -48,12 +48,13 @@ class SubscriptionEntity(BaseEntity):
         May be None if user can't verify reliably if manifest is
         uploaded or not due to missing permissions
         """
-        view = self.navigate_to(self, 'All')
         try:
-            view.add_button.wait_displayed()
+            view = self.navigate_to(self, 'Manage Manifest')
+            result = "http" in view.manifest.red_hat_cdn_url.value
+            view.close_button.click()
+            return result
         except TimedOutError:
             return None
-        return not view.add_button.disabled
 
     def add_manifest(self, manifest_file, ignore_error_messages=None):
         """Upload manifest file


airgun/entities/subscription.py
```diff
@@ -48,12 +48,12 @@ def has_manifest(self):
         May be None if user can't verify reliably if manifest is
         uploaded or not due to missing permissions
         """
-        view = self.navigate_to(self, 'All')
+        view = ManageManifestView(self.browser)
```
mirekdlugosz, line 51: I don't think this will work. To upload manifest in UI, you need to click "Manage Manifest" button, which opens modal window for file upload. When file is selected, Satellite automatically closes modal. And this line would work only if modal is already displayed in browser.

`navigate_to` method is here to ensure that we are on view that we expect to be. If you skip `navigate_to`, you need to ensure browser state through some other mean.

But then, if you o use `navigate_to`, pay close attention to state that you leave Satellite in. Most of navigation code tries to interact with menu, never checking if that's a thing they actually can do. If you leave modal open, I expect that background overlay will cover menu, and as result cause next navigation attempt to fail.

To sum up, what you *probably* need is navigation to modal at beginning of this method, and closing of modal at the end.

mirekdlugosz (commented):
This whole "check button status" was a big hack relying on undocumented side-effect, so it's not really surprising it stopped working. Unfortunately, it seems that there is still no easy and reliable way of obtaining manifest upload status through UI...

By convention, we require demonstration that proposed change can work, usually in form of some test result. `tests/foreman/ui/test_subscription.py::test_positive_end_to_end` should be good candidate.

`refresh_manifest` entity method is only used in `tests/foreman/ui/test_organization.py::test_positive_download_debug_cert_after_refresh`. Having test result for both of these would be perfect, but `test_subscription.py` is enough for me.

JacobCallahan:
@mirekdlugosz it looks like my latest version accurately checks the status. However, I'm not sure how to close the modal.

mirekdlugosz:
@JacobCallahan 
```python
view = self.navigate_to(self, 'Manage Manifest')
# check buttons visibility and everything else you need
view.close_button.click()
```

JacobCallahan:
@mirekdlugosz I've updated to close. An alternate version could look like this.

```python
        result = False
        view = self.navigate_to(self, 'Manage Manifest')
        try:
            view.wait_animation_end()
            result = "http" in view.manifest.red_hat_cdn_url.value
        except TimedOutError:
            pass
        finally:
            view.close_button.click()
            return result
```
I'm encountering another issue from a separate piece of the subscription.py file when trying to test this locally, but it seems like my change is passing. Can you try to test this locally?

JacobCallahan:
@mirekdlugosz thanks for that! I've updated the pr to include what I believe you did.

mirekdlugosz:
@JacobCallahan yes. Overall looks good, but we can't ACK until you move it out from draft state.
Did you manage to solve your other problem and can you add test run results here?

mirekdlugosz (approved):

latran (approved):
ACK, pending test result.

JacobCallahan:
Unfortunately, all the tests in test_subscription have been historically unstable. Even with my fix, the test I chose fails later down the line. However, the has_manifest check passed before this error. I'd recommend the component owner reevaluate these tests.

```
_____________ test_positive_access_manifest_as_another_admin_user ______________

test_name = 'tests.foreman.ui.test_subscription.test_positive_access_manifest_as_another_admin_user'

    @tier2
    def test_positive_access_manifest_as_another_admin_user(test_name):
        """Other admin users should be able to access and manage a manifest
        uploaded by a different admin.
    
        :id: 02e319da-3b7a-4694-9164-475c2c71b9a8
    
        :expectedresults: Other admin user should see/manage the manifest
    
        :BZ: 1669241
    
        :customerscenario: true
    
        :CaseLevel: Integration
    
        :CaseImportance: High
        """
        org = entities.Organization().create()
        user1_password = gen_string('alphanumeric')
        user1 = entities.User(
            admin=True, password=user1_password, organization=[org], default_organization=org
        ).create()
        user2_password = gen_string('alphanumeric')
        user2 = entities.User(
            admin=True, password=user2_password, organization=[org], default_organization=org
        ).create()
        # use the first admin to upload a manifest
        with Session(test_name, user=user1.login, password=user1_password) as session:
            manifests.upload_manifest_locked(org.id)
            assert session.subscription.has_manifest
            # store subscriptions that have "Red Hat" in the name for later
            rh_subs = session.subscription.search("Red Hat")
        # try to view and delete the manifest with another admin
        with Session(test_name, user=user2.login, password=user2_password) as session:
>           assert session.subscription.has_manifest

tests/foreman/ui/test_subscription.py:258: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../venvrobottelo/lib64/python3.8/site-packages/airgun/entities/subscription.py:54: in has_manifest
    result = "http" in view.manifest.red_hat_cdn_url.value
../venvrobottelo/lib64/python3.8/site-packages/widgetastic/widget/base.py:131: in __get__
    obj.child_widget_accessed(widget)
../venvrobottelo/lib64/python3.8/site-packages/widgetastic/widget/base.py:67: in wrapped
    return method(self, *new_args, **new_kwargs)
../venvrobottelo/lib64/python3.8/site-packages/widgetastic_patternfly/__init__.py:644: in child_widget_accessed
    self.select()
../venvrobottelo/lib64/python3.8/site-packages/widgetastic/widget/base.py:67: in wrapped
    return method(self, *new_args, **new_kwargs)
../venvrobottelo/lib64/python3.8/site-packages/widgetastic_patternfly/__init__.py:635: in select
    if not self.is_active():
../venvrobottelo/lib64/python3.8/site-packages/widgetastic/widget/base.py:67: in wrapped
    return method(self, *new_args, **new_kwargs)
../venvrobottelo/lib64/python3.8/site-packages/widgetastic_patternfly/__init__.py:622: in is_active
    return 'active' in self.parent_browser.classes(self.TAB_LOCATOR)
../venvrobottelo/lib64/python3.8/site-packages/widgetastic/utils.py:679: in wrap
../venvrobottelo/lib64/python3.8/site-packages/widgetastic/browser.py:605: in classes
    command, self.element(locator, *args, **kwargs),
../venvrobottelo/lib64/python3.8/site-packages/widgetastic/browser.py:341: in element
    elements = self.elements(locator, *args, **kwargs)
../venvrobottelo/lib64/python3.8/site-packages/widgetastic/browser.py:923: in elements
    return self._browser.elements(
../venvrobottelo/lib64/python3.8/site-packages/widgetastic/utils.py:679: in wrap
../venvrobottelo/lib64/python3.8/site-packages/widgetastic/browser.py:264: in elements
    root_element = self.element(parent, parent=parent.locatable_parent)
../venvrobottelo/lib64/python3.8/site-packages/widgetastic/browser.py:341: in element
    elements = self.elements(locator, *args, **kwargs)
../venvrobottelo/lib64/python3.8/site-packages/widgetastic/utils.py:679: in wrap
../venvrobottelo/lib64/python3.8/site-packages/widgetastic/browser.py:253: in elements
    locator = self._process_locator(locator)
../venvrobottelo/lib64/python3.8/site-packages/widgetastic/browser.py:202: in _process_locator
    return locator.__element__()
../venvrobottelo/lib64/python3.8/site-packages/widgetastic/widget/base.py:67: in wrapped
    return method(self, *new_args, **new_kwargs)
../venvrobottelo/lib64/python3.8/site-packages/widgetastic/widget/base.py:344: in __element__
    return self.parent_browser.element(locator)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <airgun.browser.AirgunBrowser object at 0x7f2ea7a82190>
locator = Locator(by='xpath', locator='//div[@role="dialog" and @tabindex][div//h4[text()="Manage Manifest"]]')
args = (), kwargs = {}, vcheck = None, elements = []

    def element(self, locator, *args, **kwargs):
        """Returns one :py:class:`selenium.webdriver.remote.webelement.WebElement`
    
        See: :py:meth:`elements`
    
        Returns:
            :py:class:`selenium.webdriver.remote.webelement.WebElement`
    
        Raises:
            :py:class:`selenium.common.exceptions.NoSuchElementException`
        """
        try:
            vcheck = self._locator_force_visibility_check(locator)
            if vcheck is not None:
                kwargs['check_visibility'] = vcheck
            elements = self.elements(locator, *args, **kwargs)
            if len(elements) > 1:
                visible_elements = [e for e in elements if self.is_displayed(e)]
                if visible_elements:
                    return visible_elements[0]
                else:
                    return elements[0]
            else:
                return elements[0]
        except IndexError:
>           raise NoSuchElementException(
                'Could not find an element {}'.format(repr(locator))) from None
E           selenium.common.exceptions.NoSuchElementException: Message: Could not find an element Locator(by='xpath', locator='//div[@role="dialog" and @tabindex][div//h4[text()="Manage Manifest"]]')
```

JacobCallahan:
@mirekdlugosz thoughts? it might be easier if you can test it locally

mirekdlugosz:
@JacobCallahan you have my ACK already.
@latran are you satisfied with Jake's answer? I guess it's up to you to decide if we are all good here and can merge, or there is something more we need to do.

ColeHiggins2:
@mirekdlugosz @JacobCallahan As the component owner of test_subscription, I ran the test locally with the new fix and got the same error. However, It still failed in the first "session.subscription.has_manifest" call.

After playing around with the new fix, I noticed something rather interesting and I think it relates to what @mirekdlugosz said above. If I put breakpoints on each .has_manifest assertion, wait a few seconds, and then step through, The assertion seems to pass for both user1 and user2. Maybe we need to look into the timeout-exceptions or "wait_for" method in airgun and allow for the satellite to process everything before we try to make these assertions? It still fails once we assert "not session.subscription.has_manifest" at the end, but I suspect this is because of the same issue. 

If we cannot fix this issue, I think an alternative to this failing assertion would be to import cli or api methods to assert that any subscription is there. I looked into "session.subscription" and it seems like the functionality to even check if subscriptions exist is not there and or is blocked by the error above. Would anyone be opposed to this?

ColeHiggins2:
Test output with breakpoints on first two ".has_manifest" assertions
```test_name = 'tests.foreman.ui.test_subscription.test_positive_access_manifest_as_another_admin_user'

    @tier2
    def test_positive_access_manifest_as_another_admin_user(test_name):
        """Other admin users should be able to access and manage a manifest
        uploaded by a different admin.
    
        :id: 02e319da-3b7a-4694-9164-475c2c71b9a8
    
        :expectedresults: Other admin user should see/manage the manifest
    
        :BZ: 1669241
    
        :customerscenario: true
    
        :CaseLevel: Integration
    
        :CaseImportance: High
        """
        org = entities.Organization().create()
        user1_password = gen_string('alphanumeric')
        user1 = entities.User(
            admin=True, password=user1_password, organization=[org], default_organization=org
        ).create()
        user2_password = gen_string('alphanumeric')
        user2 = entities.User(
            admin=True, password=user2_password, organization=[org], default_organization=org
        ).create()
        # use the first admin to upload a manifest
        with Session(test_name, user=user1.login, password=user1_password) as session:
            manifests.upload_manifest_locked(org.id)
            assert session.subscription.has_manifest
            # store subscriptions that have "Red Hat" in the name for later
            rh_subs = session.subscription.search("Red Hat")
        # try to view and delete the manifest with another admin
        with Session(test_name, user=user2.login, password=user2_password) as session:
            assert session.subscription.has_manifest
            assert rh_subs == session.subscription.search("Red Hat")
>           session.subscription.delete_manifest(ignore_error_messages=['404 Not Found'])

test_subscription.py:260: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../../../venv/lib/python3.7/site-packages/airgun/entities/subscription.py:89: in delete_manifest
    'Delete Manifest', has_manifest=False, ignore_error_messages=ignore_error_messages)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <airgun.entities.subscription.SubscriptionEntity object at 0x7f668fc01510>
name = 'Delete Manifest', has_manifest = False, timeout = 600
ignore_error_messages = ['404 Not Found']

    def _wait_for_process_to_finish(self, name, has_manifest=False, timeout=600,
                                    ignore_error_messages=None):
        """Helper ensuring that task (upload / delete manifest / subscription)
        has finished. Run after action invoking task to leave Satellite
        in usable state.
        Currently waits for three events. Since page is written asynchronously,
        they can happen in any order.
        :param name: Name of running task
        :param has_manifest: Should manifest exist after task ended?
        :param timeout: Waiting timeout
        :param ignore_error_messages: A List of strings representing the error messages to ignore.
        """
        view = SubscriptionListView(self.browser, logger=self.browser.logger)
        wait_for(
                lambda: not view.progressbar.is_displayed,
                handle_exception=True, timeout=timeout,
                logger=view.progressbar.logger
        )
        view.flash.assert_no_error(ignore_messages=ignore_error_messages)
        view.flash.dismiss()
        wait_for(
                lambda: self.has_manifest == has_manifest,
                handle_exception=True, timeout=10,
>               logger=view.logger
        )
E       wait_for.TimedOutError: Could not do 'lambda defined as `lambda: self.has_manifest == has_manifest,`' at /home/colehiggins/projects/venv/lib/python3.7/site-packages/airgun/entities/subscription.py:39 in time
```

JacobCallahan:
@mirekdlugosz so there could be some kind of timing issue he's encountering. Is there a way to handle this within airgun?

latran:
I'm getting failure at the first `assert session.subscription.has_manifest` like cole did and I'm unsure if there's a way to handle this timing issue in airgun.

tpapaioa:
There are a couple issues causing the test failures:
- When navigating to `Content >  Subscriptions`, the inline notifications that pop up can block the `Manage Manifest` button. In my testing, there is one asynchronous notification, which disappears after a few seconds unless clicked on, with the text `Now you can enable Red Hat inventory upload <Go to inventory upload settings>`. The other one is a synchronous notification that stays visible until dismissed, `404 Not Found`, and is associated with the fake CDN URL used in our manifest cloning procedure. In my testing, the following code resolves that issue by waiting until the async notification has disappeared, then dismissing the remaining notifications, ensuring that the `Manage Manifest` button can be clicked.

```
    def _wait_for_process_to_finish(self, name, has_manifest=False, timeout=600,
                                    ignore_error_messages=None):
[...]
        sleep(15)  # tpapaioa
        view.flash.assert_no_error(ignore_messages=ignore_error_messages)
        view.flash.dismiss()
        # tpapaioa increase timeout to 600
        wait_for(
                lambda: self.has_manifest == has_manifest,
                handle_exception=True, timeout=600,
                logger=view.logger
        )
```
(There's probably a better way to fix that.)
- The changes to `has_manifest()` aren't valid for determining whether a manifest is present in the organization. In my manual testing, the CDN URL can be present regardless of whether a manifest is present. In `ManageManifestView`, a more reliable test might be to check whether `manifest.delete_button` is clickable, and/or whether the text `robottelo-automation` (the name in our manifest) vs. `No Manifest Uploaded` appears next to the `USMA` label.

JacobCallahan:
@mirekdlugosz @mshriver do either of you have an idea on how to solve that?

airgun/entities/subscription.py
mshriver: We can actually bake this behavior into NavigateStep.pre_navigate, so that before any navigation is attempted there is a check whether the modal exists. In integration_tests use of navmazing, we actually recycled the browser instance when this happened, so that we weren't concerned with how to close the modal, we would just start navigation from scratch.

mshriver:
@tpapaioa instead of statically sleeping before dismissing flash messages, we could pass `fail_func=view.flash.dismiss` to wait_for, so that on any failure of the wait condition, it will try to clear flash messages.

@tpapaioa your second point there is the most important - we're looking at the wrong thing anyway in order to answer the `has_manifest` question. I wouldn't go as specific as checking for any manifest metadata related to robottelo or SatQE - but asserting against the presence of `No Manifest Uploaded` and whether manifest.delete_button is enabled is what we should be doing.

airgun/entities/subscription.py
```diff
@@ -48,12 +48,13 @@ def has_manifest(self):
         May be None if user can't verify reliably if manifest is
         uploaded or not due to missing permissions
         """
-        view = self.navigate_to(self, 'All')
         try:
-            view.add_button.wait_displayed()
+            view = self.navigate_to(self, 'Manage Manifest')
+            result = "http" in view.manifest.red_hat_cdn_url.value
+            view.close_button.click()
```
mshriver, line 54: Some concern about calling this close_button inside the try here, instead of in a `finally` block, what does it leave on the displayed page if there is an exception (especially other than a TimedOutError) when reading `red_hat_cdn_url.value`?

tpapaioa:
The changes in https://github.com/tpapaioa/airgun/tree/fix_has_manifest work in my testing.

JacobCallahan:
@tpapaioa i'm more than happy to close this in favor of your solution

JacobCallahan:
Closing in favor of #573

PR #515, opened 2020-09-08
Fixes #514 - Add support for oscap reports search and delete

### LINKED ISSUE #514: Add support for Oscap Compliance reports search and delete.

jameerpathan111:
Fixes #514 
This PR adds support for compliance reports search and delete operation.

### DIFF
--- a/airgun/entities/oscapreports.py
+++ b/airgun/entities/oscapreports.py
@@ -1,0 +1,76 @@
+from time import sleep
+
+from airgun.entities.base import BaseEntity
+from airgun.navigation import NavigateStep
+from airgun.navigation import navigator
+from airgun.views.oscapreports import DeletereportsDialog
+from airgun.views.oscapreports import OSCAPReportsView
+
+
+class OSCAPReportsEntity(BaseEntity):
+    endpoint_path = '/compliance/arf_reports'
+
+    def search(self, value):
+        """Search for existing Compliance Reports."""
+        view = self.navigate_to(self, 'All')
+        return view.search(value)
+
+    def read_all(self):
+        """Read all values from Compliance Reports page."""
+        view = self.navigate_to(self, 'All')
+        return view.read()
+
+    def delete(self, search_pattern, cancel=False):
+        """Delete first Compliance Report based on search_pattern."""
+        view = self.navigate_to(self, 'All')
+        view.search(search_pattern)
+        view.table.row()['Actions'].widget.fill('Delete')
+        alert_message = self.browser.get_alert().text
+        self.browser.handle_alert(cancel=cancel)
+        view.flash.assert_no_error()
+        view.flash.dismiss()
+        return alert_message
+
+    def full_report(self, search_pattern, cancel=False):
+        """View Full Compliance Report."""
+        view = self.navigate_to(self, 'All')
+        view.search(search_pattern)
+        view.table.row()['Actions'].widget.fill('Full Report')
+
+    def delete_reports(self, search_pattern):
+        """Delete all reports based on search_pattern
+           If keyword 'All' is supplied, all reports are selected
+           using the checkbox from table header
+        """
+        view = self.navigate_to(self, 'Delete reports', search_pattern=search_pattern)
+        view.submit.click()
+        view.flash.assert_no_error()
+        view.flash.dismiss()
+
+
+@navigator.register(OSCAPReportsEntity, 'All')
+class ShowAllHosts(NavigateStep):
+    """Navigate to Compliance Reports page."""
+    VIEW = OSCAPReportsView
+
+    def step(self, *args, **kwargs):
+        self.view.menu.select('Hosts', 'Reports')
+
+
+@navigator.register(OSCAPReportsEntity, 'Delete reports')
+class ReportsSelectAction(NavigateStep):
+    """Navigate to Delete reports page by selecting checkboxes for necessary reports.
+
+    Args:
+        search_pattern: search_pattern to select reports.
+    """
+    def prerequisite(self, *args, **kwargs):
+        return self.navigate_to(self.obj, 'All')
+
+    def step(self, *args, **kwargs):
+        search_pattern = kwargs.get('search_pattern')
+        self.VIEW = DeletereportsDialog
+        if search_pattern != "All":
+            self.parent.search(search_pattern)
+        self.parent.select_all.fill(True)
+        self.parent.delete_reports.click()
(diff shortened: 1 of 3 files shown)


jameerpathan111:
Travis failure seem unrelated.

airgun/entities/oscapreports.py
```diff
@@ -0,0 +1,84 @@
...
+    def full_report(self, search_pattern, cancel=False):
+        """View Full Compliance Report."""
+        view = self.navigate_to(self, 'All')
+        view.search(search_pattern)
+        view.table.row()['Actions'].widget.fill('Full Report')
+        alert_message = self.browser.get_alert().text
+        self.browser.handle_alert(cancel=cancel)
+        view.flash.assert_no_error()
+        view.flash.dismiss()
+        return alert_message
```
mirekdlugosz, line 43: Looks like copy-paste from `delete`. As far as I can tell, there is no alert and no flash messages when clicking "Full Report".

Also, you probably want to return *something* from this report, not alert text.

airgun/entities/oscapreports.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def delete_reports(self, search_pattern):
+        """Delete all reports based on search_pattern
+           If keyword 'All' is supplied, all reports are selected
+           using the checkbox from table header
+        """
+        view = self.navigate_to(self, 'Delete reports', search_pattern=search_pattern)
+        sleep(1)
```
mirekdlugosz, line 51: Is this absolutely necessary? Can't we use something more explicit instead (what exactly we are waiting for)?

Also, that wait probably belongs to navigation endpoint `post_navigate` method. There are couple of usages in Airgun if you want to see how it could be implemented.

airgun/entities/oscapreports.py
```diff
@@ -0,0 +1,84 @@
...
+        return self.navigate_to(self.obj, 'All')
+
+    def step(self, *args, **kwargs):
+        search_pattern = kwargs.get('search_pattern')
+        self.VIEW = DeletereportsDialog
+        if search_pattern == "All":
+            self.parent.select_all.fill(True)
+        else:
+            self.parent.search(search_pattern)
+            self.parent.select_all.fill(True)
```
mirekdlugosz, line 83: You can avoid duplication by reversing conditional:
```suggestion
        if search_pattern != "All":
            self.parent.search(search_pattern)
        self.parent.select_all.fill(True)
```

airgun/views/oscapreports.py
```diff
@@ -0,0 +1,47 @@
...
+        column_widgets={
+            0: Checkbox(
+                locator=".//input[@class='host_select_boxes']"),
+            'Host': Text('./a'),
+            'Reported At': Text('./a'),
+            'Policy': Text('./a'),
+            'Openscap Capsule': Text('./a'),
+            'Passed': Text('./a'),
+            'Failed': Text('./a'),
+            'Other': Text('./a'),
```
mirekdlugosz, line 25: FYI, you don't need to model columns that you don't plan to interact with.
AFAIR widgetastic can still read all columns, also by name, and will assume `Text()` when column is not declared explicitly. So you can still read values in your test.

airgun/views/oscapreports.py
```diff
@@ -0,0 +1,47 @@
...
+from widgetastic.widget import Text
+
+from airgun.views.common import BaseLoggedInView
+from airgun.views.common import SearchableViewMixin
+from airgun.widgets import ActionsDropdown
+
+
+class OSCAPReportsView(BaseLoggedInView, SearchableViewMixin):
+    title = Text("//h1[text()='Compliance Reports']")
+    delete_reports = Text("//div[@id='submit_multiple']")
```
mirekdlugosz, line 12: Technically, this is button group and not button itself. But right now this group contains only one member.
That theoretically matters, as button group dimensions on page are slightly bigger than button dimensions. So, theoretically, driver could click on area that is not button and not trigger action.

Also, buttons can be modeled by `Button` from `widgetastic_patternfly`.

Practically, neither of these really matters.

airgun/views/oscapreports.py
```diff
@@ -0,0 +1,47 @@
...
+        return self.browser.wait_for_element(
+            self.title, exception=False) is not None
+
+
+class DeletereportsDialog(BaseLoggedInView):
+    """Class for Oscap reports Delete Action."""
+    title = Text(
+        "//h4[text()='Delete reports - The following compliance reports are about to be changed']")
+    table = Table("//div[@class='modal-body']//table")
+    keep_selected = Checkbox(id='keep_selected')
```
mirekdlugosz, line 41: Is that correct? I couldn't find this element on page when modal was opened.

mirekdlugosz (changes requested):
`full_report` method doesn't seem to do what it was supposed to be doing.
Other comments are questions/suggestions, not blocking.

airgun/entities/oscapreports.py
jameerpathan111: @mirekdlugosz thanks for your comments :) Fixed few issue you mentioned.

jameerpathan111: I was thinking of covering `full report` method in other PR, but I guess I should add it in this one itself so I will move this PR to draft state till I finish it.

airgun/entities/oscapreports.py
jameerpathan111: No, this is not necessary. removed it.

airgun/entities/oscapreports.py
jameerpathan111: updated pr.

airgun/views/oscapreports.py
jameerpathan111: ahh thanks for explaining, just kept those columns which I think I would need to interact with.

airgun/views/oscapreports.py
jameerpathan111: yes, this one also not needed.

airgun/entities/oscapreports.py
mirekdlugosz: @jameerpathan111 depends on you, really. If you can get some value out of this PR without `full_report` being implemented, I won't block it. But `full_report` method should raise NotImplementedError exception, or not exist at all (yet).

If there is nothing we could use without `full_report`, then yes, drafting and finishing later sounds like a good decision.

airgun/entities/oscapreports.py
```diff
@@ -0,0 +1,76 @@
+from time import sleep
```
mshriver, line 1: This import should be removed, the use was removed.

mshriver:
@jameerpathan111 please re-visit this branch to ensure @mirekdlugosz latest comments have been addressed, and resolve the failure in CI.

jameerpathan111:
Closing this pr for now, will re-open once I will start work on this again.

PR #516, opened 2020-09-08
wait until user create button is clicked

pondrejk:
Test User UI automation occasionally failed on expecting to see the user creation dialog before the create user button was actually clicked


jameerpathan111 (approved):
ACK

mirekdlugosz (approved):
I think this is already done during navigation and before click, but if that helps
ACK

PR #517, opened 2020-09-08
fix cockpit test, cockpit now opens in new tab

tstrych:
Cockpit test is now opening in new tab by default. This PR implement that. 
I am open to different ideas how to implement this. This is first time I am switching the tabs. 

I've hit an logging error in test result (you can find at at the bottom of test result) 

Test results with loggs:
```
/automation/tests/foreman/ui/test_host.py -- -k test_positive_cockpit
collected 30 items / 29 deselected / 1 selected

test_host.py::test_positive_cockpit 
=================== 1 passed, 29 deselected in 75.07 seconds ===================

### DIFF
--- a/airgun/entities/host.py
+++ b/airgun/entities/host.py
@@ -223,18 +223,24 @@ class HostEntity(BaseEntity):
         view = self.navigate_to(self, 'Details', entity_name=entity_name)
         view.webconsole.click()
         view.validations.assert_no_errors()
+        # switch to the last opened tab,
+        self.browser.switch_to_window(self.browser.window_handles[-1])
+        self.browser.wait_for_element(
+            locator='//div[@id="index-brand"]', exception=True
+        )
         # the remote host content is loaded in an iframe, let's switch to it
-        self.browser.selenium.switch_to.frame(0)
+        self.browser.switch_to_frame(locator='//div[@id="content"]/iframe')
+
+        self.browser.wait_for_element(
+            locator='//a[@id="system_information_hostname_button"]', exception=True, visible=True
+        )
         hostname_button_view = self.browser.selenium.find_elements_by_id(
             'system_information_hostname_button'
         )
-        wait_for(
-            lambda: hostname_button_view[0].text != '',
-            handle_exception=True, timeout=10,
-            logger=view.logger
-        )
         hostname = hostname_button_view[0].text
         self.browser.switch_to_main_frame()
+        self.browser.switch_to_window(self.browser.window_handles[0])
+        self.browser.close_window(self.browser.window_handles[-1])
         return hostname
 
 

### CODE REVIEW

**Review by mirekdlugosz — APPROVED** [review]
My only concern is, shouldn't we close this new tab / switch back to original tab before returning from method?

I'm not sure where cockpit integration testing will be going, but I imagine at one point someone might want to interact with Satellite after calling `get_webconsole_content`. Something like:
```python
cockpit_hostname = session.host.get_webconsole_content(hostname)
session.host.delete(cockpit_hostname)
```
Right now, I'm afraid any interaction with Airgun after this method call will fail, as browser is left in unexpected state.

Otherwise, looks OK.

airgun/entities/host.py
```diff
@@ -223,6 +223,13 @@ def get_webconsole_content(self, entity_name):
         view = self.navigate_to(self, 'Details', entity_name=entity_name)
         view.webconsole.click()
         view.validations.assert_no_errors()
+        # switch to the last opened tab,
+        self.browser.selenium.switch_to.window(self.browser.selenium.window_handles[-1])
```
mshriver, line 227: You can use widgetastic's browser methods here to deal with window handles, you shouldn't have to use selenium directly.

airgun/entities/host.py
```diff
@@ -223,6 +223,13 @@ def get_webconsole_content(self, entity_name):
...
+        wait_for(
```
mshriver, line 228: Again there is a widgetastic browser method, `wait_for_element`, that is better to use than directly calling selenium methods.

tstrych:
so I rewrite it, 
It takes me much more time as I would expect. 
As xpaths should be in views. I was wondering if I should put those parts into view, but I didn't know how can I change the frame inside a view. Because if I need to switch frames in entity then I am probably able to do put xpaths into two different views, but that would be messy IMO. 

Test result.
```
collected 30 items / 29 deselected / 1 selected

test_host.py::test_positive_cockpit 

=================== 1 passed, 29 deselected in 76.26 seconds ===================
```

@mirekdlugosz I also tried  to add 
` assert session.host.search(settings.server.hostname)[0]['Name'] == settings.server.hostname`
at the end of [cockpit test ](https://github.com/SatelliteQE/robottelo/blob/master/tests/foreman/ui/test_host.py#L1942)
wanted to know if the airgun is capable to continue, and the test passes so I think it is

mshriver (approved):

mirekdlugosz:
> As xpaths should be in views. I was wondering if I should put those parts into view, but I didn't know how can I change the frame inside a view.

In that case, I would say switching frames would be navmazing responsibility. But I don't think there is any obvious place in navmazing where you could inject closing tab/switching view back. That probably would need to be done by entity.

But I wouldn't lose sleep over it. Cockpit already showed in the past that it doesn't fit perfectly into how Airgun sees things, so drifting away from usual way of doing things is expected in this case.

> @mirekdlugosz I also tried to add
> ` assert session.host.search(settings.server.hostname)[0]['Name'] == settings.server.hostname`
> at the end of [cockpit test ](https://github.com/SatelliteQE/robottelo/blob/master/tests/foreman/ui/test_host.py#L1942)
> wanted to know if the airgun is capable to continue, and the test passes so I think it is

Thank you.

Issue #521, opened 2020-09-18
DiscoveryRule view uses different locator for "search" field


rplevka:
there's no longer an `@id` attribute, but rather `name`.

PR #522, opened 2020-09-18
Implementing custom clear() method for the AutoCompleteTextInput

rplevka:
Since this is a react element, the browser.clear() doesn't work. We need to access the 'x' button and click it.

```
$ py.test test_discoveryrule.py::test_positive_end_to_end
=== test session starts ===
platform linux -- Python 3.7.8, pytest-4.6.3, py-1.9.0, pluggy-0.13.1
rootdir: /home/rplevka/work/rplevka/robottelo
plugins: xdist-1.33.0, services-1.3.1, forked-1.2.0, repeat-0.8.0, cov-2.10.1, mock-1.10.4
collecting ... 2020-09-18 12:23:25 - conftest - DEBUG - Collected 1 test cases
collected 1 item                                                                                                                                                             

test_discoveryrule.py . 
=== 1 passed, 4 warnings in 166.85 seconds ===
```

### DIFF
--- a/airgun/widgets.py
+++ b/airgun/widgets.py
@@ -2097,14 +2097,30 @@ class GenericRemovableWidgetItem(GenericLocatorWidget):
 
 class AutoCompleteTextInput(TextInput):
     """Autocomplete Search input field, We must remove the focus from this widget after fill to
-    force the auto-completion list to be hidden.
+    force the auto-completion list to be hidden. Since this is a react component, calling browser
+    clear method directly on the field has no effect, thus we need to clear the field using the 
+    clear button attached to the input
     """
 
+    clear_button = Text(
+        locator="//span[contains(@class,'autocomplete-clear-button') or "
+                "contains(@class,'fa-close')]"
+    )
+
+    def clear(self):
+        """Clears search field value and re-trigger search to remove all
+        filters.
+        """
+        if self.clear_button.is_displayed:
+            self.clear_button.click()
+
     def fill(self, value):
-        changes = super().fill(value)
+        old_value = self.value
+        self.clear()
+        super().fill(value)
         self.browser.plugin.ensure_page_safe()
         self.browser.execute_script('arguments[0].blur();', self.__element__())
-        return changes
+        return self.value != old_value
 
 
 class ToggleButton(Button):
--- a/airgun/views/discoveryrule.py
+++ b/airgun/views/discoveryrule.py
@@ -49,7 +49,7 @@ class DiscoveryRuleCreateView(BaseLoggedInView):
     @View.nested
     class primary(SatTab):
         name = TextInput(id='discovery_rule_name')
-        search = AutoCompleteTextInput(id='search')
+        search = AutoCompleteTextInput(name='discovery_rule[search]')
         host_group = FilteredDropdown(id='discovery_rule_hostgroup_id')
         hostname = TextInput(id='discovery_rule_hostname')
         hosts_limit = TextInput(id='discovery_rule_max_count')


rplevka:
fixes: https://github.com/SatelliteQE/airgun/issues/521

pondrejk (approved):
ack, not sure why make docs fails

lpramuk (approved):

lpramuk:
> ack, not sure why make docs fails

it fails for all PRs :(

airgun/widgets.py
```diff
@@ -2100,7 +2100,20 @@ class AutoCompleteTextInput(TextInput):
     force the auto-completion list to be hidden.
```
mirekdlugosz, line 2100: Since docstring here explains why we need this subclass in the first place, I think it would be good idea to add this new use-case (calling custom `clear()`) here as well.

airgun/widgets.py
```diff
@@ -2100,7 +2100,20 @@ class AutoCompleteTextInput(TextInput):
...
+    def clear(self):
+        """Clears search field value and re-trigger search to remove all
+        filters.
+        """
+        if self.clear_button.is_displayed:
+            self.clear_button.click()
+
     def fill(self, value):
+        self.clear()
         changes = super().fill(value)
```
mirekdlugosz, line 2117: If we were to follow widgetastic usage guidelines to the letter, calling `self.clear()` might change value, so `changes` here should be combined with `self.clear()` output in some way.

I mean situation like this:
1. widget has some content
2. I call `widget.fill('')` with intent of clearing this content
3. `fill()` calls `clear()`, clearing the content
4. `super().fill('')` is called. Since content is already cleared, it doesn't change, so returns False
5. my initial `widget.fill('')` returns False, signifying that widget content was not changed - which is not true at all

I'm personally not very concerned about this, but something to keep in mind.

mirekdlugosz (approved):
ACK with non-blocking comments

rplevka:
@mirekdlugosz I addressed your comments. Could you review the changes, and merge if happy with them?

```
$ py.test -v test_discoveryrule.py -k end
=== test session starts ===
platform linux -- Python 3.7.8, pytest-4.6.3, py-1.9.0, pluggy-0.13.1 -- /home/rplevka/.virtualenvs/robottelo/bin/python3
cachedir: .pytest_cache
rootdir: /home/rplevka/work/rplevka/robottelo
plugins: xdist-1.33.0, services-1.3.1, forked-1.2.0, repeat-0.8.0, cov-2.10.1, mock-1.10.4
collecting ... 2020-10-01 10:17:30 - conftest - DEBUG - Collected 6 test cases

test_discoveryrule.py::test_positive_end_to_end PASSED                                                          [100%]

=== 1 passed, 5 deselected, 4 warnings in 160.79 seconds ===

```

airgun/widgets.py
```diff
@@ -2097,11 +2097,30 @@ def fill(self, values):
...
+        """
+        if self.clear_button.is_displayed:
+            self.clear_button.click()
+
     def fill(self, value):
-        changes = super().fill(value)
+        if self.value != value:
+            changes = True
+        else:
+            changes = False
```
mirekdlugosz, line 2121:
```python
changes = self.value != value
```
?

I see in the past we assigned return value of `super().fill()` to changes, so probably something like below would be more in line with that intent?
```python
def fill(self, value):
    old_value = self.value
    self.clear()
    super().fill(value)
    self.browser.execute_script('arguments[0].blur();', self.__element__())
    return self.value != old_value
```

mirekdlugosz (approved):

PR #528, opened 2020-10-21
Py38 update, docs fix, Black formatting

mshriver:
pyupgrade will automatically apply f-string formatting, as well as many other python version difference changes automatically.

In this case I'm using pre-commit to run and enforce the tool's use in the future. This will automatically update any commits contributors make to use f-strings and other py36 related changes.

I applied black and pyupgrade via pre-commit in a separate commit from any manual changes made to f-strings. 

I've included some tweaks to the docs build and docblocks, so that the docs build actually works now.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,6 +5,19 @@ repos:
   rev: v1.9.0
   hooks:
   - id: reorder-python-imports
+- repo: https://github.com/asottile/pyupgrade
+  rev: v2.7.2
+  hooks:
+  - id: pyupgrade
+    args: [--py36-plus]
+- repo: https://github.com/psf/black
+  rev: 20.8b1
+  hooks:
+  - id: black
+- repo: https://gitlab.com/pycqa/flake8
+  rev: 3.8.4
+  hooks:
+  - id: flake8
 - repo: https://github.com/pre-commit/pre-commit-hooks
   rev: v2.5.0
   hooks:
@@ -12,4 +25,3 @@ repos:
   - id: end-of-file-fixer
   - id: check-yaml
   - id: debug-statements
-  - id: flake8
--- a/airgun/browser.py
+++ b/airgun/browser.py
@@ -65,13 +65,11 @@ class SeleniumBrowserFactory:
         selenium_browser = factory.get_browser()
 
         # navigate to desired url
-        # [...]
 
         # perform post-init steps (e.g. skipping certificate error screen)
         factory.post_init()
 
         # perform your test steps
-        # [...]
 
         # perform factory clean-up
         factory.finalize(passed)
@@ -357,12 +355,12 @@ class DockerBrowser:
 
         # either as context manager
         with DockerBrowser() as browser:
-            # [...]
+            # testing!
 
         # or with manual :meth:`start` and :meth:`stop` calls.
         docker_browser = DockerBrowser()
         docker_browser.start()
-        # [...]
+        # testing!
         docker_browser.stop()
 
     """
--- a/airgun/browser.py
+++ b/airgun/browser.py
@@ -133,7 +133,7 @@ class SeleniumBrowserFactory:
             'Certificate Error' in self._webdriver.title or 'Login' not in self._webdriver.title
         ):
             self._webdriver.get(
-                "javascript:document.getElementById('invalidcert_continue')" ".click()"
+                "javascript:document.getElementById('invalidcert_continue').click()"
             )
         # Workaround maximize_window() not working with chrome in docker
         if not (self.provider == 'docker' and self.browser == 'chrome'):
@@ -374,7 +374,7 @@ class DockerBrowser:
         """
         if docker is None:
             raise DockerBrowserError(
-                'Package docker-py is not installed. Install it in order to ' 'use DockerBrowser.'
+                'Package docker-py is not installed. Install it in order to use DockerBrowser.'
             )
         self.webdriver = None
         self._capabilities = capabilities
@@ -429,7 +429,7 @@ class DockerBrowser:
         else:
             # Reraise the captured exception.
             raise DockerBrowserError(
-                'Failed to connect the webdriver to the containerized ' 'selenium.'
+                'Failed to connect the webdriver to the containerized selenium.'
             ) from exception
 
     def _quit_webdriver(self):
@@ -470,14 +470,14 @@ class DockerBrowser:
         image_version = selenium.__version__
         if not self._client.images(name=self._get_image_name(image_version)):
             LOGGER.warning(
-                'Could not find docker-image for your' 'selium-version "%s"; trying with "latest"',
+                'Could not find docker-image for your selenium-version "%s"; trying with "latest"',
                 self._get_image_name(image_version),
             )
             image_version = 'latest'
             if not self._client.images(name=self._get_image_name(image_version)):
                 raise DockerBrowserError(
-                    'Could not find docker-image "%s"; please pull it'
-                    % self._get_image_name(image_version)
+                    f'Could not find docker-image "{self._get_image_name(image_version)}";'
+                    ' please pull it'
                 )
         # Grab only the test name, get rid of square brackets from parametrize
         # and add some random chars. E.g. 'test_positive_create_0_abc'
@@ -691,7 +691,7 @@ class AirgunBrowser(Browser):
         )
 
         if not result.startswith('data:'):
-            raise Exception("Failed to get file content: %s" % result)
+            raise Exception(f"Failed to get file content: {result}")
         result_index = int(result.find('base64,')) + 7
         return base64.b64decode(result[result_index:])
 
--- a/airgun/browser.py
+++ b/airgun/browser.py
@@ -46,11 +46,10 @@ def _sauce_ondemand_url(saucelabs_user, saucelabs_key):
     :param str saucelabs_key: saucelabs access key
     :return: string representing saucelabs ondemand URL
     """
-    return 'http://{0}:{1}@ondemand.saucelabs.com:80/wd/hub'.format(
-        saucelabs_user, saucelabs_key)
+    return f'http://{saucelabs_user}:{saucelabs_key}@ondemand.saucelabs.com:80/wd/hub'
 
 
-class SeleniumBrowserFactory(object):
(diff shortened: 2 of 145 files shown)


JacobCallahan:
@mshriver while you're at it, do you want to add black?

airgun/browser.py
```diff
@@ -46,11 +46,11 @@ def _sauce_ondemand_url(saucelabs_user, saucelabs_key):
     :param str saucelabs_key: saucelabs access key
     :return: string representing saucelabs ondemand URL
     """
-    return 'http://{0}:{1}@ondemand.saucelabs.com:80/wd/hub'.format(
+    return 'http://{}:{}@ondemand.saucelabs.com:80/wd/hub'.format(
```
JacobCallahan, line 49: is this an example of needing a manual f-string conversion?

mshriver: correct, and should be updated in the latest commit, I manually went through and updated _almost_ all of them.

mshriver:
@JacobCallahan Added black as well. All the pain, all at once.

mirekdlugosz (changes requested):
This will fail entire iqe-satellite plugin pipeline for Satellite 6.8. iqe pipeline installs Python 3.6. iqe pipeline for earlier Satellite versions install Airgun from z-stream branches, so they should be fine.

Why exactly do we force 3.8 in Airgun? What 3.8-exclusive feature are we using or planning to use?

mirekdlugosz:
@mshriver they are pulling in Airgun and Nailgun separately and integrate them on their own. They don't pull Robottelo at all.

.pre-commit-config.yaml
```diff
@@ -5,11 +5,19 @@ repos:
   rev: v1.9.0
   hooks:
   - id: reorder-python-imports
+- repo: https://github.com/asottile/pyupgrade
+  rev: v2.7.2
+  hooks:
+  - id: pyupgrade
+    args: [--py38-plus]
```
mirekdlugosz, line 12: Shouldn't that be `--py36-plus`?

I couldn't find authoritative list of what pyupgrade changes for each Python version, but I tried making sense out of their source code and seems they only consider two things for Python above 3.6:

* `from __future__ import generator_stop`
* `functools.lru_cache` decorator doesn't need to be called

We don't use either in Airgun, so both flags should produce same results for us right now. But I'm little worried this might result in tool changing our syntax in backward-incompatible way sometime in the future, when they decide more changes are OK in 3.8.

mirekdlugosz (commented):
Added one more comment; I don't have time now to review 282f5b6db2c852ffd31a3c39cb8b6d82b07ba349 , but it seems to mostly consist of quotation mark change (`'` -> `"`). Didn't we somehow avoid that in Robottelo?

.pre-commit-config.yaml
mshriver: If we're having to stick to 36 because of the IQE integration, then yes.

mshriver:
@JacobCallahan @mirekdlugosz I'll have to re-apply black and pyupgrade with string normalization turned off via pyproject.toml if we want the same options here as in robottelo. Thumbs up this comment to confirm you want same behavior and I'll re-run the pre-commit from a fresh checkout.

JacobCallahan:
I can go either way with single vs double quotes, so if lost other people want to stay with single, I'm fine with that.

mshriver:
@JacobCallahan same, though my finger memory always goes for single.

I think we should be consistent, I just wish I would have realized that before squashing my commits! No worries, won't take long.

.pre-commit-config.yaml
mshriver: Dialed this back to 36plus.

mshriver:
Commits can be squashed on merge, or I can do it before merge, keeping them separate for now.

mirekdlugosz:
I would keep them separate, to make it easy to skip commit with changes done by black whenever I need to look into history.

mirekdlugosz (approved):
With pyupgrade using 3.6 as baseline, same black configuration as in Robottelo and black not changing quotation characters, now I'm good with these changes. ACK.

airgun/entities/host.py
```diff
@@ -331,8 +329,9 @@ def step(self, *args, **kwargs):
         action_name = kwargs.get('action_name')
         self.VIEW = self.ACTIONS_VIEWS.get(action_name)
         if not self.VIEW:
-            raise ValueError('Please provide a valid action name.'
-                             ' action_name: "{0}" not found.')
+            raise ValueError(
+                'Please provide a valid action name.' ' action_name: "{0}" not found.'
```
JacobCallahan, line 333: i wonder if that 0 index needs to be there

airgun/entities/product.py
```diff
@@ -167,8 +172,9 @@ def step(self, *args, **kwargs):
         action_name = kwargs.get('action_name')
         self.VIEW = self.ACTIONS_VIEWS.get(action_name)
         if not self.VIEW:
-            raise ValueError('Please provide a valid action name.'
-                             ' action_name: "{0}" not found.')
+            raise ValueError(
+                'Please provide a valid action name.' ' action_name: "{0}" not found.'
```
JacobCallahan, line 176: same index question

airgun/entities/subscription.py
```diff
@@ -73,9 +77,7 @@ def refresh_manifest(self):
         view.manifest.refresh_button.click()
         org_name = view.taxonomies.current_org
         self._wait_for_process_to_finish(
-            'Refresh Manifest organization \'{}\''.format(org_name),
-            has_manifest=True,
-            timeout=1200
+            f'Refresh Manifest organization \'{org_name}\'', has_manifest=True, timeout=1200
```
JacobCallahan, line 80:
```suggestion
            f'Refresh Manifest organization "{org_name}"', has_manifest=True, timeout=1200
```

airgun/session.py
```diff
@@ -214,8 +216,7 @@ def __exit__(self, exc_type, exc_value, traceback):
         if self.browser is None:
             # browser was never started, don't do anything
             return
-        LOGGER.info(
-            u'Stopping UI session %r for user %r', self.name, self._user)
+        LOGGER.info('Stopping UI session %r for user %r', self.name, self._user)
```
JacobCallahan, line 219: f-string?

airgun/session.py
```diff
@@ -245,13 +246,14 @@ def _prepare_browser(self, url):
         satellite.
         """
         if self._session_cookie:
-            LOGGER.info('Starting UI session id: %r from a session cookie',
-                        self._session_cookie.cookies.get_dict()['_session_id'])
+            LOGGER.info(
+                'Starting UI session id: %r from a session cookie',
+                self._session_cookie.cookies.get_dict()['_session_id'],
```
JacobCallahan, line 251: f-string? looks like multiple cases of old string formatting in this file

airgun/browser.py
```diff
@@ -135,17 +129,14 @@ def post_init(self):
...
-                self.browser == 'edge' and
-                ('Certificate Error' in self._webdriver.title or
-                    'Login' not in self._webdriver.title)):
+        if self.browser == 'edge' and (
+            'Certificate Error' in self._webdriver.title or 'Login' not in self._webdriver.title
+        ):
             self._webdriver.get(
-                "javascript:document.getElementById('invalidcert_continue')"
-                ".click()"
+                "javascript:document.getElementById('invalidcert_continue')" ".click()"
```
JacobCallahan, line 136:
```suggestion
                "javascript:document.getElementById('invalidcert_continue').click()"
```

airgun/browser.py
```diff
@@ -395,8 +374,7 @@ def __init__(self, name=None, image=None, capabilities=None):
         """
         if docker is None:
             raise DockerBrowserError(
-                'Package docker-py is not installed. Install it in order to '
-                'use DockerBrowser.'
+                'Package docker-py is not installed. Install it in order to ' 'use DockerBrowser.'
```
JacobCallahan, line 377:
```suggestion
                'Package docker-py is not installed. Install it in order to use DockerBrowser.'
```

airgun/browser.py
```diff
@@ -438,23 +415,21 @@ def _init_webdriver(self):
...
+                time.sleep(0.5)
             else:
                 # Connection succeeded time to leave the for loop
                 break
         else:
             # Reraise the captured exception.
             raise DockerBrowserError(
-                'Failed to connect the webdriver to the containerized '
-                'selenium.'
+                'Failed to connect the webdriver to the containerized ' 'selenium.'
```
JacobCallahan, line 432:
```suggestion
                'Failed to connect the webdriver to the containerized selenium.'
```

airgun/browser.py
```diff
@@ -494,40 +468,34 @@ def _create_container(self):
...
+            )
             image_version = 'latest'
-            if not self._client.images(
-                    name=self._get_image_name(image_version)):
+            if not self._client.images(name=self._get_image_name(image_version)):
                 raise DockerBrowserError(
-                    'Could not find docker-image "%s"; please pull it' %
-                    self._get_image_name(image_version)
+                    'Could not find docker-image "%s"; please pull it'
+                    % self._get_image_name(image_version)
```
JacobCallahan, line 480: f-string or format please

JacobCallahan (changes requested):

airgun/session.py
mshriver: Not for logger calls, the logger does its own formatting, and takes the fields as arguments directly.

JacobCallahan: python's base logger really needs an update..

airgun/entities/host.py
mshriver: Good call - I think this is one of the situations where order of pyupgrade and black in pre-commit may have an impact.

airgun/browser.py
mshriver: I looked for an found other instances of this, they should be fixed in the latest commit.

airgun/browser.py
mshriver: Found and fixed all of these instances for both single and double quotes, good catch!

airgun/browser.py
mshriver: Found and fixed all remaining uses of % for formatting.

JacobCallahan (approved):
ACK

PR #529, opened 2020-10-26
cp #517 - fix cockpit test, cockpit now opens in new tab

tstrych:
#517 
I requested the same reviewers for ACK thanks.

Test result: 
```
test_host.py::test_positive_cockpit 

=================== 1 passed, 29 deselected in 81.73 seconds ===================
```

### DIFF
--- a/airgun/entities/host.py
+++ b/airgun/entities/host.py
@@ -218,18 +218,24 @@ class HostEntity(BaseEntity):
         view = self.navigate_to(self, 'Details', entity_name=entity_name)
         view.webconsole.click()
         view.validations.assert_no_errors()
+        # switch to the last opened tab,
+        self.browser.switch_to_window(self.browser.window_handles[-1])
+        self.browser.wait_for_element(
+            locator='//div[@id="index-brand"]', exception=True
+        )
         # the remote host content is loaded in an iframe, let's switch to it
-        self.browser.selenium.switch_to.frame(0)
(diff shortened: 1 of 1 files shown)


airgun/entities/host.py
```diff
@@ -218,18 +218,24 @@ def get_webconsole_content(self, entity_name):
         view = self.navigate_to(self, 'Details', entity_name=entity_name)
         view.webconsole.click()
         view.validations.assert_no_errors()
+        # switch to the last opened tab,
+        self.browser.switch_to_window(self.browser.window_handles[-1])
```
mshriver, line 222: This is dangerous behavior to be run directly without some mechanism to mitigate failures to do anything between when you switch tot he window and when you switch back to window handle 0.

If there are any failures in any of the commands in between, the browser needs a mechanism to get back to the original tab.

I would at a minimum put the behavior into a try/except block with a generic exception catch, so that the browser can be returned to the original window handle.

Providing a mechanism within the library, not just within this method, is something we need to consider for core airgun refactoring efforts.

mshriver: @tstrych pointed out this is a backport of previously accepted behavior.

mshriver (approved):

mshriver:
Travis failure is due to docs build, getting resolved in #528

mirekdlugosz (approved):
ACK

PR #534, opened 2020-11-12
Add the ability to create a session for a different satellite

JacobCallahan:
This change allows a user to pass in a different hostname than
what is currently in airgun's settings.

Example:
with Session('test', hostname='my.other.sat.com') as session:
    session.do_something()

### DIFF
--- a/airgun/browser.py
+++ b/airgun/browser.py
@@ -76,7 +76,9 @@ class SeleniumBrowserFactory:
 
     """
 
-    def __init__(self, provider=None, browser=None, test_name=None, session_cookie=None):
+    def __init__(
+        self, provider=None, browser=None, test_name=None, session_cookie=None, hostname=None
+    ):
         """Initializes factory with either specified or fetched from settings
         values.
 
@@ -93,6 +95,8 @@ class SeleniumBrowserFactory:
             used otherwise.
         :param requests.sessions.Session optional session_cookie: session object to be used
             to bypass login
+        :param str optional hostname: The hostname of a target that differs from
+            settings.satellite.hostname
         """
         self.provider = provider or settings.selenium.browser
         self.browser = browser or settings.selenium.webdriver
@@ -100,6 +104,7 @@ class SeleniumBrowserFactory:
         self._session = session_cookie
         self._docker = None
         self._webdriver = None
+        self._hostname = hostname or settings.satellite.hostname
 
     def get_browser(self):
         """Returns selenium webdriver instance of selected ``provider`` and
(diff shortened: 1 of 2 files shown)


mshriver (approved):

airgun/session.py
```diff
@@ -170,8 +171,11 @@ def __init__(
             to bypass login
         :param str optional url: URL path to open when starting session (without protocol
             and hostname)
+        :param str optional hostname: The hostname of a target that differs from
+            settings.satellite.hostname
         """
         self.name = session_name or gen_string('alphanumeric')
+        self._hostname = hostname or settings.satellite.hostname
```
mirekdlugosz, line 178: Right now, attributes are set in the same order named parameters are defined (if you ignore public attributes between them). I think it would be nice to follow that convention.

airgun/session.py
```diff
@@ -154,6 +154,7 @@ def __init__(
         session_cookie=None,
         url=None,
         login=True,
+        hostname=None,
```
mirekdlugosz, line 157: I think that should be before `url`, since URI is constructed by appending `url` to `hostname`.
Remember to also change docstring.

mirekdlugosz (commented):
There's reference to `settings.satellite.hostname` in `airgun/browser.py`. I think it should be changed as part of this PR.

airgun/session.py
JacobCallahan: I decided on this order so anyone using the session object with positional arguments wouldn't be affected. If we can determine that there are no instances of positional-only arguments, then I can move it up.
Docstring has been updated

airgun/session.py
JacobCallahan: Sure, I can change that

airgun/session.py
mirekdlugosz: Fair concern. Fortunately, there are no positional-only arguments past `password`. `session_cookie` effectively forces everything later to be keyword-only, as nothing in Robottelo actually uses `session_cookie` (but people may use it locally to save some time during test development).

Checked with following commands in Robottelo root dir:
```bash
$ ack 'session\((.*?, ?){3}'
$ ack 'session\('
```

iqe-satellite-plugin only ever passes test name.

JacobCallahan: if you say it is safe, then I can amend my latest changes

JacobCallahan:
@mirekdlugosz updated per your requests

airgun/browser.py
```diff
@@ -76,7 +76,7 @@ class SeleniumBrowserFactory:
 
     """
 
-    def __init__(self, provider=None, browser=None, test_name=None, session_cookie=None):
+    def __init__(self, provider=None, browser=None, test_name=None, session_cookie=None, hostname=None):
```
mirekdlugosz, line 79: Looks like black doesn't like this line and causes travis to fail

mirekdlugosz (approved):
ACK, but not merging due to travis failure.

latran (approved):
Ack, waiting for fix in formatting for travis pass.

mshriver (approved):

PR #535, opened 2020-11-13
Add support to install all CHost's errata

swadeley:
Hello

I need this for a new pagination test. Test results will be in that test soon.

Thank you

### DIFF
--- a/airgun/entities/contenthost.py
+++ b/airgun/entities/contenthost.py
@@ -125,8 +125,11 @@ class ContentHostEntity(BaseEntity):
         :return: Returns a dict containing task status details
         """
         view = self.navigate_to(self, 'Edit', entity_name=entity_name)
-        view.errata.search(errata_id)
-        view.errata.table.row(id=errata_id)[0].widget.fill(True)
+        if errata_id == "All":
+            view.errata.select_all.fill(True)
+        else:
+            view.errata.search(errata_id)
+            view.errata.table.row(id=errata_id)[0].widget.fill(True)
         install_via_dict = {
             'katello': 'via Katello agent',
             'rex': 'via remote execution',


swadeley:
Hello

This is the test:

[ Test chost errata pagination #8173 ](https://github.com/SatelliteQE/robottelo/pull/8173)

airgun/entities/contenthost.py
```diff
@@ -141,6 +141,33 @@ def install_errata(self, entity_name, errata_id, install_via=None):
...
+        """Install all errata on a content host
+
+        :param name: content host name to apply errata on
+        :param str install_via: via which mean to install errata. Available
+            options: "katello", "rex", "rex_customize"
+
+        :return: Returns a dict containing task status details
+        """
+        view = self.navigate_to(self, 'Edit', entity_name=entity_name)
+        breakpoint()
```
lhellebr, line 154: I suppose this isn't intended?

swadeley: oh dear

swadeley: @lhellebr Fixed now, thank you for spotting that.

mirekdlugosz (commented):
Since this is almost identical to `install_errata` method, I think it would be better to teach existing method to handle special errata id `All`, similar to what we already do in [airgun/entities/host.py](https://github.com/SatelliteQE/airgun/blob/master/airgun/entities/host.py#L336) for entities list:

```python
    def install_errata(self, entity_name, errata_id, install_via=None):
        if errata_id == "All":
            view.errata.select_all.fill(True)
        else:
            view.errata.search(errata_id)
            view.errata.table.row(id=errata_id)[0].widget.fill(True)
        install_via_dict = {
            # skipping rest of method
```

Reviewers: this PR depends on change introduced in #532

mirekdlugosz (approved):
Thanks, @swadeley !
ACK

omkarkhatavkar (approved):
ACK

PR #538, opened 2020-11-25
removed trends

pondrejk:
Trends and statistics dropped from 6.9


omkarkhatavkar (approved):
ACK

mirekdlugosz (changes requested):
There are remaining references to `trend` in `airgun/session.py`:
https://github.com/SatelliteQE/airgun/blob/582ce07e5f5a5c5c691739774ee33104420ff219/airgun/session.py#L76
https://github.com/SatelliteQE/airgun/blob/582ce07e5f5a5c5c691739774ee33104420ff219/airgun/session.py#L618-L621

This unresolved import makes Airgun broken completely (you can't import `Session`, which is our single point of entry for Robottelo), and also fails travis.

mirekdlugosz (approved):
Thanks for update. Looks good to me now.
ACK

PR #542, opened 2020-12-01
Fix Http proxy password field

jameerpathan111:


mirekdlugosz (approved):
I see this button changes "disabled" state of accompanied input. Right now it's disabled by default, but possibly that would not always be a case - probably we should only click if input is disabled? So we don't disable it by mistake.

If similar widget (input + button to disable) is used elsewhere, probably we should abstract it away into new widget. Then `fill()` method would check if input is disabled and enable it, if needed.

But these concerns are not urgent, so I see no reason to block it right now. ACK

omkarkhatavkar:
> I see this button changes "disabled" state of accompanied input. Right now it's disabled by default, but possibly that would not always be a case - probably we should only click if input is disabled? So we don't disable it by mistake.
> 
> If similar widget (input + button to disable) is used elsewhere, probably we should abstract it away into new widget. Then `fill()` method would check if input is disabled and enable it, if needed.
> 
> But these concerns are not urgent, so I see no reason to block it right now. ACK

Same suggestion from my side which I was discussing with you. can be done in future in case same happens

omkarkhatavkar (approved):
ACK

PR #544, opened 2020-12-14
locator fix for account menu

tstrych:
locator for account menu changed in 6.9 

```
collected 30 items / 29 deselected / 1 selected

test_host.py::test_positive_read_from_details_page 2020-12-14 15:30:34 - robottelo - DEBUG - Finished Test: test_host.py/test_positive_read_from_details_page
========== 1 passed, 29 deselected, 21 warnings in 132.23s (0:02:12) ===========
```

### DIFF
--- a/airgun/views/common.py
+++ b/airgun/views/common.py
@@ -39,8 +39,8 @@ class BaseLoggedInView(View):
     taxonomies = ContextSelector()
     flash = SatFlashMessages()
     validations = ValidationErrors()
-    current_user = Text("//a[@id='account_menu']")
-    account_menu = Text("//a[@id='account_menu']")
+    current_user = Text("//li[@id='account_menu']")
+    account_menu = Text("//li[@id='account_menu']")
     logout = Text("//a[@href='/users/logout']")
 
     def select_logout(self):


omkarkhatavkar:
Black is complaining @tstrych

tstrych:
thank you will take a look :)

airgun/views/common.py
```diff
@@ -34,13 +34,13 @@ class BaseLoggedInView(View):
     """Base view for Satellite pages"""
 
     menu = SatVerticalNavigation(
-        './/div[@id="vertical-nav" or contains(@class, "nav-pf-vertical")]/ul'
+            './/div[@id="vertical-nav" or contains(@class, "nav-pf-vertical")]/ul'
```
tpapaioa, line 37: This extra indentation should be removed.

airgun/views/common.py
```diff
@@ -34,13 +34,13 @@ class BaseLoggedInView(View):
...
     )
     taxonomies = ContextSelector()
     flash = SatFlashMessages()
     validations = ValidationErrors()
-    current_user = Text("//a[@id='account_menu']")
-    account_menu = Text("//a[@id='account_menu']")
+    current_user = Text("//li[@id='account_menu']")
+    account_menu = Text("//li[@id='account_menu']")
```
tpapaioa, line 43: Both `current_user` and `account_menu` are duplicates, using the same absolute locator. Looking at the commit log, it looks like `current_user` is older and obsolete, since it's not referenced anywhere else in the code. It can be removed in favor of the more accurately-named `account_menu`, which is referenced in the `select_logout()` method.

tstrych: Finally has time to look at it, sorry for the black error, I probably miss-typed `\tab`.
@tpapaioa I do agree on that current user is older and obsolete. 

But in the robottelo in many tests there is `current_user` and on the other hand no `account_menu` used.
Sure I agree with  you current user should be renamed to account menu  even though account menu is used only in few cases and current user in many more. 
IMO this should be part of another PR, but I also think this is not so important and if I will not create the new PR I guess nobody will because this is low priority. 
@tpapaioa if you wish I can do that :) 
So another PR in robottelo to rename the `current_user` and another PR in airgun to delete `current_user`

For reference: I used pycharm for search and I searched robottelo and airgun repos. When I used github search with both of the words I didn't find anything except issues and I don't know why. 

Please let me know what you think :)

lhellebr: I think this PR is so simple it's not necessary to create yet another PR. But if you need this fixed ASAP and don't have time to fix it in Robottelo, why not. I just think it will never get fixed in that case :)

tpapaioa: I think maybe `current_user` could be made into a property/method that reads the text from the `account_menu` widget, so that robottelo tests don't need to be changed. If you don't want to do that here, I can work on it later.

tpapaioa (approved):

swadeley (approved):
ACK

mshriver (approved):

PR #545, opened 2021-01-05
Add ROOT for Search widget

tpapaioa:
Some robottelo tests are failing because the `search_field` subwidget cannot be found. This is happening because there are two hidden search fields higher up in the page, associated with the Organization and Location context dropdowns. The `search_field` locator matches those, and because they are not visible, any attempts to interact with `search_field` fail.

This PR fixes the issue by adding a `ROOT` locator to `Search`:

`ROOT = '//div[@id="search-bar" or contains(@class, "toolbar-pf-filter")]'`

The two locators are necessary because there are two distinct search bar implementations. Some pages (e.g., `Administer > Locations`) use the `search-bar` version, whereas others (e.g., `Content > Products`), use `toolbar-pf-filter`.

### DIFF
--- a/airgun/widgets.py
+++ b/airgun/widgets.py
@@ -628,6 +628,7 @@ class ActionDropdownWithCheckbox(ActionsDropdown):
 class Search(Widget):
     """Searchbar for table filtering"""
 
+    ROOT = '//div[@id="search-bar" or contains(@class, "toolbar-pf-filter")]'
     search_field = TextInput(
         locator=(
             ".//input[@id='search' or contains(@placeholder, 'Filter') or "


airgun/widgets.py
```diff
@@ -627,7 +627,7 @@ def fill(self, item):
 
 class Search(Widget):
     """Searchbar for table filtering"""
-
+    ROOT = '//div[@id="search-bar" or contains(@class, "toolbar-pf-filter")]'
     search_field = TextInput(
```
jameerpathan111, line 631: @tpapaioa thanks for fixing this, the search seems to be working for me.
But I am facing one weird issue, when I am on read page of a repository, and try to create new repo then it navigates to settings page instead, and fails with `*** selenium.common.exceptions.NoSuchElementException: Message: Could not find an element Locator(by='xpath', locator=".//span[contains(@class, 'editable-click')]")
` error. Not sure if I did anything wrong.

jameerpathan111: tried again, something wrong with my test only, normally this is working fine. Need to debug. :|

tpapaioa: That error appears to be an issue with finding the following element in `airgun/widgets.py`:

```
class PopOverWidget(Widget):
[...]
    ROOT = '.'
    column_value = Text(".//span[contains(@class, 'editable-click')]")
```

That widget is used in `airgun/views/settings.py`:

```
class SettingsView(BaseLoggedInView, SearchableViewMixin):
    title = Text("//h1[text()='Settings']")
    table = Table(
        './/table',
        column_widgets={'Value': PopOverWidget()},
    )
```

I don't think that the change to the `Search` widget would affect this usage of `Table` / `PopOverWidget`, but if you see the error again I'll be happy to look into it.

jameerpathan111: yes, it's not because of your changes. Created https://github.com/SatelliteQE/airgun/issues/546, figured it out what's the root cause and how to fix it.

jameerpathan111 (approved):
LGTM

lhellebr (approved):
LGTM

omkarkhatavkar (approved):
ACK

PR #548, opened 2021-01-06
Fix logger text

tpapaioa:
This PR makes a small fix to the logging in `session`, where an f-string was not properly defined. Log mesages of the form

`INFO     airgun.session:session.py:264 Setting initial URL to {url}`

are now properly formatted:

`INFO     airgun.session:session.py:264 Setting initial URL to https://satellite.example.com/settings`


swadeley (approved):
ACK

omkarkhatavkar (approved):
ACK

PR #555, opened 2021-02-08
wait for registration dialog to come up

pondrejk:
previously airgun failed on attempting to fill in the form before actually getting to the dialog

### DIFF
--- a/airgun/entities/host.py
+++ b/airgun/entities/host.py
@@ -47,9 +47,11 @@ class HostEntity(BaseEntity):
     def get_register_command(self, values):
         """Get curl command generated on Register Host page"""
         view = self.navigate_to(self, 'Register')
+        view.breadcrumb.wait_displayed()
         view.fill(values)
         self.browser.click(view.generate_command)
         self.browser.plugin.ensure_page_safe()
+        view.registration_command.wait_displayed()
         return view.registration_command.read()
 
     def search(self, value):


airgun/entities/host.py
```diff
@@ -47,6 +47,11 @@ def create(self, values):
     def get_register_command(self, values):
         """Get curl command generated on Register Host page"""
         view = self.navigate_to(self, 'Register')
+        wait_for(
+            lambda: HostRegisterView(self.browser).is_displayed is True,
+            timeout=60,
+            delay=1,
+        )
```
lhellebr, line 54: Shouldn't this be done by `navigate_to`? Isn't it a bug if you need to do it manually?

tpapaioa: As @lhellebr already mentioned, `navigate_to` already calls `is_displayed` on the view after performing the navigation steps. It would have already returned `True` by the time you call it a second time, so if that seems to resolve the issue, then it's probably only because of the additional delay you've introduced.

A better fix would be to make sure `is_displayed` is more robust. Right now, it only checks if the breadcrumb widget is displayed properly:

```
airgun/views/host.py:

class HostRegisterView(BaseLoggedInView):
[...]
    @property
    def is_displayed(self):
        breadcrumb_loaded = self.browser.wait_for_element(self.breadcrumb, exception=False)
        return (
            breadcrumb_loaded
            and self.breadcrumb.locations[0] == 'Registrations'
            and self.breadcrumb.read() == 'Register Host'
        )
```
Checking for the visibility of the other widgets here is probably what you want.

pondrejk: The problem here is that navigate_to (navmazing?) checks for the breadcrumbs before they have a chance to load, so the delay is what I need, checking for additional widgets won't make much difference.  I see it is not ideal but it unblocks me for my feature automation. I welcome suggestions from airgun gurus on how to make the navigation wait for the results of this actions but maybe without blocking this PR :)

lhellebr: If the only goal here is delay, I think it would be better (though suboptimal) to just add delay explicitly.

mshriver: The correct thing to do for this behavior is to use `widget.wait_displayed` within the `is_displayed` check:

https://github.com/RedHatQE/widgetastic.core/blob/master/src/widgetastic/widget/base.py#L529

pondrejk: I don't see why, wait_for was introduced for exactly for these cases so that we don't have to guesswork the hard delay duration.

lhellebr: As discussed above, the waiting condition is already met from the start so this probably does the same thing as explicit delay, it's just more complicated.
I think @tpapaioa suggested the right thing to do here. If you don't feel like doing it, why not just keep the workaround as simple as possible?

pondrejk: Guys, it is not easy to argue when you don't see what I see when I run this. I ensure you that any changes made within `is_displayed` had zero effect on the issue, maybe the view.fill kicks in too soon, I don't honestly know. Also, passing this with wait_displayed, I get to another point when airgun doesn't wait on the result of clicking the generate_command button, hence the second wait_displayed

mshriver: @pondrejk with your latest commit using wait_displayed,  are you still seeing the behavior of it skipping past the generate_command check?

I think that `after_fill` behavior may capture the interaction with this form - where after fill a button gets clicked and some redirection/redraw happens, there is a wait on registration_command.

That being said, your current commit adding wait_displayed looks fine to me.

mshriver (approved):
There is an open discussion about the pattern used in this function, and whether or not the widget waits could be part of `is_displayed` for the relevant view.

I believe the outcome of that open discussion will be targeted for general airgun improvements and refactoring.

I'm fine with the addition of these explicit `wait_displayed` calls in this function.

tpapaioa (approved):

airgun/entities/host.py
pondrejk: @mshriver without the second wait_displayed() it gets past the generate_command, I can try after_fill (I just learned about it now), I think the original author intended the `self.browser.plugin.ensure_page_safe()` to do the job, so that's maybe another area we need to look at

PR #556, opened 2021-02-16
fixed hostgroup locator for global registration

pondrejk:
Id didn't match the reality, PR brings the fix plus matches the var name to the id


omkarkhatavkar (approved):
ACK

swadeley (approved):
ACK

PR #563, opened 2021-04-12
Fix CustomSearch widget used in ModuleStreamsDetailsView

tpapaioa:
The `CustomSearch` widget used in `ModuleStreamsDetailsView` was failing because the locator for the `search_field` sub-widget doesn't exist (anymore?). This PR fixes the issue by updating the locator to match on `id`. I've also removed the `clear_button` sub-widget, which doesn't exist on this particular page's search bar. Any calls to the `clear()` method should still work, because it falls back on clearing the text in `search_field` input field if `clear_button` isn't available. In `airgun/widgets.py`:
```
class Search(Widget):
[...]
    def clear(self):
        """Clears search field value and re-trigger search to remove all
        filters.
        """
        if self.clear_button.is_displayed:
            self.clear_button.click()
        else:
            self.browser.clear(self.search_field)
        if self.search_button.is_displayed:
            self.search_button.click()
```

### DIFF
--- a/airgun/views/modulestream.py
+++ b/airgun/views/modulestream.py
@@ -12,8 +12,7 @@ from airgun.widgets import Search
 
 
 class CustomSearch(Search):
-    search_field = TextInput(locator=".//input[@role='combobox']")
-    clear_button = Text(".//span[contains(@class,'fa-times')]")
+    search_field = TextInput(id='downshift-0-input')
     search_button = Button('Search')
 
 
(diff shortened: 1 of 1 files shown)


JacobCallahan (approved):

latran (approved):
This does help in fixing `test_positive_module_stream_details_search_in_repo`, but the test still does fail with a regression.  A bz has been filed.  This is an ACK from me.

PR #566, opened 2021-04-16
Fix logging if screenshot save failed

tpapaioa:
Airgun's session context manager tries to capture a screenshot of the browser if an exception has been raised, and then log any exception that `take_screenshot()` itself might raise. But there are also cases in which `take_screenshot()` fails, returning `False` without raising an exception. In particular, the underlying selenium webdriver method `get_screenshot_as_file()` that ends up getting called will catch any exceptions from the OS while trying to save the screenshot to file: 

```
        try:
            with open(filename, 'wb') as f:
                f.write(png)
        except IOError:
            return False
```

I've added a check on the return value of `take_screenshot()`, and some error logging in case of failure.

### DIFF
--- a/airgun/session.py
+++ b/airgun/session.py
@@ -306,7 +306,8 @@ class Session:
             f'{self.name}-screenshot-{now.strftime("%Y-%m-%d_%H_%M_%S")}.png',
         )
         LOGGER.debug('Saving screenshot %s', path)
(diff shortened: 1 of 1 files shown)


swadeley (approved):
ACK

mshriver (approved):

PR #568, opened 2021-04-27
Update rh_cloud inventory page code to support pf4

jameerpathan111:
This PR fixes Entities/Views for Inventory upload page.
The page is based on PF4 so I have used widgets available in widgetastic_patternfly4 library(Specifically the Switch widget).

### DIFF
--- a/airgun/browser.py
+++ b/airgun/browser.py
@@ -658,13 +658,14 @@ class AirgunBrowser(Browser):
         downloads_uri = 'chrome://downloads'
         if not self.url.startswith(downloads_uri):
             self.url = downloads_uri
-        return self.execute_script(
-            """
-            return downloads.Manager.get().items_
-              .filter(e => e.state === "COMPLETE")
-              .map(e => e.file_url || e.fileUrl);
-        """
+        time.sleep(3)
+        script = (
+            'return document.querySelector("downloads-manager")'
+            '.shadowRoot.querySelector("#downloadsList")'
+            '.items.filter(e => e.state === "COMPLETE")'
+            '.map(e => e.filePath || e.file_path || e.fileUrl || e.file_url);'
         )
+        return self.execute_script(script)
 
     def get_file_content(self, uri):
         """Get file content by its URI from browser's downloads page.
--- a/airgun/entities/cloud_inventory.py
+++ b/airgun/entities/cloud_inventory.py
@@ -12,22 +12,19 @@ class CloudInventoryEntity(BaseEntity):
     def read(self, entity_name=None, widget_names=None):
         view = self.navigate_to(self, 'All')
         if entity_name:
-            view.searchbox.search(entity_name)
             entity_item = next(view.inventory_list[entity_name])
             return entity_item.read()
-        view.searchbox.clear()
         return view.read(widget_names=widget_names)
 
     def generate_report(self, entity_name):
         view = self.navigate_to(self, 'All')
-        view.searchbox.search(entity_name)
+        #entity_item = view.inventory_list.expand(entity_name)
         entity_item = next(view.inventory_list[entity_name])
         entity_item.browser.click(entity_item.generating.restart, ignore_ajax=True)
         wait_for(lambda: entity_item.status == 'idle', timeout=180, delay=1, logger=view.logger)
 
     def download_report(self, entity_name):
         view = self.navigate_to(self, 'All')
-        view.searchbox.search(entity_name)
         entity_item = next(view.inventory_list[entity_name])
         entity_item.browser.click(entity_item.uploading.download_report, ignore_ajax=True)
         wait_for(lambda: entity_item.status == 'idle', timeout=180, delay=1, logger=view.logger)
--- a/airgun/entities/cloud_inventory.py
+++ b/airgun/entities/cloud_inventory.py
@@ -12,20 +12,22 @@ class CloudInventoryEntity(BaseEntity):
     def read(self, entity_name=None, widget_names=None):
         view = self.navigate_to(self, 'All')
         if entity_name:
-            entity_item = next(view.inventory_list[entity_name])
+            view.inventory_list.toggle(entity_name)
+            entity_item = view.inventory_list
             return entity_item.read()
         return view.read(widget_names=widget_names)
 
     def generate_report(self, entity_name):
         view = self.navigate_to(self, 'All')
-        #entity_item = view.inventory_list.expand(entity_name)
-        entity_item = next(view.inventory_list[entity_name])
+        view.inventory_list.toggle(entity_name)
+        entity_item = view.inventory_list
         entity_item.browser.click(entity_item.generating.restart, ignore_ajax=True)
         wait_for(lambda: entity_item.status == 'idle', timeout=180, delay=1, logger=view.logger)
 
     def download_report(self, entity_name):
         view = self.navigate_to(self, 'All')
-        entity_item = next(view.inventory_list[entity_name])
+        view.inventory_list.toggle(entity_name)
+        entity_item = view.inventory_list
         entity_item.browser.click(entity_item.uploading.download_report, ignore_ajax=True)
         wait_for(lambda: entity_item.status == 'idle', timeout=180, delay=1, logger=view.logger)
         return self.browser.save_downloaded_file()
--- a/airgun/entities/cloud_inventory.py
+++ b/airgun/entities/cloud_inventory.py
@@ -1,3 +1,5 @@
+from time import sleep
+
 from wait_for import wait_for
 
 from airgun.entities.base import BaseEntity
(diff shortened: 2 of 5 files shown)


airgun/entities/cloud_inventory.py
```diff
@@ -11,33 +13,59 @@ class CloudInventoryEntity(BaseEntity):
...
         view = self.navigate_to(self, 'All')
-        view.searchbox.search(entity_name)
-        entity_item = next(view.inventory_list[entity_name])
-        entity_item.browser.click(entity_item.generating.restart, ignore_ajax=True)
-        wait_for(lambda: entity_item.status == 'idle', timeout=180, delay=1, logger=view.logger)
+        view.inventory_list.toggle(entity_name)
+        view.browser.click(view.inventory_list.generating.restart, ignore_ajax=True)
+        # FIX:
+        sleep(5)
+        wait_for(
```
omkarkhatavkar, line 28: instead hard sleep can you able to increase the delay here to 5, if that works ?

tpapaioa (changes requested):

airgun/views/cloud_inventory.py
```diff
@@ -3,15 +3,13 @@
 from widgetastic.utils import ParametrizedLocator
 from widgetastic.widget import Text
 from widgetastic.widget import View
-from widgetastic.widget import Widget
 from widgetastic_patternfly import Button
-from widgetastic_patternfly import ItemsList
-from widgetastic_patternfly import ListItem
 from widgetastic_patternfly import Tab
+from widgetastic_patternfly4.switch import Switch
```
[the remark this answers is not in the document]
mshriver: @jameerpathan111 The widgetastic_patternfly4 requirement needs to be covered by airgun's requirements.   

Please add it to the requirements here: https://github.com/SatelliteQE/airgun/blob/master/setup.py#L26

airgun/widgets.py
```diff
@@ -2225,3 +2226,21 @@ def count(self):
             return int(self.browser.text(self.browser.element(self.COUNT)))
         except NoSuchElementException:
             return None
+
+
+class Accordion(View, ClickableMixin):
+    """PF4 Accordion widget"""
```
mshriver, line 2232: TODO: contribute this to widgetastic_patternfly.pf4 and import it.

airgun/widgets.py
```diff
@@ -2225,3 +2226,21 @@ def count(self):
...
+
+    ROOT = ParametrizedLocator("{@locator}")
+    ITEMS = ".//button[contains(@class, 'pf-c-accordion__toggle')]"
+    ITEM = ".//span[contains(text(), '{}')]"
+
+    def __init__(self, parent=None, id_=None, locator=None, logger=None):
```
mshriver, line 2238: Why the underscore after `id_` here?

mshriver (changes requested):
This is looking great, just a few things to address. Congrats on adding the first PF4 compatible view :)

mshriver:
@jameerpathan111 is this page update included in 6.9 z-streams or just in 6.10? We may have to block it on 6.9 branching.

jameerpathan111:
Test result: (Chrome version 90)
```

$ pytest tests/foreman/ui/test_rhcloud_inventory.py 
platform linux -- Python 3.8.6, pytest-6.2.2, py-1.10.0, pluggy-0.13.1
rootdir: /home/jpathan/projects/robottelo, configfile: pyproject.toml
plugins: ibutsu-1.13, services-2.2.1, cov-2.11.1, mock-3.5.1, reportportal-5.0.7, xdist-2.2.1, forked-1.3.0

tests/foreman/ui/test_rhcloud_inventory.py .                             [100%]
=========== 1 passed, 5 deselected, 15 warnings in 596.61s (0:09:56) ===========
```

airgun/widgets.py

airgun/entities/cloud_inventory.py
```diff
@@ -11,33 +13,59 @@ class CloudInventoryEntity(BaseEntity):
...
         view = self.navigate_to(self, 'All')
-        view.searchbox.search(entity_name)
-        entity_item = next(view.inventory_list[entity_name])
-        entity_item.browser.click(entity_item.uploading.download_report, ignore_ajax=True)
-        wait_for(lambda: entity_item.status == 'idle', timeout=180, delay=1, logger=view.logger)
+        view.inventory_list.toggle(entity_name)
+        view.browser.click(view.inventory_list.uploading, ignore_ajax=True)
+        view.browser.click(view.inventory_list.uploading.download_report, ignore_ajax=True)
+        # FIX:
+        sleep(5)
```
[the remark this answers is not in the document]
jameerpathan111: Added the workaround. No longer need to use sleep.

airgun/entities/cloud_inventory.py
```diff
@@ -11,33 +13,59 @@ class CloudInventoryEntity(BaseEntity):
...
+
+    def obfuscate_ips(self, value):
+        """ Enable/Disable Obfuscate host ipv4 addresses setting"""
+        view = self.navigate_to(self, 'All')
+        view.obfuscate_ips.fill(value)
+
+    def exclude_packages(self, value):
+        """ Enable/Disable Exclude Packages setting"""
+        view = self.navigate_to(self, 'All')
+        view.exclude_packages.fill(value)
```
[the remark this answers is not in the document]
jameerpathan111: updated pr.

airgun/browser.py
```diff
@@ -658,13 +658,14 @@ def get_downloads_list(self):
...
-              .filter(e => e.state === "COMPLETE")
-              .map(e => e.file_url || e.fileUrl);
-        """
+        time.sleep(3)
+        script = (
+            'return document.querySelector("downloads-manager")'
+            '.shadowRoot.querySelector("#downloadsList")'
+            '.items.filter(e => e.state === "COMPLETE")'
+            '.map(e => e.filePath || e.file_path || e.fileUrl || e.file_url);'
         )
```
jameerpathan111, line 667: the old script was not working for chrome 90, so fixed this as well.

airgun/entities/cloud_inventory.py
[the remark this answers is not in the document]
jameerpathan111: Using `view.inventory_list.generating.terminal.read()` instead of `view.inventory_list.status` doesn't seem good. So reverted the fix. Also noticed that the `STATUS` locator was wrong, so fixed it as well.

jameerpathan111:
@mshriver @omkarkhatavkar @tpapaioa updated the pr. Need your review.

mshriver:
@jameerpathan111 is this change in 6.9.z or only in 6.10?

jameerpathan111:
> @jameerpathan111 is this change in 6.9.z or only in 6.10?

It's for 6.9.z, haven't tried for 6.10 yet.

airgun/entities/cloud_inventory.py
```diff
@@ -11,26 +13,31 @@ class CloudInventoryEntity(BaseEntity):
...
 
     def generate_report(self, entity_name):
         view = self.navigate_to(self, 'All')
-        view.searchbox.search(entity_name)
-        entity_item = next(view.inventory_list[entity_name])
-        entity_item.browser.click(entity_item.generating.restart, ignore_ajax=True)
-        wait_for(lambda: entity_item.status == 'idle', timeout=180, delay=1, logger=view.logger)
+        view.inventory_list.toggle(entity_name)
+        view.browser.click(view.inventory_list.generating.restart, ignore_ajax=True)
+        # Fix
```
mshriver, line 26: Please provide some more depth here - will this comment help you when you read it 6 months from now?  Is this a fix, or does this need to be fixed?

What is this fixing? Why must we sleep here, even though we're using `wait_for`? What about the page, browser, or Satellite UI is causing the need to have a static sleep here? (5 seconds its ages to statically wait in an automation pipeline) What have you tried to work around it that didn't work?

The comment should answer these types of questions, so that the next person reading it can understand what's going on - and maybe on day fix it.

Sometimes providing all the context in-line gets too verbose, in which case its good to add more explanation to the commit. Someone reading this in the future has a reasonable chance of knowing to read the commit that `git blame` identifies for this line.

airgun/entities/cloud_inventory.py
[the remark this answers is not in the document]
mshriver: Can we just use the `handle_exception` kwarg for `wait_for` so that the exception doesn't stop the loop?

mshriver (commented):
Looking good, just want to discuss the static sleep a bit more and get some more context around an in-line comment.

airgun/entities/cloud_inventory.py
jameerpathan111: You are right should have mentioned the issue before, will keep this in mind for future issues. here's what the problem is:
When we click on `restart` button then the process of generating and uploading a report gets started. There are two icons one for generating tab and other for uploading tab. They indicate users whether process of uploading or generating report is finished or in progress. So we were using those icon to wait till reports are generated and uploaded based on `entity_item.status` can result in `uploading`, `generating` and `idle` state. But the issue is that those icon don't change their state immediately after we start report generation process(clicking on `restart` button), which makes automation think that the process is finished(`idle`), which is not correct. That's why we have added sleep to wait for icons to change their state(`generating` or `uploading`). Also the process of generating/uploading reports takes some time either way, so I thought we can wait for few seconds.

airgun/entities/cloud_inventory.py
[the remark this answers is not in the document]
jameerpathan111: added.

tpapaioa:
Can you move the description from the commit msg into the code itself?

jameerpathan111:
> Can you move the description from the commit msg into the code itself?

updated pr.

airgun/entities/cloud_inventory.py
mshriver: Thanks @jameerpathan111

mshriver (approved):

tpapaioa:
I spent some time looking into the page / status update issues, and it looks like the `sleep` calls can be removed if we simply increase the default `timeout` value in `AirgunBrowserPlugin.ensure_page_safe()` from `'30s'` to, e.g., `'60s'`. The cloud inventory generation and upload tabs sometimes take longer than 30 seconds to finishing loading and the associated spinner element to disappear. I'll create a new issue/PR for this.

PR #569, opened 2021-05-07
Fix default http proxy name value

tpapaioa:
This PR fixes the behavior of `RepositoryEntity.global_default_http_proxy`. When a global default http proxy has been configured under `Administer > Settings > Content`, the setting is displayed as:

`proxy_name (proxy_url)`

and `RepositoryEntity.global_default_http_proxy` should return a string matching what the corresponding value will be in the `HTTP Proxy Policy` dropdown when creating a new custom repository:

`Global Default (proxy_name)`

Previously, the method was using the full setting value, like this:

`Global Default (proxy_name (proxy_url))`

Now it uses a regular expression to extract `proxy_name` from the setting and return the correct string.

### DIFF
--- a/airgun/entities/repository.py
+++ b/airgun/entities/repository.py
@@ -1,3 +1,5 @@
+import re
+
 from wait_for import wait_for
 
 from airgun.entities.base import BaseEntity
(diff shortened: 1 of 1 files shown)


JacobCallahan (approved):

latran (approved):
ACK

PR #576, opened 2021-05-26
Add support for Red Hat Insights page

jameerpathan111:
This PR adds airgun support for RH Insights page.
It allows to sync, search and remediate insights recommendations.
I will provide some kind of test result soon.

### DIFF
--- a/airgun/entities/cloud_insights.py
+++ b/airgun/entities/cloud_insights.py
@@ -1,5 +1,3 @@
-from wait_for import wait_for
-
 from airgun.entities.base import BaseEntity
 from airgun.navigation import NavigateStep
 from airgun.navigation import navigator
@@ -33,7 +31,7 @@ class CloudInsightsEntity(BaseEntity):
         """Sync RH Cloud - Insights recommendations."""
         view = self.navigate_to(self, 'All')
         view.start_hits_sync.click()
-        view.browser.plugin.ensure_page_safe(timeout='60s')
+        self.browser.plugin.ensure_page_safe(timeout='60s')
 
     def read(self, widget_names=None):
         """Read all values for created activation key entity"""
@@ -45,7 +43,7 @@ class CloudInsightsEntity(BaseEntity):
         view = self.navigate_to(self, 'Token')
         view.rhcloud_token.fill(value)
         view.save_token.click()
-        view.browser.plugin.ensure_page_safe(timeout='60s')
+        self.browser.plugin.ensure_page_safe(timeout='60s')
 
     def update(self, values):
         """Update RH Cloud - Insights view."""
@@ -55,12 +53,7 @@ class CloudInsightsEntity(BaseEntity):
     def run_job(self):
         """Run remediation job"""
         view = self.navigate_to(self, 'Run')
-        wait_for(
-            lambda: view.is_displayed is True,
-            timeout=20,
-            delay=1,
-            logger=view.logger,
-        )
+        self.browser.plugin.ensure_page_safe(timeout='60s')
         view.submit.click()
 
 
--- a/airgun/entities/cloud_insights.py
+++ b/airgun/entities/cloud_insights.py
@@ -1,3 +1,5 @@
+from wait_for import wait_for
+
 from airgun.entities.base import BaseEntity
 from airgun.navigation import NavigateStep
 from airgun.navigation import navigator
@@ -14,8 +16,7 @@ class CloudInsightsEntity(BaseEntity):
         :param value: text to filter (default: no filter)
         """
         view = self.navigate_to(self, 'All')
-        view.search(value)
-        return view.recommendation_table.read()
+        return view.search(value)
 
     def remediate(self, entity_name):
         """Remediate hosts based on search input."""
@@ -51,7 +52,12 @@ class CloudInsightsEntity(BaseEntity):
     def run_job(self):
         """Run remediation job"""
         view = self.navigate_to(self, 'Run')
-        self.browser.plugin.ensure_page_safe(timeout='60s')
+        wait_for(
+            lambda: view.is_displayed is True,
+            timeout=20,
+            delay=1,
+            logger=view.logger,
+        )
         view.submit.click()
 
 
--- a/airgun/entities/cloud_insights.py
+++ b/airgun/entities/cloud_insights.py
@@ -9,12 +9,13 @@ class CloudInsightsEntity(BaseEntity):
     endpoint_path = '/foreman_rh_cloud/insights_cloud'
 
     def search(self, value):
-        """Search for 'value' and return hostname/recommendation names that match.
+        """Search for 'query' and return hostname/recommendation names that match.
 
         :param value: text to filter (default: no filter)
         """
         view = self.navigate_to(self, 'All')
-        return view.search(value)
+        view.search(value)
+        return view.recommendation_table.read()
 
     def remediate(self, entity_name):
         """Remediate hosts based on search input."""
@@ -26,8 +27,13 @@ class CloudInsightsEntity(BaseEntity):
         view.remediation_window.remediate.click()
         self.run_job()
 
-    def read_all(self, widget_names=None):
-        """Return dict with properties of RH Cloud - Insights."""
+    def sync_hits(self):
+        """Sync RH Cloud - Insights recommendations."""
+        view = self.navigate_to(self, 'All')
+        view.start_hits_sync.click()
+
+    def read(self, widget_names=None):
+        """Read all values for created activation key entity"""
         view = self.navigate_to(self, 'All')
         return view.read(widget_names=widget_names)
 
--- a/airgun/entities/cloud_insights.py
+++ b/airgun/entities/cloud_insights.py
@@ -1,0 +1,26 @@
+from wait_for import wait_for
+
+from airgun.entities.base import BaseEntity
+from airgun.navigation import NavigateStep
+from airgun.navigation import navigator
+from airgun.views.cloud_insights import CloudInsightsView
+
+
+class CloudInsightsEntity(BaseEntity):
+    endpoint_path = '/foreman_rh_cloud/insights_cloud'
+
+    def save_token_sync_hits(self, value):
+        """Update Insights cloud view."""
+        view = self.navigate_to(self, 'All')
+        view.rhcloud_token.fill(value)
+        view.save_token.click()
+
+
+@navigator.register(CloudInsightsEntity, 'All')
+class ShowCloudInsightsView(NavigateStep):
+    """Navigate to main Red Hat Insights page"""
+
+    VIEW = CloudInsightsView
+
+    def step(self, *args, **kwargs):
+        self.view.menu.select('Configure', 'Insights')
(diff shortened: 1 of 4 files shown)


airgun/widgets.py
```diff
@@ -2229,3 +2229,54 @@ def items(self):
 
     def toggle(self, value):
         self.browser.click(self.ITEM.format(value))
+
+
+class InventoryBootstrapSwitch(Widget):
+    """Checkbox-like Switch control, representing On and Off state. But with
```
jameerpathan111, line 2235: This was written by mirek sometime back, I am adding it again because it's required for `insights_sync_switcher`.

airgun/views/cloud_insights.py
```diff
@@ -0,0 +1,89 @@
...
+class RemediationView(Modal):
+    """ Remediation window view"""
+
+    remediate = Pf4Button(locator='//button[text()="Remediate" and @type="submit"]')
+    cancel = Pf4Button(locator='//button[text()="Cancel"]')
+    table = PatternflyTable(
+        locator='.//table[@aria-label="remediations Table"]',
+        column_widgets={
+            'Hostname': Text('./a'),
+            'Recommendation': Text('./a'),
```
omkarkhatavkar, line 26: both locators are same is this working for you?

airgun/views/cloud_insights.py
```diff
@@ -0,0 +1,89 @@
...
+            'Resolution': Text(".//a"),
+            'Reboot Required': Text(".//a"),
```
omkarkhatavkar, line 28: use single quote

airgun/entities/cloud_insights.py
```diff
@@ -0,0 +1,71 @@
...
+from airgun.navigation import navigator
+from airgun.views.cloud_insights import CloudInsightsView
+from airgun.views.job_invocation import JobInvocationCreateView
+
+
+class CloudInsightsEntity(BaseEntity):
+    endpoint_path = '/foreman_rh_cloud/insights_cloud'
+
+    def search(self, value):
+        """Search for 'query' and return hostname/recommendation names that match.
```
omkarkhatavkar, line 12: `Search for 'query' and return matched hostnames/recommendations.`

omkarkhatavkar (approved):
LGTM, some pending comments.

airgun/views/cloud_insights.py
```diff
@@ -0,0 +1,89 @@
+from wait_for import wait_for
+from widgetastic.widget import Checkbox
+from widgetastic.widget import Text
+from widgetastic.widget import TextInput
+from widgetastic.widget import View
+from widgetastic_patternfly4 import Button
+from widgetastic_patternfly4 import Pagination
+from widgetastic_patternfly4.button import Button as Pf4Button
```
mshriver, line 8: This is a duplicate import as the import directly from widgetastic_patternfly4 package.

For PF4 widgets, you should be importing the widget class through `widgetastic_patternfly4.ouia`, and using the component ID (`data-ouia-component-id` attribute in DOM) as the locator.

OUIA compatible PF4 components should construct a standard locator using only the component ID - no further locator xpath should be necessary for ROOT widget locators.

This means for example the remediate/cancel buttons will be created not by passing a locator with xpath for the button text, but by inspecting the DOM for a component ID and using it when instantiating the widget within a view.

swadeley:
Will there be test results?

jameerpathan111:
> Will there be test results?

@swadeley yes, will add test result soon.

jameerpathan111:
@swadeley @mshriver @omkarkhatavkar 
Test result is at https://github.com/SatelliteQE/robottelo/pull/8767 
The test I have written currently fails at teardown step but it's not related to this airgun pr.
Also there's https://github.com/SatelliteQE/airgun/issues/584 issue related to search which I will try to fix separately later.

jameerpathan111:
@swadeley @mshriver @latran can I get your review on this pr?

jaryn (approved):

airgun/views/cloud_insights.py
```diff
@@ -0,0 +1,97 @@
...
+    @property
+    def is_displayed(self):
+        return self.browser.wait_for_element(self.rhcloud_token, exception=False) is not None
+
+
+class RemediationView(Modal):
+    """ Remediation window view"""
+
+    remediate = Button(locator='//button[text()="Remediate" and @type="submit"]')
+    cancel = Button(locator='//button[text()="Cancel"]')
```
mshriver, line 31: There's no need to pass a full locator here, the `Button` widget will build a locator using `normalize-space` when a string is passed, and it will match the button's displayed text. You should be able to use `Button('Cancel')` here.

airgun/views/cloud_insights.py
```diff
@@ -0,0 +1,97 @@
...
+
+    @property
+    def is_displayed(self):
+        return self.browser.wait_for_element(self.rhcloud_token, exception=False) is not None
+
+
+class RemediationView(Modal):
+    """ Remediation window view"""
+
+    remediate = Button(locator='//button[text()="Remediate" and @type="submit"]')
```
mshriver, line 30: Are there multiple remediate buttons on the page? Why the additional xpath component for `@type`?

airgun/views/cloud_insights.py
mshriver: column_widgets should apply that locator only in the context of the column with the name matching the key from the dict, so it shouldn't cross locator text fields in other columns.

airgun/views/cloud_insights.py
```diff
@@ -0,0 +1,97 @@
...
+            'Hostname': Text('./a'),
+            'Recommendation': Text('./a'),
+            'Resolution': Text('.//a'),
+            'Reboot Required': Text('.//a'),
+        },
+    )
+
+    @property
+    def is_displayed(self):
+        return self.browser.wait_for_element(self.title, exception=False) is not None
```
mshriver, line 44: I think its a better pattern to use wt.core `Widget.wait_displayed` method against `self.title` widget. `self.title.wait_displayed()`

This uses the widget's `is_displayed` method, which in turn will use `Browser.move_to_element().is_displayed`. It will handle `NoSuchElementException` and `MoveTargetOutOfBoundsException`

airgun/views/cloud_insights.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+class CloudInsightsView(BaseLoggedInView, SearchableViewMixin):
+    """Main RH Cloud Insights view."""
+
+    title = Text('//h1[text()="Red Hat Insights"]')
+    insights_sync_switcher = InventoryBootstrapSwitch(class_name='insights_sync_switcher')
+    start_hits_sync = Button(locator='//button[text()="Start recommendations sync"]')
```
mshriver, line 52: Same recommendations to use `Button` class locator construction and just define these as `Button('Start recommendations sync')` and `Button('Remediate')`

airgun/views/cloud_insights.py
```diff
@@ -0,0 +1,97 @@
...
+        },
+    )
+    select_all_hits = Button(locator='//button[text()="Select recommendations from all pages"]')
+    clear_hits_selection = Button(locator='//button[text()="Clear Selection"]')
+    pagination = Pagination()
+    remediation_window = View.nested(RemediationView)
+
+    @property
+    def is_displayed(self):
+        return self.browser.wait_for_element(self.title, exception=False) is not None
```
mshriver, line 72: Consider using `self.title.wait_displayed`.

airgun/views/cloud_insights.py
jameerpathan111: Thanks, didn't know that. Updated the pr.

airgun/views/cloud_insights.py
jameerpathan111: Updated the pr.

airgun/views/cloud_insights.py
```diff
@@ -0,0 +1,97 @@
...
+        :rtype: list
+        """
+        if not hasattr(self.__class__, 'table'):
+            raise AttributeError(
+                f'Class {self.__class__.__name__} does not have attribute "table". '
+                'SearchableViewMixin only works with views, which have table for results. '
+                'Please define table or use custom search implementation instead'
+            )
+        self.searchbox.search(query)
+        wait_for(
```
mshriver, line 90: Can `self.table.wait_displayed` not be used here? Is it raising exceptions that wait_displayed doesn't catch?

airgun/widgets.py
```diff
@@ -2238,3 +2238,54 @@ def items(self):
...
+    def __locator__(self):
+        return f"//div[@class='{self.class_name}']/div"
+
+    @property
+    def selected(self):
+        classes = self.browser.classes(self)
+        return 'bootstrap-switch-on' in classes
+
+    @property
+    def is_displayed(self):
```
mshriver, line 2267: You shouldn't have to implement this. The base widget class already has it.

https://github.com/RedHatQE/widgetastic.core/blob/master/src/widgetastic/widget/base.py#L506

airgun/widgets.py
```diff
@@ -2238,3 +2238,54 @@ def items(self):
...
+    """
+
+    ON_TOGGLE = ".//span[contains(@class, 'bootstrap-switch-handle-on')]"
+    OFF_TOGGLE = ".//span[contains(@class, 'bootstrap-switch-handle-off')]"
+
+    def __init__(self, parent, class_name, **kwargs):
+        Widget.__init__(self, parent, logger=kwargs.pop('logger', None))
+        self.class_name = class_name
+
+    def __locator__(self):
```
mshriver, line 2258: Consider instead using `ParametrizedLocator` and defining `ROOT = ParametrizedLocator("//div[@class='{@class_name}']/div)"`

airgun/widgets.py
```diff
@@ -2238,3 +2238,54 @@ def items(self):
...
+        """
+        locator = self.ON_TOGGLE
+        if not self.selected:
+            locator = self.OFF_TOGGLE
+        return self.browser.element(locator=locator)
+
+    def fill(self, value):
+        value = bool(value)
+        current_value = self.selected
+        if value == current_value:
```
mshriver, line 2284: Consider flattening this a bit, since value and current_value are never re-used.

airgun/entities/cloud_insights.py
```diff
@@ -0,0 +1,91 @@
...
+    def update(self, values):
+        """Update RH Cloud - Insights view."""
+        view = self.navigate_to(self, 'All')
+        view.fill(values)
+
+    def run_job(self):
+        """Run remediation job"""
+        view = self.navigate_to(self, 'Run')
+        wait_for(
+            lambda: view.is_displayed is True,
```
mshriver, line 59: Same comment here as below - is there a reason `view.wait_displayed` can't be used here?

mshriver (changes requested):
Couple remaining comments, otherwise this entity+view+widget looks good, well done @jameerpathan111

airgun/entities/cloud_insights.py
jameerpathan111: updated the pr.

airgun/widgets.py

airgun/widgets.py

mshriver (approved):

Issue #578, opened 2021-05-28
Default ensure_page_safe timeout is too short for some pages


tpapaioa:
https://github.com/SatelliteQE/airgun/pull/568 introduced some `sleep` calls to deal with errors like the following on Cloud Inventory pages:

```
2021-05-27 15:20:58 - widgetastic_null - DEBUG - [CloudInventoryListView/inventory_list]: click started
2021-05-27 15:20:58 - widgetastic_null - DEBUG - click: <airgun.views.cloud_inventory.InventoryItemsView object at 0x7f0e92a86a90>
2021-05-27 15:20:58 - widgetastic_null - DEBUG - move_to_element: <airgun.views.cloud_inventory.InventoryItemsView object at 0x7f0e92a86a90>
2021-05-27 15:21:28 - widgetastic_null - ERROR - [CloudInventoryListView/inventory_list]: An exception happened during click call (elapsed 29803 ms)
2021-05-27 15:21:28 - widgetastic_null - ERROR - [CloudInventoryListView/inventory_list]: Could not do 'function _check()' at /robottelo/venv/lib64/python3.8/site-packages/widgetastic/browser.py:75 in time
    Traceback (most recent call last):
[...]
      File "/robottelo/venv/lib64/python3.8/site-packages/widgetastic/browser.py", line 1085, in elements
        return self._browser.elements(
      File "/robottelo/venv/lib64/python3.8/site-packages/widgetastic/utils.py", line 695, in wrap
        return method(*args, **kwargs)
      File "/robottelo/venv/lib64/python3.8/site-packages/widgetastic/browser.py", line 302, in elements
        self.plugin.ensure_page_safe()
      File "/robottelo/venv/lib64/python3.8/site-packages/airgun/browser.py", line 588, in ensure_page_safe
        super().ensure_page_safe(timeout)
      File "/robottelo/venv/lib64/python3.8/site-packages/widgetastic/browser.py", line 84, in ensure_page_safe
        wait_for(_check, timeout=timeout, delay=0.2, very_quiet=True)
      File "/robottelo/venv/lib64/python3.8/site-packages/wait_for/__init__.py", line 237, in wait_for
        raise TimedOutError(timeout_msg)
    wait_for.TimedOutError: Could not do 'function _check()' at /robottelo/venv/lib64/python3.8/site-packages/widgetastic/browser.py:75 in time
```

[...]

PR #579, opened 2021-06-01
Add support for configure_cloud_connector and sync_inventory_status

jameerpathan111:
This PR adds support for `configure_cloud_connector`, `sync_inventory_status` and tries to address sleep issue #578

### DIFF
--- a/airgun/entities/cloud_inventory.py
+++ b/airgun/entities/cloud_inventory.py
@@ -21,58 +21,16 @@ class CloudInventoryEntity(BaseEntity):
         """Configure Cloud Connector"""
         view = self.navigate_to(self, 'All')
         view.cloud_connector.click()
-        wait_for(
-            lambda: view.cloud_connector_status.is_displayed is True,
-            handle_exception=True,
-            timeout=20,
-            delay=1,
-            logger=view.logger,
-        )
-        wait_for(
-            lambda: view.cloud_connector_status.is_displayed is False,
-            handle_exception=True,
-            timeout=180,
-            delay=1,
-            logger=view.logger,
-        )
 
     def sync_inventory_status(self):
         """Sync Inventory status"""
         view = self.navigate_to(self, 'All')
         view.sync_status.click()
-        wait_for(
-            lambda: view.sync_status_disabled.is_displayed is True,
-            handle_exception=True,
-            timeout=20,
-            delay=1,
-            logger=view.logger,
-        )
-        wait_for(
-            lambda: view.sync_status.is_displayed is True,
-            handle_exception=True,
-            timeout=20,
-            delay=1,
-            logger=view.logger,
-        )
 
     def generate_report(self, entity_name):
         view = self.navigate_to(self, 'All')
         view.inventory_list.toggle(entity_name)
         view.browser.click(view.inventory_list.generating.restart, ignore_ajax=True)
-        wait_for(
-            lambda: view.inventory_list.status == 'generating',
-            handle_exception=True,
-            timeout=20,
-            delay=1,
-            logger=view.logger,
-        )
-        wait_for(
-            lambda: view.inventory_list.status == 'idle',
-            handle_exception=True,
-            timeout=180,
-            delay=1,
-            logger=view.logger,
-        )
 
     def download_report(self, entity_name):
         view = self.navigate_to(self, 'All')
--- a/airgun/entities/cloud_inventory.py
+++ b/airgun/entities/cloud_inventory.py
@@ -31,7 +31,7 @@ class CloudInventoryEntity(BaseEntity):
         wait_for(
             lambda: view.cloud_connector_status.is_displayed is False,
             handle_exception=True,
-            timeout=100,
+            timeout=180,
             delay=1,
             logger=view.logger,
         )
--- a/airgun/entities/cloud_inventory.py
+++ b/airgun/entities/cloud_inventory.py
@@ -1,5 +1,3 @@
-from time import sleep
-
 from wait_for import wait_for
 
 from airgun.entities.base import BaseEntity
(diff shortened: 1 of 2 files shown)


lhellebr:
LGTM.

I understand there are no test results for the new stuff since there are no tests yet but the `generate_report` method isn't new. Can you provide results of some test using it to show this doesn't cause regression?

jameerpathan111:
Test result:
```
$ pytest tests/foreman/ui/test_rhcloud_inventory.py 
platform linux -- Python 3.8.6, pytest-6.2.2, py-1.10.0, pluggy-0.13.1
plugins: ibutsu-1.13, services-2.2.1, cov-2.11.1, mock-3.5.1, reportportal-5.0.7, xdist-2.2.1, forked-1.3.0

tests/foreman/ui/test_rhcloud_inventory.py .                             [100%]
=========== 1 passed, 5 deselected, 22 warnings in 728.74s (0:12:08) ===========
```
@lhellebr I had tried `configure_cloud_connector` and `sync_inventory_status` with following test code:
```
    with session:
        session.organization.select(org_name=org)
        session.cloudinsights.save_token_sync_hits(token)
        timestamp = (datetime.utcnow() - timedelta(minutes=2)).strftime('%Y-%m-%d %H:%M')
        session.cloudinventory.sync_inventory_status()
        wait_for_tasks(
        search_query=(
            'label = InventorySync::Async::InventoryFullSync'
            f' and started_at >= "{timestamp}"'
        ),
        search_rate=15,
        max_tries=10,
        )
        timestamp = (datetime.utcnow() - timedelta(minutes=2)).strftime('%Y-%m-%d %H:%M')
        session.cloudinventory.configure_cloud_connector()
        wait_for_tasks(
            search_query=(
                'Remote action: Configure cloud connector'
                f' and started_at >= "{timestamp}"'
            ),
            search_rate=15,
            max_tries=10,
        )
```
Configuring cloud connector task fails because of [BZ#1966435](https://bugzilla.redhat.com/show_bug.cgi?id=1966435), but I guess that's not relevant here.

airgun/entities/cloud_inventory.py
```diff
@@ -19,27 +17,55 @@ def read(self, entity_name=None, widget_names=None):
...
+            delay=1,
+            logger=view.logger,
+        )
+        wait_for(
+            lambda: view.cloud_connector_status.is_displayed is False,
+            handle_exception=True,
+            timeout=180,
+            delay=1,
+            logger=view.logger,
+        )
```
tpapaioa, line 37: Until we have a good resolution for the `ensure_page_safe` default timeout issue (https://github.com/SatelliteQE/airgun/issues/578), I would suggest adding calls to that method instead of `sleep()` or extra `wait_for()` calls.

E.g., after the `click()`, `view.browser.plugin.ensure_page_safe(timeout='60s')` would wait up to 60s for the spinner and other script / component activity to complete. `ensure_page_safe()` is called automatically before every element lookup, but the default timeout of `30s` isn't enough here.

lhellebr (approved):

airgun/entities/cloud_inventory.py
jameerpathan111: @tpapaioa I tried using `view.browser.plugin.ensure_page_safe(timeout='60s')` and I see no issue in using `configure_cloud_connector` and `sync_inventory_status`. But for `generate_report` and `download_report` I am facing the same issue. i.e it's not waiting for `generate_report` to finish first and goes directly to upload tab.

jameerpathan111: @tpapaioa @mshriver what should we do? should we keep `wait_for`?

tpapaioa: Use whatever logic most accurately determines whether the action has completed successfully. I doubt that waiting for spinners or buttons to change state temporarily, and then wait again for them to change a second time, is the best strategy. In the case of clicking `Configure Cloud Connector`, for example, it starts a job task, so waiting on the completion of that job and returning its output status would probably be the most reliable method. In the case of report generation and upload, there are foreman/dynflow tasks that you can wait on.

jameerpathan111: @tpapaioa makes sense. I completely forgot that we could use `wait_for_tasks` in our tests itself instead of relying on airgun. 
I will simply update my test and wait for tasks to finish.
like this:
```
timestamp = (datetime.utcnow() - timedelta(minutes=2)).strftime('%Y-%m-%d %H:%M')
wait_for_tasks(
    search_query=(
        'label = InventorySync::Async::InventoryFullSync'
        f' and started_at >= "{timestamp}"'
    ),
    search_rate=15,
    max_tries=10,
)
wait_for_tasks(
    search_query=(
        'Remote action: Configure cloud connector'
        f' and started_at >= "{timestamp}"'
    ),
    search_rate=15,
    max_tries=10,
)
wait_for_tasks(
    search_query='label = ForemanInventoryUpload::Async::GenerateReportJob'
                 f' and started_at >= "{timestamp}"',
    search_rate=15,
    max_tries=10,
)
```

jameerpathan111: update the pr.

airgun/views/cloud_inventory.py
```diff
@@ -113,6 +114,8 @@ class CloudInventoryListView(BaseLoggedInView):
     )
     obfuscate_ips = Switch('.//label[@for="rh-cloud-switcher-obfuscate_inventory_ips"]')
     exclude_packages = Switch('.//label[@for="rh-cloud-switcher-exclude_installed_packages"]')
+    cloud_connector = Pf4Button(locator='//button[text()="Configure Cloud Connector"]')
+    sync_status = Pf4Button(locator='//button[text()=" Sync inventory status"]')
```
omkarkhatavkar, line 118: if possible use the contains in the locator rather having adding ` ` extra space

airgun/entities/cloud_inventory.py
```diff
@@ -19,41 +17,25 @@ def read(self, entity_name=None, widget_names=None):
...
-            delay=1,
-            logger=view.logger,
-        )
 
     def download_report(self, entity_name):
         view = self.navigate_to(self, 'All')
         view.inventory_list.toggle(entity_name)
         view.browser.click(view.inventory_list.uploading.download_report, ignore_ajax=True)
-        # Fix for status icons not changing state immediately after restart
-        sleep(5)
```
omkarkhatavkar, line 56: nice! removing the hard sleep

omkarkhatavkar (approved):
ACK, LGTM. pending the small suggestion

airgun/views/cloud_inventory.py
mshriver: Even better, you should be able to use the OUIA component ID only, and not write an xpath locator at all when using PF4 OUIA compatible widgets.

Please import PF4 widgets through `widgetastic_patternfly4.ouia`

mshriver (approved):

PR #585, opened 2021-07-14
Fix for a report template test

ppochop:
A test that tries to generate a report and download it couldn't pass because of the way `generating` was located. I found a workaround for that and because of the different nature of the element I used, I renamed it to `generated`.

```
pytest tests/foreman/ui/test_reporttemplates.py::test_positive_generate_registered_hosts_report
collected 1 item                                                               

tests/foreman/ui/test_reporttemplates.py .                               [100%]
================== 1 passed, 3 warnings in 307.62s (0:05:07) ===================
```

### DIFF
--- a/airgun/views/report_template.py
+++ b/airgun/views/report_template.py
@@ -104,4 +104,3 @@ class ReportTemplateGenerateView(BaseLoggedInView):
             and self.breadcrumb.locations[0] == 'Report Templates'
             and self.breadcrumb.read() == 'Generate a Report'
         )
-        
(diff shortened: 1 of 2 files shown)


airgun/views/report_template.py
```diff
@@ -104,3 +104,4 @@ def is_displayed(self):
             and self.breadcrumb.locations[0] == 'Report Templates'
             and self.breadcrumb.read() == 'Generate a Report'
         )
+
```
mshriver, line 107: From this and the removed `,` in the report_template.py file, it looks like you aren't running pre-commit in your environment. 

Travis is skipping right now, we need to get moved over to a GHA to enforce this.

ppochop: Indeed I wasn't. I've installed the hooks and ran them this time.

mshriver (approved):

jameerpathan111 (approved):
LGTM

PR #588, opened 2021-07-16
Changes needed to fix templatesync tests

ppochop:
`tests/foreman/ui/test_templatesync.py::test_positive_import_templates` now passes after https://github.com/SatelliteQE/robottelo/pull/8777 
`tests/foreman/ui/test_templatesync.py::test_positive_export_templates` unfortunately doesn't on 6.10 due to an automation-unrelated bug (but passes on 6.9)

### DIFF
--- a/airgun/entities/provisioning_template.py
+++ b/airgun/entities/provisioning_template.py
@@ -54,6 +54,14 @@ class ProvisioningTemplateEntity(BaseEntity):
         view.flash.assert_no_error()
         view.flash.dismiss()
 
+    def is_locked(self, entity_name):
+        """Check if provisioning template is locked for editing"""
+        view = self.navigate_to(self, 'All')
+        view.search(entity_name)
+        return "This template is locked for editing." in view.table.row(name=entity_name)[
+            'Locked'
+        ].widget.browser.element('.').get_property('innerHTML')
+
     def update(self, entity_name, values):
         """Update provisioning template"""
         view = self.navigate_to(self, 'Edit', entity_name=entity_name)
--- a/airgun/views/provisioning_template.py
+++ b/airgun/views/provisioning_template.py
@@ -35,6 +35,7 @@ class ProvisioningTemplatesView(BaseLoggedInView, SearchableViewMixin):
         './/table',
         column_widgets={
             'Name': Text('./a'),
+            'Locked': Text('.'),
             'Actions': ActionsDropdown("./div[contains(@class, 'btn-group')]"),
         },
     )
--- a/airgun/views/sync_templates.py
+++ b/airgun/views/sync_templates.py
@@ -33,24 +33,24 @@ class SyncTemplatesView(BaseLoggedInView):
 
     @template.register('Import')
     class ImportTemplates(View):
-        associate = Select(name='associate')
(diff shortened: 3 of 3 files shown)


airgun/entities/provisioning_template.py
```diff
@@ -54,6 +54,14 @@ def unlock(self, entity_name):
         view.flash.assert_no_error()
         view.flash.dismiss()
 
+    def is_locked(self, entity_name):
+        """Check if provisioning template is locked for editing"""
+        view = self.navigate_to(self, 'All')
+        view.search(entity_name)
+        return "This template is locked for editing." in view.table.row(name=entity_name)[
+            'Locked'
+        ].widget.browser.element('.').get_property('innerHTML')
```
mshriver, line 63: Why are you trying to reference innerHTML instead of using the `Text` widget's `read` method?

ppochop: That's because I basically copied this method from the [ReportTemplateEntity](https://github.com/SatelliteQE/airgun/blob/master/airgun/entities/report_template.py#L59) so I just assumed that's the best way. Either way, I tried using `view.table.row(name=entity_name)['Locked'].widget.read()` and that resulted in an empty string.

jameerpathan111: @ppochop you're getting empty string because there's no text content for the `Locked` field(html tag on which it relies).
@mshriver probably because of field @ppochop trying to work with doesn't have text content in the related tag, workaround mentioned in [ReportTemplateEntity](https://github.com/SatelliteQE/airgun/blob/master/airgun/entities/report_template.py#L59)  was used.

jameerpathan111:
I have cross checked the changes in locator values and they seem correct to me.

jameerpathan111 (approved):

airgun/entities/provisioning_template.py
mshriver: Ack - so one of the other approaches that we can use for things like this is to define a `TemplateText` (naming it whatever makes sense) widget that overrides the `read` method of `Text` to return the element's innerHTML property.

That way when we've got one of these widgets, it can still be read, and we don't have to make specific element references against the view.

omkarkhatavkar: @ppochop are you willing to try the approach mentioned by @mshriver ?

omkarkhatavkar (approved):
ACK

airgun/entities/provisioning_template.py
ppochop: @mshriver @omkarkhatavkar Yes, I can look into it. Should I draft this and the robottelo PR and work on it here or should I close this PR and open new ones?

omkarkhatavkar: No need to create a new one, if you are working on the PR make if draft status. that's fine

mshriver: @ppochop its not a required enhancement for this PR, as you're continuing to read the widget just like it is now. But in general we should try to refrain from such customizations when interacting with the widget - if that's what it takes to read the widget, it should be in the widget's `read` method.

mshriver (approved):

PR #593, opened 2021-07-22
GHA for airgun

latran:
Airgun doesn't have GHA for commits and this PR is for implementing that to be like robottelo.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -1,0 +1,38 @@
+# CI stages to execute against Pull Requests
+name: Airgun - CI
+
+on:
+  pull_request:
+    types: ["opened", "synchronize", "reopened"]
+
+env:
+    PYCURL_SSL_LIBRARY: openssl
+
+jobs:
+  codechecks:
+    name: Code Quality
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.8, 3.9]
+    steps:
+      - name: Checkout Airgun
+        uses: actions/checkout@v2
+
+      - name: Set Up Python-${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install Dependencies
+        run: |
+          sudo apt update
+          pip install -U pip
+          pip install -U -r requirements.txt -r requirements-optional.txt
+
+      - name: Pre Commit Checks
+        uses: pre-commit/action@v2.0.0
+
+      - name: Analysis (git diff)
+        if: failure()
+        run: git diff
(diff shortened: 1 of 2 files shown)


.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,53 @@
...
+          cp broker_settings.yaml.example broker_settings.yaml
+
+      - name: Pre Commit Checks
+        uses: pre-commit/action@v2.0.0
+
+      - name: Analysis (git diff)
+        if: failure()
+        run: git diff
+
+      - name: Upload Codecov Coverage
```
JacobCallahan, line 49: With no actual tests, I think skipping code coverage is fine.

JacobCallahan (changes requested):

.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,53 @@
+# CI stages to execute against Pull Requests
+name: Robottelo - CI
```
mshriver, line 2: Your copy-paste is showing :smile:

.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,53 @@
...
+
+on:
+  pull_request:
+    types: ["opened", "synchronize", "reopened"]
+
+env:
+    PYCURL_SSL_LIBRARY: openssl
```
mshriver, line 9: This customization shouldn't be necessary, I'd only add it if the install fails (see below comment)

.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,53 @@
...
+        uses: actions/checkout@v2
+
+      - name: Set Up Python-${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install Dependencies
+        run: |
+          sudo apt update
```
mshriver, line 29: Looks like this is copy-pasta from robottelo GHA, and this block can be simplified a lot.

I think its good to keep the block around to continue testing airgun installation under pip, but could be simplified all the way down to:

```
sudo apt update
pip install -U pip
pip install -Ur requirements.txt
```

mshriver: For a bit more thorough coverage, you could also try installing from `requirements-optional.txt`, or install as `pip install .`, since requirements for airgun are really included in `setup.py`.

.github/workflows/pull_request.yml
latran: Description does say it's implemented to be like robottelo ;)

.github/workflows/pull_request.yml
latran: done

.github/workflows/pull_request.yml
latran: I left it in just in case, if that's ok.

.github/workflows/pull_request.yml
latran: removed.

mshriver (approved):

JacobCallahan (approved):

PR #599, opened 2021-07-29
ansible collection view end entity

pondrejk:
wip, tests pending 
https://bugzilla.redhat.com/show_bug.cgi?id=1989989

### DIFF
--- a/airgun/entities/ansible_collections.py
+++ b/airgun/entities/ansible_collections.py
@@ -1,0 +1,73 @@
+from wait_for import wait_for
+
+from airgun.entities.base import BaseEntity
+from airgun.navigation import NavigateStep
+from airgun.navigation import navigator
+from airgun.views.ansible_collections import AnsibleCollectionsView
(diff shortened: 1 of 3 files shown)


JacobCallahan:
@pondrejk any luck with this?

pondrejk:
just for update, found out breadcrumbs not around on the collections page, I filed a Bz,  I'll see how it evolves, maybe will use a different identifier if thing won't move

lhellebr:
Is this still current?
Maybe link the mentioned BZ here, @pondrejk?

pondrejk:
yeah the bz is still in the new state, target unspecified, so we'll probably need to navigate without breadcrumbs 
https://bugzilla.redhat.com/show_bug.cgi?id=1989989

pondrejk:
Closing, modeling the ansible collection list page is not currently required by any high-prio test scenario and the blocking bz for the breadcrumb issue is going to be auto-closed. We can revisit this should we need it in the future

PR #611, opened 2021-08-31
Fix searchbox and sync recommendation button of Insights page.

jameerpathan111:
In snap 12 some new UI changes were made to Configure > Insights page.
This PR fixes airgun support for searchbox and sync recommendation button on Insights page.

I am using `Keys.ENTER` from `selenium.webdriver.common.keys` because new searchbox doesn't have search button and user need to use enter key to perform search operation.

I have tested the code change with test written in https://github.com/SatelliteQE/robottelo/pull/8767

### DIFF
--- a/airgun/views/cloud_insights.py
+++ b/airgun/views/cloud_insights.py
@@ -1,15 +1,17 @@
+from selenium.webdriver.common.keys import Keys
 from widgetastic.widget import Checkbox
 from widgetastic.widget import Text
 from widgetastic.widget import TextInput
 from widgetastic.widget import View
 from widgetastic_patternfly4 import Button
 from widgetastic_patternfly4 import Pagination
+from widgetastic_patternfly4.dropdown import Dropdown
 from widgetastic_patternfly4.ouia import Modal
 from widgetastic_patternfly4.ouia import PatternflyTable
+from widgetastic_patternfly4.ouia import Switch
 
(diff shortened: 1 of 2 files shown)


airgun/views/cloud_insights.py
```diff
@@ -47,9 +49,9 @@ class CloudInsightsView(BaseLoggedInView, SearchableViewMixin):
     """Main RH Cloud Insights view."""
 
     title = Text('//h1[text()="Red Hat Insights"]')
-    insights_sync_switcher = InventoryBootstrapSwitch(class_name='insights_sync_switcher')
-    start_hits_sync = Button('Start recommendations sync')
+    insights_sync_switcher = Switch('OUIA-Generated-Switch-1')
     remediate = Button('Remediate')
+    insights_dropdown = Dropdown(locator='.//div[contains(@class, "title-dropdown")]')
```
mshriver, line 54: No OUIA attribute for the div element here?  No need to import the OUIA version of the widget if we're not using the relevant locator.

mshriver (approved):

airgun/views/cloud_insights.py
mshriver: @jameerpathan111 sorry, I read your import wrong, you can ignore this comment. Thanks for reaching out.

omkarkhatavkar (approved):
ACK

Issue #608, opened 2021-08-20
xpath locator is unable to find an element, the element title changes once the mouse hovers over the element


tstrych:
Most of the times it's only <a> element.
To fix this we can use the title, before we hover over the element. 
The title element `@title` instead of `@data-original-title` should be used in elements.

[BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1996035) connected for more information.

hkx303:
Hey, I also found several virt-who plugin cases also failed due to this issue, do we have some workarounds now?

tstrych:
Not yet, but I have planned it for this sprint.  I will be probably changing the locators. I will ping you once I made a progress. 
I suppose for virt-who we also use airgun, is that right @hkx303  ? 
Can you mention in this issue where the problem occurs?

hkx303:
Yes, we also use airgun for the UI testing.
Take case test_positive_debug_option as an example:
Faild log: https://main-jenkins-csb-virtwhoqe.apps.ocp4.prod.psi.redhat.com/job/Virtwho-Plugin-Runtest/7/testReport/junit/tests.foreman.virtwho.ui.test_esx/TestVirtwhoConfigforEsx/test_positive_debug_option/
It occurs when we finished the step for creating virt-who configration, and then we need to  `Checking if already at All` ->  `I'm not at All`, so we need to select the menu to navigate to `Virt-whoconfigure ALL View`,(Selecting ['Infrastructure', 'Virt-who configurations'] in navigation) and you can see the error:
`ERROR    navmazing_null:__init__.py:281 NAVIGATE: Got an error Could not do 'function other_div_displayed()' at /usr/local/lib/python3.8/site-packages/widgetastic_patternfly/__init__.py:583 in time`

we can see a lot of debug info about : 
`DEBUG    widgetastic_null:browser.py:707 css classes for './ul/li[a[span[normalize-space(.)="Infrastructure"]]]' => {'active', 'list-group-item', 'secondary-nav-item-pf'}`
But it should be `class=secondary-nav-item-pf active list-group-item`if we mouse over the Infrastructure...

tstrych:
thanks @hkx303, 

my workaround is simple as you can see,  just added or locator to catch the button before and after you hover over it
my issues were connected mainly to title element as written in description, seems like you are facing something a bit different :) 

we can schedule some call and look at it if you want and I will try to help you :smile:

tstrych:
[...]

PR #615, opened 2021-09-10
hover over issue resolved

tstrych:
connected to https://bugzilla.redhat.com/show_bug.cgi?id=1996035
and with issue #608 

the problem is we can't have only one of these locators, because sometimes we use the mouse to hover over and sometimes not, so just fixing is this way, even though I think it's not the best solution but definitely with the least effort possible. 

this is not fixing the issue,  just showing the way how to workaround it, issue can occur also with other elements


omkarkhatavkar (approved):
ACK, Please update this PR with test result

mshriver (approved):

PR #617, opened 2021-09-17
fixing error 'NoneType' object has no attribute 'replace'

omkarkhatavkar:
### PR Description

We were missing to pass the `component id` which was started failing the test for the with **`'NoneType' object has no attribute 'replace'`**   

This is a experiment PR to see if this actually fix on CI, seems to be working locally fine.

### DIFF
--- a/airgun/widgets.py
+++ b/airgun/widgets.py
@@ -790,13 +790,9 @@ class ValidationErrors(Widget):
         do_not_read_this_widget()
 
 
-class ContextSelector(OUIAContextSelector):
-    CURRENT_ORG = (
-        '//div[@id="organization-dropdown"]//span[@class="pf-c-context-selector__toggle-text"]'
-    )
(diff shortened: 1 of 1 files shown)


omkarkhatavkar:
The experiment seems to be working now see the PRT job result for 292 and compare same with 291

jameerpathan111 (approved):
See PRT job 292 and 293 for results

mshriver (approved):

mshriver:
Approved, but I think the fix is a bit incomplete.

Changing the inheritance and defining those two class attributes as widgets changes the context here - this is now a View, not just a widget.

The current references within the class to `CURRENT_ORG` and `CURRENT_LOC` are assuming its a string, not a widget. Several function calls could be updated to use it as a widget, and the `read` method for this view can likely now be implemented.

PR #618, opened 2021-09-21
Fix error 'NoneType' object has no attribute 'replace' for insights

jameerpathan111:


jameerpathan111:
See PRT job 303 for test result.

omkarkhatavkar (approved):
ACK

jameerpathan111:
Test result:

```
15:06:36  + py.test tests/foreman/ui/test_rhcloud_insights.py --junit-xml=sat-results.xml -o junit_suite_name=sat-result
15:06:51  ============================= test session starts ==============================
15:06:51  collected 1 item
15:06:51  
15:20:50  tests/foreman/ui/test_rhcloud_insights.py .                              [100%]
15:20:50  ================== 1 passed, 4 warnings in 835.28s (0:13:55) ===================
```

devendra104 (approved):

mshriver:
I don't think changing these imports is the right course of action. @omkarkhatavkar already resolved this in https://github.com/SatelliteQE/airgun/pull/617 - the locators were wrong for use with OUIA.

airgun/views/cloud_insights.py
```diff
@@ -3,12 +3,12 @@
 from widgetastic.widget import Text
 from widgetastic.widget import TextInput
 from widgetastic.widget import View
-from widgetastic_patternfly4 import Button
-from widgetastic_patternfly4 import Pagination
+from widgetastic_patternfly4.button import Button
```
mshriver, line 6: This doesn't actually change the imported class. 
https://github.com/RedHatQE/widgetastic.patternfly4/blob/master/src/widgetastic_patternfly4/__init__.py#L4

mshriver:
A bit hard to tell without a stack trace for the exception to reference, but I think your problem is actually coming from the use of `self.title.wait_displayed` on the `Modal` based view. `Modal.title` returns text, not a widget that you can call wait_displayed against.

https://github.com/RedHatQE/widgetastic.patternfly4/blob/ad70ac0c93b5206dfe0e5581d3513220cb394388/src/widgetastic_patternfly4/modal.py#L26

https://github.com/SatelliteQE/airgun/blob/master/airgun/views/cloud_insights.py#L45

airgun/views/cloud_insights.py
jameerpathan111: yes, it doesn't. I changed it to just make code look good/consistent.

jameerpathan111:
> A bit hard to tell without a stack trace for the exception to reference, but I think your problem is actually coming from the use of `self.title.wait_displayed` on the `Modal` based view. `Modal.title` returns text, not a widget that you can call wait_displayed against.
> 
> https://github.com/RedHatQE/widgetastic.patternfly4/blob/ad70ac0c93b5206dfe0e5581d3513220cb394388/src/widgetastic_patternfly4/modal.py#L26
> 
> https://github.com/SatelliteQE/airgun/blob/master/airgun/views/cloud_insights.py#L45

@mshriver I tried with other widget and I am still facing the same issue. Can you help with this?

airgun/views/cloud_insights.py
mshriver: I think this is going the wrong way with the consistency.

mshriver (changes requested):
Debugged and discussed offline - the real problem with this view is OUIA based `RemediateView` class for the modal doesn't have an `OUIA_ID` set, or `component_id` passed on instantiation.

@jameerpathan111 is testing either adding the class attribute, or changing up the inclusion pattern for the modal to pass it.

jameerpathan111:
@mshriver I have updated the pr and tried change locally.

There's an issue with running ansible job and because of which I can't provide test result.
See BZ#2010863 for more info.

PR #620, opened 2021-09-30
handling the browser factory using the webkaifuku

omkarkhatavkar:
### PR Description 

- Addition of Web-Kaifuku 
- Removing dead code for the Browser Factory
- Removing dead code form sauce lab, docker selenium and adding them in web-kaifuku repo
- Addition of ui.yaml file for configuration (https://github.com/SatelliteQE/robottelo/pull/9015)       
- CI PR : 399

### DIFF
--- a/airgun/browser.py
+++ b/airgun/browser.py
@@ -9,44 +9,18 @@ import urllib
 from datetime import datetime
 from urllib.parse import unquote
 
-import selenium
-from fauxfactory import gen_string
+import yaml
+from attrdict import AttrDict
 from selenium import webdriver
 from wait_for import wait_for
+from webdriver_kaifuku import BrowserManager
 from widgetastic.browser import Browser
 from widgetastic.browser import DefaultPlugin
 
 from airgun import settings
 
-try:
-    import docker
-except ImportError:
-    # Let it fail later if not installed
-    docker = None
-
-try:
-    import sauceclient
-except ImportError:
-    # Optional requirement, airgun will report results back to saucelabs if
-    # installed
-    sauceclient = None
-
 
 LOGGER = logging.getLogger(__name__)
-
-
-class DockerBrowserError(Exception):
-    """Indicates any issue with DockerBrowser."""
-
-
-def _sauce_ondemand_url(saucelabs_user, saucelabs_key):
-    """Get sauce ondemand URL for a given user and key.
-
-    :param str saucelabs_user: saucelabs username
-    :param str saucelabs_key: saucelabs access key
-    :return: string representing saucelabs ondemand URL
-    """
-    return f'http://{saucelabs_user}:{saucelabs_key}@ondemand.saucelabs.com:80/wd/hub'
 
 
 class SeleniumBrowserFactory:
@@ -83,12 +57,9 @@ class SeleniumBrowserFactory:
         values.
 
         :param str optional provider: Browser provider name. One of
-            ('selenium', 'docker', 'saucelabs', 'remote'). If none specified -
+            ('selenium', 'remote'). If none specified -
             :attr:`settings.selenium.browser` is used.
-        :param str optional browser: Browser name. One of ('chrome', 'firefox',
-            'ie', 'edge', 'phantomjs'). Not required for ``docker``
-            provider as it currently supports firefox only. If none specified -
-            :attr:`settings.selenium.webdriver` is used.
+        :param str optional browser: Browser name. One of ('chrome', 'firefox').
         :param str optional test_name: Name of the test using this factory. It
             is useful for `saucelabs` provider to update saucelabs job name, or
             for `docker` provider to create container with meaningful name, not
@@ -100,6 +71,7 @@ class SeleniumBrowserFactory:
         """
         self.provider = provider or settings.selenium.browser
         self.browser = browser or settings.selenium.webdriver
+        self.web_kaifuku = AttrDict(yaml.safe_load(settings.webkaifuku.config))
         self.test_name = test_name
         self._session = session_cookie
         self._docker = None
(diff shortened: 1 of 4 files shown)


airgun/browser.py
```diff
@@ -115,16 +86,12 @@ def get_browser(self):
...
         elif self.provider == 'remote':
             return self._get_remote_browser()
         else:
             raise ValueError(
                 f'"{self.provider}" browser is not supported. '
-                f'Please use one of {("selenium", "saucelabs", "docker", "remote")}'
+                f'Please use one of {("selenium", "remote")}'
             )
 
     def post_init(self):
```
latran, line 97: Not sure what the purpose of this after removing everything.  Should it also be removed?

airgun/browser.py
```diff
@@ -194,78 +148,18 @@ def _get_selenium_browser(self):
...
-            self._docker._capabilities = webdriver.DesiredCapabilities.CHROME.copy()
-            self._docker._capabilities.update({'args': 'start-maximized'})
-        elif self.browser == 'firefox':
-            self._docker._image = 'selenium/standalone-firefox'
-            self._docker._capabilities = webdriver.DesiredCapabilities.FIREFOX.copy()
-        else:
-            raise ValueError(
-                f'"{self.browser}" webdriver in docker container is currently not'
-                f'supported. Please use one of {("chrome", "firefox")}'
+                f'Please use one of {("chrome", "firefox")}'
```
latran, line 161: It looks so nice to use just 2 now.

latran (commented):
Pending comment.  Can you also provide a sample run on a UI file.  It can be one just to show that it works.

omkarkhatavkar:
> Pending comment. Can you also provide a sample run on a UI file. It can be one just to show that it works.
@latran for this you can take a look at CI PR : 399

airgun/browser.py
omkarkhatavkar: Yes we can add later once we add that support in webkaifuku

setup.py
```diff
@@ -25,6 +25,7 @@
         'widgetastic.core',
         'widgetastic.patternfly',
         'widgetastic.patternfly4',
+        'webdriver-kaifuku',
```
Gauravtalreja1, line 28: Optional: Alphabetically this should be above wt.core and wt.pf/4

latran (approved):
ACK

airgun/browser.py
```diff
@@ -274,256 +168,16 @@ def _get_remote_browser(self):
 
         Note: should not be called directly, use :meth:`get_browser` instead.
         """
-        self._webdriver = webdriver.Remote(
-            command_executor=settings.selenium.command_executor,
-            desired_capabilities=self._get_webdriver_capabilities(),
-        )
+        desired_capabilities = self.web_kaifuku['webdriver_options']['desired_capabilities']
+        desired_capabilities.update({'name': self.test_name})
```
mshriver, line 172: There should be a setter method for this property, rather than updating the dictionary by reference here.

airgun/browser.py
```diff
@@ -100,6 +70,7 @@ def __init__(
         """
         self.provider = provider or settings.selenium.browser
         self.browser = browser or settings.selenium.webdriver
+        self.web_kaifuku = yaml.load(settings.webkaifuku.config)
```
mshriver, line 73: Here is where I would look to consolidate the settings for airgun browser and webdriver in airgun.

Does airgun need a separate settings specification for browser and webdriver, with the same options repeated again in webkaifuku?  I don't think it does.

If we want to isolate this integration for airgun, I would not include a webkaifuku yaml block at all. Airgun maintains its own yaml spec for browser and webdriver settings. It then maps all of those settings to the kaifuku spec to call `BrowserManager.from_conf`.

If we're not as concerned about isolating the integration, then we consolidate on kaifuku yaml spec for browser and webdriver. Within airgun we read the relevant fields from this spec, instead of having a separate and unique yaml structure for the necessary options.

I think you should evaluate how `self.provider`, `self.browser`, and `settings.selenium` are used in airgun to help guide the decision of which of the above options best suits us.

I'm inclined to consolidate these settings on kaifuku's yaml spec.  Right now kaifuku isn't using any yaml validation tools, like dynaconf, to validate the structure of the dictionary passed. It's an enhancement for kaifuku as a library that we would want to discuss with Ronny.  

It's also a deeper enhancement for airgun - moving to using dynaconf for settings validation and specification, instead of ConfigParser. @latran @JacobCallahan

Maybe this is outside of the scope of kaifuku integration for now, but I would write it as an immediate issue/TODO against this settings repetition for webdriver and browser options.

airgun/browser.py
mshriver: I'd just drop the whole function for now, unless you're going to put some statements in it.

IIRC doing anything after starting the browser instance is not something Ronny wanted included in the kaifuku design. I wouldn't assume that if such support was added, it would be done via a function called `post_init`.

mshriver:
@omkarkhatavkar this is a great start to the integration with Kaifuku!

I think you can take it a bit further though, with integration at the airgun session, removing the factories from airgun. Kaifuku provides a browser factory, not just the manager:
https://github.com/RonnyPfannschmidt/webdriver_kaifuku/blob/master/src/webdriver_kaifuku/__init__.py#L57

That might be a big step for one PR though, since the airgun factories may have some relevant custom code that's necessary for now.  We would also want to make sure we announce this change to airgun consumers other than robottelo, like the Insights QE community.

airgun/browser.py
JacobCallahan: I also agree that airgun's settings should be moved over to dynaconf, but this is definitely outside the scope of this change.

omkarkhatavkar:
> @omkarkhatavkar this is a great start to the integration with Kaifuku!
> 
> I think you can take it a bit further though, with integration at the airgun session, removing the factories from airgun. Kaifuku provides a browser factory, not just the manager: https://github.com/RonnyPfannschmidt/webdriver_kaifuku/blob/master/src/webdriver_kaifuku/__init__.py#L57
> 
> That might be a big step for one PR though, since the airgun factories may have some relevant custom code that's necessary for now. We would also want to make sure we announce this change to airgun consumers other than robottelo, like the Insights QE community.

@mshriver you meant by passing the webkaifuku yaml object itself as part of airgun session and workign with that spwn the browser as rather keeping separate setting object ?

PR #624, opened 2021-10-01
updated the widget for activation key

omkarkhatavkar:


airgun/views/hostgroup.py
```diff
@@ -17,6 +17,18 @@
 from airgun.widgets import RadioGroup
 
 
+class ActivationKeyDropDown(ActionsDropdown):
+    dropdown = Text('.//*[self::div]')
+
+    @property
+    def items(self):
+        """Returns a list of all dropdown items as strings."""
+        self.dropdown.click()
```
mshriver, line 26: I think you should use `self.open()` here, to include the check against `is_open` before clicking the dropdown.

airgun/views/hostgroup.py
```diff
@@ -17,6 +17,18 @@
 from airgun.widgets import RadioGroup
 
 
+class ActivationKeyDropDown(ActionsDropdown):
+    dropdown = Text('.//*[self::div]')
```
mshriver, line 21: This makes the dropdown widget locator VERY broad, when the `ActionsDropdown` locator is currently specific for a few xpath attributes with some variation.

Is the activation key page component missing all of those attributes used in the `ActionsDropdown.dropdown` locator? (Below for reference)

```
dropdown = Text(
        ".//*[self::a or self::button][contains(@class, 'dropdown-toggle') or "
        "contains(@ng-click, 'toggleDropdown')][contains(@class, 'btn')]"
        "[*[self::span or self::i][contains(@class, 'caret')]]"
    )
```

airgun/views/hostgroup.py
```diff
@@ -118,9 +130,7 @@ class organizations(SatTab):
     @View.nested
     class activation_keys(SatTab):
         TAB_NAME = 'Activation Keys'
-        activation_keys = TextInput(
-            locator=".//foreman-react-component[contains(@data-props, 'kt_activation_keys')]"
-        )
+        activation_keys = ActivationKeyDropDown(locator=".//div[@class='rbt clearfix open']")
```
mshriver, line 133: Do you want to explicitly restrict the locator to these three classes, or the div that includes these three classes? We don't want to write locators that will be highly sensitive to page style changes.

airgun/views/hostgroup.py
mshriver: The video helps immensely, I see what you mean.

airgun/views/hostgroup.py
omkarkhatavkar: I will try to see if I can add more solid locator here.

airgun/views/hostgroup.py
omkarkhatavkar: I will try to see if I can build the more solid locator

omkarkhatavkar: @mshriver .//*[self::div] looks inside the drop down locator, I did not find anything that rock solid

airgun/views/hostgroup.py
omkarkhatavkar: updated !

jameerpathan111 (approved):

PR #625, opened 2021-10-05
hardware modul, locator fix, plus dialog fix

tstrych:
- endpoint changed 
- dialog introduced
- when it is not possible to delete the hardware model the dialog is not cancelled, so it's need to be cancelled manually


airgun/entities/hardware_model.py
```diff
@@ -37,13 +37,22 @@ def update(self, entity_name, values):
...
+
+        cancel_dialog - dialog needs to be cancelled, if action expects an error
+        """
         view = self.navigate_to(self, 'All')
         view.search(entity_name)
-        view.table.row(name=entity_name)['Actions'].widget.click(handle_alert=True)
-        view.flash.assert_no_error()
-        view.flash.dismiss()
+        view.table.row(name=entity_name)['Actions'].widget.click()
+        if cancel_dialog:
```
mshriver, line 48: https://github.com/SatelliteQE/robottelo/pull/9028/files#r722256318

related discussion going on in the robottelo PR

airgun/views/hardware_model.py
```diff
@@ -4,17 +4,24 @@
...
 class HardwareModelsView(BaseLoggedInView, SearchableViewMixin):
+    delete_dialog = DeleteHardwareModelDialog()
     title = Text("//h1[text()='Hardware Models']")
     new = Text("//a[contains(@href, '/models/new')]")
     table = SatTable(
         './/table',
         column_widgets={
             'Name': Text('.//a'),
-            'Actions': Text('.//a[@data-method="delete"]'),
+            'Actions': Text(".//button[text()='Delete']"),
```
omkarkhatavkar, line 24: Instead of the Text widget, you can use Button widget. (non blocking )

airgun/entities/hardware_model.py
```diff
@@ -37,13 +37,22 @@ def update(self, entity_name, values):
...
+            view.delete_dialog.confirm()
```
omkarkhatavkar, line 49: `view.delete_dialog.confirm()` this is can be written outside of the if : else block

mshriver (approved):

PR #637, opened 2021-11-25
Incorrect is_searchable return value | Incorrect condition

jyejare:
False Conditioning in `is_displayed()` returns a wrong boolean value making tests to search entity even though the search field is not available on the page.

### DIFF
--- a/airgun/views/common.py
+++ b/airgun/views/common.py
@@ -417,9 +417,9 @@ class SearchableViewMixin(WTMixin):
         That means that we have search field present on the page and that page
         is not a welcome one
         """
-        if not self.searchbox.search_field.is_displayed and self.welcome_message.is_displayed:
-            return False
-        return True
+        if self.searchbox.search_field.is_displayed and (not self.welcome_message.is_displayed):
+            return True
+        return False
 
     def search(self, query):
         """Perform search using searchbox on the page and return table


pondrejk:
Applied this patch and run robottelo locally, but getting this failure:
```
>           assert session.host.search(hostname)[0]['Name'] == hostname
E           TypeError: 'NoneType' object is not subscriptable
```

jyejare:
@pondrejk Does the failure genuine ?

omkarkhatavkar:
> ting this failure:

@pondrejk  which test is it ?

pondrejk:
Hello @jyejare @omkarkhatavkar, basically all ui/test_remoteexecution.py tests that do session.host.search are failing with NoSuchElement error complaining about about this line https://github.com/SatelliteQE/airgun/blob/ebdd0096855dde96b9298bc4ea0c26e4fcb27dfb/airgun/widgets.py#L632. When I apply this patch I no longer get this error but the search returns None, hence assertion error mentioned above

jyejare:
@peterdragun What tests is failing ? I will run this on my local setup, but I think the failure ur getting is expected but I will debug locally first !

pondrejk:
@jyejare, for example `pytest tests/foreman/ui/test_remoteexecution.py -k test_positive_run_custom_job_template_by_ip`, I don't see what could go wrong on the test side, it just searches for the host, I checked that the host exists

jyejare:
@pondrejk Let me test it on my local setup, and I shld feed u back !

pondrejk:
ok so the problem was in missing location context in the test, no error in this pr

pondrejk (approved):

jyejare:
@pondrejk Hey, 

I tested the changes locally and especially debug the test you mentioned in the comment above. I see the issue in the test:
1. While execution reaches to search the host on the page, it just fails to see the host cause the location context is not switched to the desired location.
2. Since host search returns none, the hostname fetching out of host search breaks with `TypeError: 'NoneType' object is not subscribable`.

This is the same issue as https://github.com/SatelliteQE/robottelo/pull/9195.

omkarkhatavkar (approved):

PR #638, opened 2021-11-26
removing the idle-timeout, webkaifuku takes care!

omkarkhatavkar:
No need for the code to set the idleTimeout as it set directly handled by webkaifuku 

For Test Result : git lab satelliteqe-jenkins - 448 PR

### DIFF
--- a/airgun/browser.py
+++ b/airgun/browser.py
@@ -174,9 +174,6 @@ class SeleniumBrowserFactory:
         manager = BrowserManager.from_conf(self.web_kaifuku)
         self._webdriver = manager.start()
         self._set_session_cookie()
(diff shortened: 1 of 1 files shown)


swadeley (approved):
ACK

jyejare:
@swadeley Even though Kaifuku takes care of default Timeout, we should be able to configure that from robottelo. In short we should make it configurable!

swadeley:
> @swadeley Even though Kaifuku takes care of default Timeout, we should be able to configure that from robottelo. In short we should make it configurable!

@omkar ^

omkarkhatavkar:
@jyejare Yes it is configurable ui.yaml file in robottelo repo

jyejare (approved):

PR #639, opened 2021-11-30
Add support for insights tab on host details page.

jameerpathan111:


.pre-commit-config.yaml
```diff
@@ -11,13 +11,13 @@ repos:
...
   hooks:
   - id: black
-- repo: https://gitlab.com/pycqa/flake8
-  rev: 3.9.2
-  hooks:
-  - id: flake8
+#- repo: https://gitlab.com/pycqa/flake8
+#  rev: 3.9.2
+#  hooks:
+#  - id: flake8
```
omkarkhatavkar, line 20: Did you just pasted the local changes, please revert this.

airgun/entities/host.py
```diff
@@ -129,6 +130,11 @@ def read_insights_recommendations(self, entity_name):
         view = self.navigate_to(self, 'Recommendations')
         return view.table.read()
 
+    def insights_tab(self, entity_name):
```
omkarkhatavkar, line 133: no need for this I can `session.host.get_details('entity-host-name', widget_names='insights')`

airgun/entities/host.py
```diff
@@ -378,3 +384,18 @@ class ShowRecommendations(NavigateStep):
     """Navigate to Insights recommendations page"""
 
     VIEW = CloudInsightsView
+
+
+@navigator.register(HostEntity, 'InsightsTab')
```
omkarkhatavkar, line 389: This is also not required as this can be achieved using existing code

omkarkhatavkar (changes requested):
pending test result

airgun/entities/host.py
jameerpathan111: There's one more way using which user can reach the insights tab on host details page.
User can click on recommendation count on all host page which navigates directly to insights tab.
I have added this code to test that navigation scenario.

airgun/entities/host.py
jameerpathan111: User can click on recommendation count on all host page which navigates directly to insights tab.
I have added this code to test that navigation scenario.

jameerpathan111: I have updated the pr so that I can also use `session.host.get_details('entity-host-name', widget_names='insights')`.

jameerpathan111:
> pending test result

Test result:
```
(Pdb) output = session.host.get_details(hostname, widget_names='insights')
(Pdb) output
{'insights': {'insights_tab': [{'name': 'Decreased security: Yum GPG verification disabled (third-party repos)', 'label': 'Critical', 'text': 'Decreased security: Yum GPG verification disabled (third-party repos) Knowledgebase article Read more about it in RH cloud insights'}, {'name': 'When the Insights Client is earlier than "3.1.0", it is not able to get recommendations dependent on core collection', 'label': 'Important', 'text': 'When the Insights Client is earlier than "3.1.0", it is not able to get recommendations dependent on core collection Knowledgebase article Read more about it in RH cloud insights'}, {'name': 'The dnf installs lower versions of packages when the "best" option is not present in the /etc/dnf/dnf.conf', 'label': 'Moderate', 'text': 'The dnf installs lower versions of packages when the "best" option is not present in the /etc/dnf/dnf.conf Read more about it in RH cloud insights'}, {'name': 'Kdump will fail if the minimum required crashkernel reservation size is not configured correctly when the third-party modules are loaded', 'label': 'Moderate', 'text': 'Kdump will fail if the minimum required crashkernel reservation size is not configured correctly when the third-party modules are loaded Knowledgebase article Read more about it in RH cloud insights'}]}}


(Pdb) result = session.host.insights_tab(hostname)
(Pdb) result
[{'name': 'Decreased security: Yum GPG verification disabled (third-party repos)', 'label': 'Critical', 'text': 'Decreased security: Yum GPG verification disabled (third-party repos) Knowledgebase article Read more about it in RH cloud insights'}, {'name': 'When the Insights Client is earlier than "3.1.0", it is not able to get recommendations dependent on core collection', 'label': 'Important', 'text': 'When the Insights Client is earlier than "3.1.0", it is not able to get recommendations dependent on core collection Knowledgebase article Read more about it in RH cloud insights'}, {'name': 'The dnf installs lower versions of packages when the "best" option is not present in the /etc/dnf/dnf.conf', 'label': 'Moderate', 'text': 'The dnf installs lower versions of packages when the "best" option is not present in the /etc/dnf/dnf.conf Read more about it in RH cloud insights'}, {'name': 'Kdump will fail if the minimum required crashkernel reservation size is not configured correctly when the third-party modules are loaded', 'label': 'Moderate', 'text': 'Kdump will fail if the minimum required crashkernel reservation size is not configured correctly when the third-party modules are loaded Knowledgebase article Read more about it in RH cloud insights'}]

```

omkarkhatavkar (approved):

airgun/views/host.py
```diff
@@ -553,6 +554,70 @@ def before_fill(self, values):
...
+class RecommendationWidget(GenericLocatorWidget):
+    """The widget representation of recommendation item."""
+
+    EXPAND_BUTTON = ".//div[contains(@class, 'expand')]"
+    NAME = ".//div/p[contains(@class, 'item-heading')]"
+    RISK_LABEL = ".//div/span[contains(@class, 'risk-label')]"
+    TEXT = ".//div[contains(@class, 'list-group-item-container')]"
+
+    @property
+    def expand_button(self):
```
jyejare, line 566: I think this should not be needed from outside world since you have created a property out of it.

You can either directly use it or create a var and use the var in expanded def.

Non-blocking comment though!

jyejare (approved):

PR #607, opened 2021-08-19
Add REX support to chost and errata tests

swadeley:
Hello airgunners

As REX is default in 6.10 I had to add support for the Job Invocation page which is used when applying errata, to track task progress, rather than CHost Details Task view.

Used by tests in `tests/foreman/ui/test_contenthost.py` and `tests/foreman/ui/test_hostcollection.py`.

Thank you

[Update test ui test_contenthost to use REX #8817](https://github.com/SatelliteQE/robottelo/pull/8817)
[Convert test hostcollection to use REX #8847](https://github.com/SatelliteQE/robottelo/pull/8847)

PR #641, opened 2021-12-06
Fix applying errata progress page

latran:
The entities previouslyhad the jobinvocationpage for display of the applying errata progress which is incorrect.  I'm rewriting it to be in the `ErrataTaskDetailsView`.

results can be found with https://github.com/SatelliteQE/robottelo/pull/9156

### DIFF
--- a/airgun/entities/errata.py
+++ b/airgun/entities/errata.py
@@ -7,7 +7,6 @@ from airgun.views.errata import ErrataDetailsView
 from airgun.views.errata import ErrataInstallationConfirmationView
 from airgun.views.errata import ErrataTaskDetailsView
 from airgun.views.errata import ErratumView
-from airgun.views.job_invocation import JobInvocationStatusView
 
 
 class ErrataEntity(BaseEntity):
@@ -59,16 +58,11 @@ class ErrataEntity(BaseEntity):
             view.content_hosts.environment_filter.fill(environment)
         return view.read(widget_names=widget_names)
 
-    def install(self, entity_name, host_name, installed_via=None):
+    def install(self, entity_name, host_name):
         """Install errata on content host.
-
-        The installation method is not set here, but the path changes according to the method used.
-        For katello-agent, the Content Hosts' Task tab displays the progress. If REX is used,
-        the Job Invocation view displays the progress. In 6.10, REX became the default method.
 
         :param str entity_name: errata id or title
         :param str host_name: content host name to apply errata on
-        :param installed_via: what installation method was used
         """
         view = self.navigate_to(
             self,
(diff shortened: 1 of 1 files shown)


swadeley:
Hello @latran 

that change was introduced here:
 Add REX support to chost and errata tests #607 

Can you confirm you are using Sat6.10 for testing and have not enabled katello-agent on your test Satellite?

Thank you

latran:
@swadeley I am using sat6.10 and I don't think sat enables katello-agent by default.  And `ErrataDetailsViews` is different than the `CHostDetailsTaskView`.

latran:
@akhil-jha Can you explain to me the difference between the `entities/errata.py` versus the `entities/contesthost.py` with regards to the `install` versus `install_errata` function?  I'm not clear on when to use one versus the other and because I'm using `install` specifically for incremental update for one test case (it seems no other test cases uses it) and if that is its main functionally, I'm thinking of renaming it to `incremental_update` in `erratay.py`.

akhil-jha:
Hey sorry for the late response. As for the difference, yes, they almost do the same thing. However, I don't think renaming it to incremental_update is required. If I am not wrong, the navigation of errata.py(sat.com/errata) and contenthost.py(sat.com/content_hosts) are different. So I think keeping both of them separate seems okay.

swadeley:
> @swadeley I am using sat6.10 and I don't think sat enables katello-agent by default. And `ErrataDetailsViews` is different than the `CHostDetailsTaskView`. I will look more into how two are executed. Looking at the code that I changed, it's completely redundant.

Hi @latran , I tested this, so it was my mistake, that change  made was not required for the `ErrataDetailsViews` .

I suggest change the description:

Install errata on a content host using Errata's details view.

More importantly, there are other changes that need to be reverted:


```
-    def install(self, entity_name, host_name):
+    def install(self, entity_name, host_name, installed_via=None):
         """Install errata on content host.
 
+        The installation method is not set here, but the path changes according to the method used.
+        For katello-agent, the Content Hosts' Task tab displays the progress. If REX is used,
+        the Job Invocation view displays the progress. In 6.10, REX became the default method.
+
         :param str entity_name: errata id or title
         :param str host_name: content host name to apply errata on
+        :param installed_via: what installation method was used
         """
```

I can make a new PR for you with the reverted changes if you like. other wise replace lines marked + with lines marked -

Thank you

swadeley:
Hi

Don't forget, second line of git commit message must be blank.

latran:
@swadeley not sure what you mean about the git commit, but your requested changes are in.

swadeley (approved):
ACK

omkarkhatavkar (approved):
ACK

Issue #591, opened 2021-07-20
Xpath locators should use `normalize-space(.)` for text matching


mshriver:
Currently there is a mix of `.` and `text()` use in locators, which are sensitive to leading/trailing whitespace and newlines in the element text.

I think we should standardize on using `normalize-space(.)` within xpath locators when doing element text comparisons, as this removes the need to include any leading/trailing whitespace in the expected text match.

It's a simple change to the locators that stabilizes them against small UI changes to whitespace.

PR #642, opened 2021-12-08
Add ansible roles view and entity

synkd:
This PR adds views and entities for the Ansible Roles page (located in the webUI at Configure > Roles) and the Ansible Variables page (located in the webUI at Configured > Variables).

This is my first Airgun PR, so I would appreciate any feedback.

Note that there are two widgets in the `NewAnsibleVariableView` view that are not currently functioning properly; I have marked these with comments. However, in the interest of getting these changes merged ahead of 6.10 branching next week, I thought it best to get this up for review in its current state.

### DIFF
--- a/airgun/entities/ansible_role.py
+++ b/airgun/entities/ansible_role.py
@@ -20,8 +20,12 @@ class AnsibleRolesEntity(BaseEntity):
 
     def delete(self, entity_name):
         """Delete Ansible Role from Satellite"""
-        # This method currently does not work as the Name column table header
-        # cell includes the `▲` character used for sorting that column
+        # This method currently relies on searching for a specific role as
+        # directly accessing the 'Name' column is not possible due to the
+        # presence of the `▲` character used for sorting that column in the table
+        # header cell. The Satellite UX team is planning to address this in a
+        # future release, likely by wrapping the sort character in a separate
+        # span tag from the column header text.
         view = self.navigate_to(self, 'All')
         view.search(entity_name)
         view.table.row()['Actions'].widget.fill('Delete')
--- a/airgun/entities/ansible_role.py
+++ b/airgun/entities/ansible_role.py
@@ -1,0 +1,75 @@
+from navmazing import NavigateToSibling
+
+from airgun.entities.base import BaseEntity
+from airgun.navigation import NavigateStep
+from airgun.navigation import navigator
+from airgun.views.ansible_role import AnsibleRolesImportView
+from airgun.views.ansible_role import AnsibleRolesView
+
+
+class AnsibleRolesEntity(BaseEntity):
+    """Main Ansible roles entity"""
+
+    endpoint_path = '/ansible/ansible_roles'
+
+    def search(self, value):
+        """Search for existing Ansible Role"""
+        view = self.navigate_to(self, 'All')
+        view.search(value)
+        return view.table.read()
+
+    def delete(self, entity_name):
+        """Delete Ansible Role from Satellite"""
+        # This method currently does not work as the Name column table header
+        # cell includes the `▲` character used for sorting that column
+        view = self.navigate_to(self, 'All')
+        view.search(entity_name)
+        view.table.row()['Actions'].widget.fill('Delete')
+        view.dialog.confirm_dialog.click()
+        view.flash.assert_no_error()
+        view.flash.dismiss()
+
+    @property
+    def imported_roles_count(self):
+        """Return the number of Ansible roles currently imported into Satellite"""
+        view = self.navigate_to(self, 'All')
+        return int(view.total_imported_roles.read())
+
+    def import_all_roles(self):
+        """Import all available roles and return the number of roles
+        that were available at import time
+        """
+        view = self.navigate_to(self, 'Import')
+        available_roles_count = int(view.total_available_roles.read())
+        view.select_all.fill(True)
+        view.submit.click()
+        return available_roles_count
+
+    # Since Ansible Variables need to be assigned to specific Ansible Roles,
+    # Ansible Variable tests will fail if no Ansible Roles have been imported.
+    def preimport_check(self):
+        view = self.navigate_to(self, 'All')
+        if not view.pagination.is_displayed:
+            return False
+
+
+@navigator.register(AnsibleRolesEntity, 'All')
+class ShowAllRoles(NavigateStep):
+    """Navigate to the Ansible Roles page"""
+
+    VIEW = AnsibleRolesView
+
+    def step(self, *args, **kwargs):
+        self.view.menu.select('Configure', 'Roles')
+
+
+@navigator.register(AnsibleRolesEntity, 'Import')
+class ImportAnsibleRole(NavigateStep):
+    """Navigate to the Import Roles page"""
+
+    VIEW = AnsibleRolesImportView
+
+    prerequisite = NavigateToSibling('All')
+
+    def step(self, *args, **kwargs):
+        self.parent.import_button.click()
--- a/airgun/entities/ansible_variable.py
+++ b/airgun/entities/ansible_variable.py
@@ -1,0 +1,74 @@
+from navmazing import NavigateToSibling
+
+from airgun.entities.base import BaseEntity
+from airgun.navigation import NavigateStep
+from airgun.navigation import navigator
+from airgun.views.ansible_variable import AnsibleVariablesView
+from airgun.views.ansible_variable import NewAnsibleVariableView
+
+
+class AnsibleVariablesEntity(BaseEntity):
+    """Main Ansible variables entity"""
+
+    endpoint_path = '/ansible/ansible_variables'
+
+    def search(self, value):
+        """Search for existing Ansible variable"""
+        view = self.navigate_to(self, 'All')
+        view.search(value)
+        return view.table.read()
+
+    def delete(self, entity_name):
+        """Delete Ansible variable from Satellite"""
+        view = self.navigate_to(self, 'All')
+        view.search(entity_name)
+        view.table.row(name=entity_name)['Actions'].widget.click(handle_alert=True)
+        view.confirm.confirm_dialog.click()
+        view.flash.assert_no_error()
+        view.flash.dismiss()
+
+    def read_total_variables(self):
+        """Returns the number of Ansible variables currently in Satellite"""
+        view = self.navigate_to(self, 'All')
+        return view.total_variables.read()
+
+    def create(self, values):
+        """Create a new Ansible variable with minimum inputs"""
+        view = self.navigate_to(self, 'New')
+        view.fill(values)
+        view.submit.click()
+        view.flash.assert_no_error()
+        view.flash.dismiss()
+
+    def create_with_overrides(self, values):
+        """Create a new Ansible variable that is managed by Satellite"""
+        view = self.navigate_to(self, 'New')
+        view.override.fill(True)
+        view.expand()
+        view.matcher_section.before_fill(values)
+        view.fill(values)
+        view.submit.click()
+        view.flash.assert_no_error()
+        view.flash.dismiss()
+
+
+@navigator.register(AnsibleVariablesEntity, 'All')
+class ShowAllVariables(NavigateStep):
+    """Navigate to Ansible Variables page"""
+
+    VIEW = AnsibleVariablesView
+
+    def step(self, *args, **kwargs):
+        self.view.menu.select('Configure', 'Variables')
+
+
+@navigator.register(AnsibleVariablesEntity, 'New')
+class NewAnsibleVariable(NavigateStep):
+    """Navigate to Create Ansible Variable page"""
+
+    VIEW = NewAnsibleVariableView
+
+    prerequisite = NavigateToSibling('All')
+
+    def step(self, *args, **kwargs):
+        self.parent.new_variable.click()
--- a/airgun/entities/ansible_variable.py
+++ b/airgun/entities/ansible_variable.py
@@ -22,8 +22,8 @@ class AnsibleVariablesEntity(BaseEntity):
         """Delete Ansible variable from Satellite"""
         view = self.navigate_to(self, 'All')
         view.search(entity_name)
-        view.table.row(name=entity_name)['Actions'].widget.click(handle_alert=True)
-        view.confirm.confirm_dialog.click()
+        view.table.row(name=entity_name)['Actions'].widget.click()
+        view.dialog.confirm_dialog.click()
         view.flash.assert_no_error()
         view.flash.dismiss()
 
--- a/airgun/views/ansible_role.py
+++ b/airgun/views/ansible_role.py
@@ -1,0 +1,70 @@
+from widgetastic.widget import Checkbox
+from widgetastic.widget import Table
+from widgetastic.widget import Text
+from widgetastic_patternfly import BreadCrumb
+from widgetastic_patternfly4 import Button
+from widgetastic_patternfly4 import PatternflyTable
+
+from airgun.views.common import BaseLoggedInView
+from airgun.views.common import SearchableViewMixin
+from airgun.widgets import ActionsDropdown
+from airgun.widgets import Pagination
+
+
+class ImportPagination(Pagination):
+    PER_PAGE_BUTTON_DROPDOWN = ".//div[button[@id='pagination-options-menu-toggle-2']]"
+
+
+class AnsibleRolesView(BaseLoggedInView, SearchableViewMixin):
+    """Main Ansible Roles view. Prior to importing any roles, only the import_button
+    is present, without the search widget or table.
+    """
+
+    title = Text("//h1[contains(., text()='Ansible Roles')")
+    import_button = Text("//a[contains(@href, '/ansible_roles/import')]")
+    submit = Button('Submit')
+    total_imported_roles = Text("//span[contains(@class, 'pagination-pf-items-total')]")
+    table = Table(
+        './/table',
+        column_widgets={
+            'Name': Text("./a"),
+            'Hostgroups': Text("./a"),
+            'Hosts': Text("./a"),
+            'Imported at': Text("./a"),
+            'Actions': ActionsDropdown("./div[contains(@class, 'btn-group')]"),
+        },
+    )
+    pagination = Pagination()
+
+    @property
+    def is_displayed(self):
+        return self.title.is_displayed and self.import_button.is_displayed
+
+
+class AnsibleRolesImportView(BaseLoggedInView):
+    """View while selecting Ansible roles to import."""
+
+    breadcrumb = BreadCrumb()
+    total_available_roles = Text("//div[@class='pf-c-pagination__total-items']/b[2]")
+    select_all = Checkbox(locator="//input[@id='select-all']")
+    table = PatternflyTable(
+        component_id='OUIA-Generated-Table-2',
+        column_widgets={
+            0: Checkbox(locator='.//input[@type="checkbox"]'),
+            'Name': Text('.//a'),
+            'Operation': Text('.//a'),
+            'Variables': Text('.//a'),
+            'Hosts Count': Text('.//a'),
+            'Hostgroups Count': Text('.//a'),
+        },
+    )
+    pagination = ImportPagination()
+    submit = Button('Submit')
+    cancel = Button('Cancel')
+
+    @property
+    def is_displayed(self):
+        return (
+            self.breadcrumb.locations[0] == 'Roles'
+            and self.breadcrumb.read() == 'Changed Ansible roles'
+        )
--- a/airgun/views/ansible_role.py
+++ b/airgun/views/ansible_role.py
@@ -7,12 +7,13 @@ from widgetastic_patternfly4 import PatternflyTable
 
 from airgun.views.common import BaseLoggedInView
 from airgun.views.common import SearchableViewMixin
-from airgun.widgets import ActionsDropdown
+from airgun.widgets import ActionsDropdown, Pf4ConfirmationDialog
 from airgun.widgets import Pagination
 
 
 class ImportPagination(Pagination):
     PER_PAGE_BUTTON_DROPDOWN = ".//div[button[@id='pagination-options-menu-toggle-2']]"
+    total_items = Text("//span[@class='pf-c-optionsmenu__toggle-text']/b[2]")
 
 
 class AnsibleRolesView(BaseLoggedInView, SearchableViewMixin):
@@ -23,14 +24,10 @@ class AnsibleRolesView(BaseLoggedInView, SearchableViewMixin):
     title = Text("//h1[contains(., text()='Ansible Roles')")
     import_button = Text("//a[contains(@href, '/ansible_roles/import')]")
     submit = Button('Submit')
-    total_imported_roles = Text("//span[contains(@class, 'pagination-pf-items-total')]")
+    total_imported_roles = Text(".//span[contains(@class, 'pagination-pf-items-total')]")
     table = Table(
         './/table',
         column_widgets={
-            'Name': Text("./a"),
-            'Hostgroups': Text("./a"),
-            'Hosts': Text("./a"),
-            'Imported at': Text("./a"),
             'Actions': ActionsDropdown("./div[contains(@class, 'btn-group')]"),
         },
     )
@@ -51,11 +48,6 @@ class AnsibleRolesImportView(BaseLoggedInView):
         component_id='OUIA-Generated-Table-2',
         column_widgets={
             0: Checkbox(locator='.//input[@type="checkbox"]'),
-            'Name': Text('.//a'),
-            'Operation': Text('.//a'),
-            'Variables': Text('.//a'),
-            'Hosts Count': Text('.//a'),
-            'Hostgroups Count': Text('.//a'),
         },
     )
     pagination = ImportPagination()
--- a/airgun/views/ansible_role.py
+++ b/airgun/views/ansible_role.py
@@ -7,7 +7,7 @@ from widgetastic_patternfly4 import PatternflyTable
 
 from airgun.views.common import BaseLoggedInView
 from airgun.views.common import SearchableViewMixin
-from airgun.widgets import ActionsDropdown, Pf4ConfirmationDialog
+from airgun.widgets import ActionsDropdown
 from airgun.widgets import Pagination
 
 
--- a/airgun/views/ansible_variable.py
+++ b/airgun/views/ansible_variable.py
@@ -11,7 +11,6 @@ from airgun.views.common import SearchableViewMixin
 from airgun.widgets import CustomParameter
 from airgun.widgets import FilteredDropdown
 from airgun.widgets import Pagination
-from airgun.widgets import Pf4ConfirmationDialog
 from airgun.widgets import SatSelect
 
 
--- a/airgun/views/ansible_variable.py
+++ b/airgun/views/ansible_variable.py
@@ -84,6 +84,9 @@ class NewAnsibleVariableView(BaseLoggedInView):
         add_matcher = Text("//a[contains(@class, 'add_nested_fields')]")
         params = MatcherTable(
             locator=".//table[@class='table white-header']",
+            # new_row_bottom is passed to the __init__ method of the
+            # CustomParameter widget, see comment there for additional details
+            new_row_bottom=True,
             column_widgets={
                 'Attribute type': MatcherActions(),
                 'Value': TextInput(locator=".//textarea[@id='new_lookup_value_value']"),
--- a/airgun/views/ansible_variable.py
+++ b/airgun/views/ansible_variable.py
@@ -55,19 +55,26 @@ class NewAnsibleVariableView(BaseLoggedInView):
     """View while creating a new Ansible Variable"""
 
     breadcrumb = BreadCrumb()
+
+    # 'Ansible Variable Details' section
     key = TextInput(id='ansible_variable_key')
     description = TextInput(id='ansible_variable_description')
     ansible_role = FilteredDropdown(id='ansible_variable_ansible_role_id')
-    override = Checkbox(id='ansible_variable_override')
 
+    # 'Default Behavior' section
+    override = Checkbox(id='ansible_variable_override')
     # Accessing all widgets except the ones above requires that the `override` checkbox is filled
     parameter_type = SatSelect(id='ansible_variable_parameter_type')
     default_value = TextInput(id='ansible_variable_default_value')
     hidden_value = Checkbox(id='ansible_variable_hidden_value')
+
+    # 'Optional Input Validator' section
     expand_optional_input_validator = Text("//h2[@class='expander collapsed']")
     required = Checkbox(id='ansible_variable_required')
     validator_type = SatSelect(id='ansible_variable_validator_type')
     validator_rule = TextInput(id='ansible_variable_validator_rule')
+
+    # 'Prioritize Attribute Order' section
     attribute_order = TextInput(id='order')
     merge_overrides = Checkbox(id='ansible_variable_merge_overrides')
     merge_default = Checkbox(id='ansible_variable_merge_default')
@@ -77,6 +84,8 @@ class NewAnsibleVariableView(BaseLoggedInView):
 
     @View.nested
     class matcher_section(View):
+        """'Specify Matchers' section"""
+
         add_matcher = Text("//a[contains(@class, 'add_nested_fields')]")
         params = MatcherTable(
             locator=".//table[@class='table white-header']",
--- a/airgun/views/ansible_variable.py
+++ b/airgun/views/ansible_variable.py
@@ -20,19 +20,14 @@ class AnsibleVariablesView(BaseLoggedInView, SearchableViewMixin):
 
     title = Text("//h1[contains(., text()='Ansible Variables')")
     new_variable = Text("//a[contains(@href, '/ansible/ansible_variables/new')]")
-    total_variables = Text("//span[@class='pagination-pf-items-total']")
+    total_variables = Text(".//span[@class='pagination-pf-items-total']")
     table = SatTable(
         './/table',
         column_widgets={
-            'Name': Text("./a"),
-            'Role': Text("./a"),
-            'Type': Text("./a"),
-            'Imported?': Text("./a"),
             'Actions': Text(".//a[@data-method='delete']"),
         },
     )
     pagination = Pagination()
-    confirm = Pf4ConfirmationDialog()
 
     @property
     def is_displayed(self):
--- a/airgun/views/ansible_variable.py
+++ b/airgun/views/ansible_variable.py
@@ -1,0 +1,116 @@
+from widgetastic.widget import Checkbox
+from widgetastic.widget import Select
+from widgetastic.widget import Text
+from widgetastic.widget import TextInput
+from widgetastic.widget import View
+from widgetastic_patternfly import BreadCrumb
+
+from airgun.views.common import BaseLoggedInView
+from airgun.views.common import SatTable
+from airgun.views.common import SearchableViewMixin
+from airgun.widgets import CustomParameter
+from airgun.widgets import FilteredDropdown
+from airgun.widgets import Pagination
+from airgun.widgets import Pf4ConfirmationDialog
+from airgun.widgets import SatSelect
+
+
+class AnsibleVariablesView(BaseLoggedInView, SearchableViewMixin):
+    """Main Ansible Variables view"""
+
+    title = Text("//h1[contains(., text()='Ansible Variables')")
+    new_variable = Text("//a[contains(@href, '/ansible/ansible_variables/new')]")
+    total_variables = Text("//span[@class='pagination-pf-items-total']")
+    table = SatTable(
+        './/table',
+        column_widgets={
+            'Name': Text("./a"),
+            'Role': Text("./a"),
+            'Type': Text("./a"),
+            'Imported?': Text("./a"),
+            'Actions': Text(".//a[@data-method='delete']"),
+        },
+    )
+    pagination = Pagination()
+    confirm = Pf4ConfirmationDialog()
+
+    @property
+    def is_displayed(self):
+        return self.title.is_displayed and self.new_variable.is_displayed
+
+
+class MatcherTable(CustomParameter):
+    add_new_value = Text("..//a[contains(text(),'+ Add Matcher')]")
+
+
+class MatcherActions(View):
+    """Interface table has Attribute type column that contains select field and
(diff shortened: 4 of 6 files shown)


airgun/entities/ansible_role.py
```diff
@@ -0,0 +1,57 @@
...
+    def delete(self, entity_name, values):
+        """Delete Ansible Role from Satellite"""
+        view = self.navigate_to(self, 'All')
+        view.search(entity_name)
+        view.table.row(name=entity_name)['Actions'].widget.fill('Delete')
+        self.browser.handle_alert()
+        view.flash.assert_no_error()
+        view.flash.dismiss()
+
+    def count_imported_roles(self):
```
mshriver, line 25: Not sure about naming for this in airgun, but I'd think about two things with this function:

1. It's a property (should be decorated, takes no arguments and doesn't change state)
2. I'd call it `imported_roles_count`, but this is kinda subjective and might go against the grain with other entity function names in airgun.

airgun/entities/ansible_role.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+@navigator.register(AnsibleRolesEntity, 'Import')
+class ImportAnsibleRole(NavigateStep):
+    """Navigate to the Import Roles page"""
+
+    endpoint_path = '/ansible/ansible_roles'
+    VIEW = AnsibleRolesImportView
+
+    def prerequisite(self, *args, **kwargs):
```
mshriver, line 53: You can use `NavigateToSibling` here to flatten this quite a bit. I don't think its in wide use in airgun, you followed the 'regular' pattern here.

```
from navmazing import NavigateToSibling # at the top of the file

# then within the class, where you currently are defining prerequisite as a function

prerequisite = NavigateToSibling('All')
```

The class takes care of resolving the object, and navigating at the right time. 

https://github.com/RedHatQE/navmazing/blob/master/navmazing/__init__.py#L165

airgun/entities/ansible_variable.py
```diff
@@ -0,0 +1,67 @@
...
+        self.view.menu.select('Configure', 'Variables')
+
+
+@navigator.register(AnsibleVariablesEntity, 'New')
+class NewAnsibleVariable(NavigateStep):
+    """Navigate to Create Ansible Variable page"""
+
+    VIEW = NewAnsibleVariableView
+
+    def prerequisite(self, *args, **kwargs):
```
mshriver, line 63: Same thing applies to this prerequisite too - `NavigateToSibling` and `NavigateToAttribute` are handy classes to define navigation functions that relate to other destinations on the entity they're registered against.

airgun/views/ansible_role.py
```diff
@@ -0,0 +1,75 @@
...
+
+class AnsibleRolesView(BaseLoggedInView, SearchableViewMixin):
+    """Main Ansible Roles view. Prior to importing any roles, only the import_button
+    is present, without the search widget or table.
+    """
+
+    title = Text("//h1[contains(., text()='Ansible Roles')")
+    import_button = Text("//a[contains(@href, '/ansible_roles/import')]")
+    submit = Button('Submit')
+    total_imported_roles = Text("//span[@class='pagination-pf-items-total']")
```
mshriver, line 26: From the xpath, this looks like the pagination widget's `total_items` attribute:
https://github.com/SatelliteQE/airgun/blob/master/airgun/widgets.py#L1453

You can reference it straight through `view.pagination.total_items` which I generally recommend, but if you want to create a more direct reference to the number of imported roles you could define a small property function.

```
@property
def total_imported_roles(self):
    return self.pagination.total_items
```

airgun/views/ansible_role.py
```diff
@@ -0,0 +1,75 @@
...
+            'Hosts': Text("./a"),
+            'Imported at': Text("./a"),
+            'Actions': ActionsDropdown("./div[contains(@class, 'btn-group')]"),
+        },
+    )
+    pagination = Pagination()
+
+    @property
+    def is_displayed(self):
+        return self.browser.wait_for_element(
```
mshriver, line 41: Consider leveraging more of the defined widgets to assert the view:

```
return (
    self.title.is_displayed and
    self.import_button.is_displayed and
    self.table.is_displayed
)
```

airgun/views/ansible_variable.py
```diff
@@ -0,0 +1,91 @@
...
+    validator_rule = TextInput(id='ansible_variable_validator_rule')
+    # the below attribute_order widget throws an error when a string value is passed to it
+    attribute_order = TextInput(id='order')
+    merge_overrides = Checkbox(id='ansible_variable_merge_overrides')
+    merge_default = Checkbox(id='ansible_variable_merge_default')
+    avoid_duplicates = Checkbox(id='ansible_variable_avoid_duplicates')
+    add_matcher_button = Text("//a[contains(@class, 'add_nested_fields')]")
+    attribute_type = SatSelect("//div[@class='matcher-group']/select[1]")
+    # the below attribute_value widget is not locating the element correctly as currently written
+    attribute_value = Text("//div[@class='matcher-group']/input[1]")
```
latran, line 64: These will be a little bit difficult to handle because there could be potential multiple `Attribute type` when you click `Add Matcher`.  One shouldn't give much issue when it's trying to search for the one available field though.  Try `@class='matcher_value'` and see if that works.  if it doesn't, you may need to set the `locator` to that or pass in the `input` one.

latran (changes requested):
Minor changes

airgun/views/ansible_role.py
synkd: Been trying to figure out how to implement this. I see a couple of tests in Robottelo that use do something similar, such as test_contenthost.py:

https://github.com/SatelliteQE/robottelo/blob/9edf53cb6f30c77e13ab4557b1668033406904c1/tests/foreman/ui/test_contenthost.py#L1321

So it looks like I'll need to add something like the `read_all` method from the content host entity in order to access that data:

https://github.com/SatelliteQE/airgun/blob/e6a29c3c0a46dc8a63c9818477fc733ae3a613fb/airgun/entities/contenthost.py#L31

Is something like this what you had in mind, or are there other possibilities?

airgun/views/ansible_role.py
synkd: I'm making this change but leaving off `self.table.is_displayed` as the table is not present when no roles have been imported.

airgun/entities/ansible_role.py
```diff
@@ -0,0 +1,58 @@
...
+        view = self.navigate_to(self, 'All')
+        view.search(entity_name)
+        view.table.row(name=entity_name)['Actions'].widget.fill('Delete')
+        self.browser.handle_alert()
+        view.flash.assert_no_error()
+        view.flash.dismiss()
+
+    @property
+    def imported_roles_count(self):
+        view = self.navigate_to(self, 'All')
```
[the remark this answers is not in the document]
synkd: Added.

airgun/entities/ansible_role.py
```diff
@@ -0,0 +1,58 @@
...
+        view = self.navigate_to(self, 'Import')
+        available_roles_count = view.total_available_roles.read()
+        view.select_all.fill(True)
+        view.submit.click()
+        return available_roles_count
+
+
+@navigator.register(AnsibleRolesEntity, 'All')
+class ShowAllRoles(NavigateStep):
+
```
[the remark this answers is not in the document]
synkd: Added.

airgun/entities/ansible_variable.py
```diff
@@ -0,0 +1,67 @@
...
+    def delete(self, entity_name, values):
+        """Delete Ansible Role from Satellite"""
+        view = self.navigate_to(self, 'All')
+        view.search(entity_name)
+        view.table.row(name=entity_name)['Actions'].widget.fill('Delete')
+        self.browser.handle_alert()
+        view.flash.assert_no_error()
+        view.flash.dismiss()
+
+    def read_total_variables(self):
```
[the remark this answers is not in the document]
synkd: Added.

airgun/entities/ansible_variable.py
```diff
@@ -0,0 +1,67 @@
...
+        view = self.navigate_to(self, 'All')
+        return view.total_variables.read()
+
+    def create(self, values):
+        view = self.navigate_to(self, 'New')
```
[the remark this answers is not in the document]
synkd: Added.

airgun/entities/ansible_variable.py
```diff
@@ -0,0 +1,67 @@
...
+        view.fill(values)
+        view.submit.click()
+        view.flash.assert_no_error()
+        view.flash.dismiss()
+
+    def create_with_overrides(self, values):
```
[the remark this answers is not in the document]
synkd: Added.

airgun/views/ansible_variable.py
```diff
@@ -0,0 +1,91 @@
...
+
+    # Accessing all widgets except the ones above requires that the `override` checkbox is filled
+    parameter_type = SatSelect(id='ansible_variable_parameter_type')
+    default_value = TextInput(id='ansible_variable_default_value')
+    hidden_value = Checkbox(id='ansible_variable_hidden_value')
+    expand_input_validator = Text("//h2[@class='expander collapsed']")
+    required = Checkbox(id='ansible_variable_required')
+    validator_type = SatSelect(id='ansible_variable_validator_type')
+    validator_rule = TextInput(id='ansible_variable_validator_rule')
+    # the below attribute_order widget throws an error when a string value is passed to it
```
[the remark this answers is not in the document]
synkd: `E               ValueError: Cannot locate option with visible text: 'fqdn'. Available options: '', ''`

This element is a bit strange; I don't know that there is a similar one elsewhere in Satellite. It's a `textarea` with a `rows` attribute. In the source HTML, the values are listed as a space-separated string (by default, `fqdn hostgroup os domain`). When the HTML is rendered, the order is preserved, but each value is on a separate row. 

I've tried to change the order by passing in both a list (`'attribute_order': ['domain', 'fqdn', 'hostgroup', 'os']` and a string (\'attribute_order': 'domain fqdn hostgroup os'`), but both of these result in the error above.

latran:
@synkd Any updates on this and were you able to resolve your issue?

synkd:
@latran Thanks for following up. I will revisit this PR tomorrow and take stock.

airgun/views/ansible_role.py
mshriver: To access this value through the entity itself, yes you would need an entity function like `read_all` to read the view and return the dict.

The Pagination widget itself can just be directly read, or its `total_items` property can be accessed.

airgun/views/ansible_variable.py
synkd: I've been working on this today. Still unable to get the `attribute_value` field working. I've tried the following XPath expressions:
- `TextInput("//div[@class='matcher-group']/input[1]")`
- `TextInput("//input[contains(@class, 'matcher_value')")`
- `TextInput("//input[@class='matcher_value']")`
- several other minor variations on the above.

All of these throw a `NoSuchElementException` when my test reaches the `attribute_value` variable. Weirdly, the current XPath expression in the PR (`Text("//div[@class='matcher-group']/input[1]"`) does throw the same exception. However, `Text` objects do not have a `fill` method, and changing `Text` to `TextInput` brings back the exception. I do not have an explanation for this discrepancy in finding the element.

At this point, I am inclined to relegate tests involving matchers to manual tests for the time being, as this issue is preventing us from automating Ansible roles and variables tests that do not involve matchers. @latran @mshriver I would be interested to hear your perspectives on this idea.

synkd: @mshriver @latran I was able to spend some more time on this today. One of the things that I tried was a much more verbose XPath expression to point to this input field:
`TextInput("//div[@class='matcher-parent']/fieldset[2]/div[1]/table[1]/tbody[1]/tr[1]/td[1]/div[1]/input[1]")`
I verified in a browser console that this expression does indeed point to the intended text input field. However, my test still throws a `NoSuchElementException` when it attempts to use this input field. For reference, the test I'm using is here [1]. 

I could really use some help figuring out why I'm getting this error with an apparently correct XPath expression. 

[1] https://github.com/synkd/robottelo/blob/cf480c9c8c90905acf4912a04f17d27680e88cee/tests/foreman/ui/test_ansible.py#L78

synkd: Incidentally, using XPath expressions that point to specific table rows is a possible solution to the issue of multiple `attribute_type` variables that Lai inititally raised in this thread. I'm thinking of something like:
```
attribute_types = [
    TextInput("//div[@class='matcher-parent']/fieldset[2]/div[1]/table[1]/tbody[1]/tr[1]/td[1]/div[1]/input[1]"),
    TextInput("//div[@class='matcher-parent']/fieldset[2]/div[1]/table[1]/tbody[1]/tr[2]/td[1]/div[1]/input[1]"),
]
```
The list could contain a number items equal to the largest number of Ansible Variable matchers needed by a test.

omkarkhatavkar: @synkd I think you need to refactor this code. I think adding a table view makes more sense and attaching that to Main View e.g.

```
class MatcherView(View):
    ROOT = '//h2[text()="Specify Matchers  "]/..'
    add_matcher = Text("//a[contains(@class, 'add_nested_fields')]")
    table = SatTable(
        locator=".//table[@class='table white-header']",
        column_widgets={
            'Attribute type': Select(".//select"),
            'Value': TextInput(".//textarea"),
            'Actions': Text(".//a"),
        },
    ) 
```

airgun/views/ansible_variable.py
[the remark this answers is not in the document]
omkarkhatavkar: If I understand this problem, this is a time synchronization issue. To solve this you can take the help of the before_fill method.

you can take this as example https://github.com/SatelliteQE/airgun/blob/b822d70e44a6e6d46b3ec4abac527b0c7efc07b0/airgun/views/host.py#L531

airgun/views/ansible_variable.py
synkd: @omkarkhatavkar Thank you for looking at this. Using a table widget does make sense, since that's how that part of the HTML is structured. However, I'm not clear on what the advantage would be to adding a table view, as opposed to adding a table to the existing `NewAnsibleVariableView`. Could you please explain your thinking there? Is it to separate filling the values in the `Specify Matchers` section from filling the data on the rest of the page?

airgun/views/ansible_variable.py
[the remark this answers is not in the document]
synkd: Thank you, @omkarkhatavkar. I will try this and report back on the results.

airgun/views/ansible_variable.py
omkarkhatavkar: Added more details implementation here https://github.com/synkd/airgun/pull/1, you can modify this behaviour if you like

airgun/views/ansible_variable.py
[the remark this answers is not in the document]
synkd: @omkarkhatavkar was correct about this. With a `before_fill` method added, this is now functioning as expected.

airgun/views/ansible_variable.py
synkd: Thanks to some amazing collab from @omkarkhatavkar, this is now working as expected. I need to do some additional testing to verify that Omkar's changes to the `CustomParameter` widget do not affect other tests, and  need to resolve an issue with my delete method in the Ansible Variable entity where the confirmation dialog is not being handled properly. Once these tasks are completed, this PR will be ready for re-review.

synkd: One additional note: I had to recreate separate methods for creating a variable with and without overrides. When the `Override` checkbox is selected but no values are entered into the section of the form enabled by that checkbox, Satellite throws an error when the form is submitted. This does not strike me as a product bug, so I've added separate methods for those two scenarios (with and without overrides) to the entity.

synkd:
I believe this PR is ready for review again. A couple of notes:
- For some reason, `handle_alert` is not functioning properly in the Ansible variables delete method. I have not been able to find any meaningful differences between its implementation here and in other entities. I ran a few tests locally that use `handle_alert`, and it worked as expected in those cases. I've added the confirmation button to the Ansible variables view as a workaround.
- I've added a `preimport_check` to the Ansible roles entity. This can be used in Ansible variables tests to ensure that we don't try to create an Ansible variable before importing any roles to which the variable could be assigned.
- Omkar pointed out to me that we need to ensure that his changes to the `CustomParameter` widget don't affect any existing tests that use this widget. As a smoke test, I locally ran the `tests/foreman/ui/test_domain.py` module, which includes 4 tests that use the `CustomParameter` widget. 3 of these passed and 1 failed. However, the failed test also failed in the most recent CI run with the same error that I encountered, so I don't believe the failure is related to the changes to the widget. There are a few other tests that use this widget, and I'm happy to do additional testing if necessary.

omkarkhatavkar:
@synkd can you please rebase your branch and see why the CI showing errors.

synkd:
@omkarkhatavkar I had defined the same variable twice in one of the view classes, and Docs Build was failing as a result. The CI checks are passing with that fixed.

airgun/views/ansible_role.py
```diff
@@ -0,0 +1,72 @@
...
+    """
+
+    title = Text("//h1[contains(., text()='Ansible Roles')")
+    import_button = Text("//a[contains(@href, '/ansible_roles/import')]")
+    submit = Button('Submit')
+    total_imported_roles = Text("//span[contains(@class, 'pagination-pf-items-total')]")
+    table = Table(
+        './/table',
+        column_widgets={
+            '▲ Name	': Text("./a"),
```
mshriver, line 30: I think I would raise this as an issue with the UX team, for testability. They have a gchat channel we can engage on instead of going straight to writing a BZ.

mshriver: In the meantime try using `widgetastic.utils.partial_match` instance for the key here, like `partial_match('Name')`. I'm not confident it will work correctly as a dictionary key, but its worth a try.

airgun/views/ansible_role.py
```diff
@@ -0,0 +1,72 @@
...
+            'Hostgroups Count': Text('.//a'),
+        },
+    )
+    pagination = ImportPagination()
+    submit = Button('Submit')
+    cancel = Button('Cancel')
+
+    @property
+    def is_displayed(self):
+        breadcrumb_loaded = self.browser.wait_for_element(self.breadcrumb, exception=False)
```
mshriver, line 67: Is there an exception pattern/loading time that you're concerned about here?

There is `wait_displayed`, which wraps `is_displayed` in a wait_for loop, and we generally discourage doing any waiting in the `is_displayed` implementation.  Page 'readiness' issues should also be dealt with in the browser plugin, not in individual `is_displayed` functions.

airgun/views/ansible_variable.py
```diff
@@ -0,0 +1,129 @@
...
+    def is_displayed(self):
+        return self.title.is_displayed and self.new_variable.is_displayed
+
+
+class MatcherTable(CustomParameter):
+    add_new_value = Text("..//a[contains(text(),'+ Add Matcher')]")
+
+
+class MatcherActions(View):
+    """Interface table has Acttibute type column that contains select field and
```
mshriver, line 47: typo here

airgun/views/ansible_variable.py
```diff
@@ -0,0 +1,129 @@
...
+    text input field."""
+
+    matcher_key = Select(".//select")
+    matcher_value = TextInput(locator=".//input[@class='matcher_value']")
+
+    def fill(self, values):
```
mshriver, line 53: This should not be necessary, what's happening here is the standard fill pattern for a view class. 

If you want a custom exception class for this (not just an exception message like you have), then just check the fill dict for keys before calling `super().fill()`

airgun/views/ansible_variable.py
```diff
@@ -0,0 +1,129 @@
...
+        )
+
+    def expand(self):
+        """Expand the Optional Input Validator section"""
+        if not self.expanded:
+            self.browser.click(self.expand_optional_input_validator, parent=self)
+
+    @property
+    def is_displayed(self):
+        breadcrumb_loaded = self.browser.wait_for_element(self.breadcrumb, exception=False)
```
mshriver, line 124: Same consideration as the other class about using `wait_for` within is_displayed.

airgun/widgets.py
```diff
@@ -916,20 +916,25 @@ class CustomParameter(Table):
 
     add_new_value = Text("..//a[contains(text(),'+ Add Parameter')]")
 
-    def __init__(self, parent, locator=None, id=None, logger=None):
+    def __init__(self, parent, locator=None, id=None, logger=None, column_widgets=None):
```
mshriver, line 919: `Table` has a lot more kwargs than this, so the function signature here should be updated to take `**kwargs` and pull any references from there, instead of trying to maintain an exact signature with the parent class.

airgun/widgets.py
```diff
@@ -938,9 +943,9 @@ def read(self):
         """
         parameters = []
         for row in self.rows():
-            name = row['Name'].widget.read()
-            value = row['Value'].widget.read()
-            parameters.append({'name': name, 'value': value})
+            name = row[self.name].widget.read()
+            value = row[self.value].widget.read()
+            parameters.append({self.name_key: name, self.value_key: value})
```
mshriver, line 948: Are `name_key` and `value_key` doing anything functionally here, other than replacing the use of the string `'name'` and `'value'`?

mshriver: I think it just clicked that all of this is being done to work around the fact that there's a table that is including the sort character.

That doubles my commitment to the idea that we should push back on UX and exclude the sorting indicator from the text content of the column.

mshriver (changes requested):
Looks like some questionable things in the interest of working around sorting behavior that I'd like us to discuss with dev.

airgun/views/ansible_role.py
synkd: Thanks for pointing me to the UX team's channel. I will try the `partial_match` util as well and update with the results here.

airgun/views/ansible_role.py
synkd: I honestly do not recall why I added the `exception=False` here. My `is_displayed` methods were largely borrowed/stolen from other Airgun views, so I have no issue with modifying them if this implementation is not in line with best practices.

airgun/widgets.py
omkarkhatavkar: Yes because of the hard coding done on the table name this was not reusable, I was just trying to make that reusable.

airgun/views/ansible_variable.py
omkarkhatavkar: Yes if the values are coming from the test correctly no need to have this. I was apparently testing something and that added this code. You can add a similar test code that references the fields. 

e.g.  `matcher_section.params.matcher_key`

airgun/views/ansible_variable.py
synkd: addressed

airgun/views/ansible_role.py
synkd: Removed `wait_for_element`.

airgun/views/ansible_variable.py
synkd: Removed the `wait_for_element`.

airgun/widgets.py
synkd: I've attempted to address this but would appreciate feedback on my implementation.

airgun/views/ansible_role.py
synkd: I could not figure out how to get `partial_match` working as a dictionary key, but I came up with another workaround: as long as we're searching for a specific role to delete, there will only be one row in the search results, so `view.table.row()['Actions'].widget.fill('Delete')` works. This is not optimal but I believe it will suffice until the underlying behavior changes (i.e. the UX team decides to either not sort roles by name on import or wraps the sort character in a separate `span` tag).

airgun/widgets.py
```diff
@@ -916,20 +916,25 @@ class CustomParameter(Table):
...
+        column_widgets = column_widgets or {
             'Name': TextInput(locator=".//input[@placeholder='Name']"),
             'Value': TextInput(locator=".//textarea[@placeholder='Value']"),
             'Actions': Text(
                 locator=".//a[@data-original-title='Remove Parameter' "
                 "or @title='Remove Parameter']"
             ),
         }
+        self.name = list(column_widgets.keys())[0]
+        self.name_key = '_'.join(self.name.lower().split(' '))
```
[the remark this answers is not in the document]
synkd: @mshriver Was there a change that you wanted made relating to this comment? It was not totally clear to me.

airgun/entities/ansible_role.py
```diff
@@ -0,0 +1,75 @@
...
+        return view.table.read()
+
+    def delete(self, entity_name):
+        """Delete Ansible Role from Satellite"""
+        # This method currently does not work as the Name column table header
+        # cell includes the `▲` character used for sorting that column
+        view = self.navigate_to(self, 'All')
+        view.search(entity_name)
+        view.table.row()['Actions'].widget.fill('Delete')
+        view.confirm.confirm_dialog.click()
```
mshriver, line 28: You should be able to use `view.dialog.confirm()` here, as the PF4 confirmation dialog modal is on the BaseView as `dialog`.

airgun/entities/ansible_variable.py
```diff
@@ -0,0 +1,74 @@
...
+        """Create a new Ansible variable that is managed by Satellite"""
+        view = self.navigate_to(self, 'New')
+        view.override.fill(True)
+        view.expand()
+        view.matcher_section.before_fill(values)
```
mshriver, line 48: This is slightly odd, the pattern for before_fill/after_fill is such that they shouldn't have to be called directly.

As `view.fill` executes, it will call `view.matcher_section.fill` with `values['matcher_section']`, which in turn should execute `self.before_fill`.

Not a blocking comment, going to keep reading, as I'm guessing this is trying to enforce fill order.

airgun/entities/configgroup.py
```diff
@@ -43,6 +43,7 @@ def delete(self, entity_name):
         view = self.navigate_to(self, 'All')
         view.searchbox.search(entity_name)
         view.table.row(name=entity_name)['Actions'].widget.click(handle_alert=True)
+        view.dialog.confirm()
```
mshriver, line 46: I think you need to rebase against master, this is already in.

airgun/views/ansible_role.py
```diff
@@ -0,0 +1,72 @@
...
+class ImportPagination(Pagination):
+    PER_PAGE_BUTTON_DROPDOWN = ".//div[button[@id='pagination-options-menu-toggle-2']]"
+
+
+class AnsibleRolesView(BaseLoggedInView, SearchableViewMixin):
+    """Main Ansible Roles view. Prior to importing any roles, only the import_button
+    is present, without the search widget or table.
+    """
+
+    title = Text("//h1[contains(., text()='Ansible Roles')")
```
mshriver, line 24: non-blocking, but `normalized-space()` is the best way to do this locator.

https://github.com/SatelliteQE/airgun/issues/591

airgun/views/ansible_role.py
```diff
@@ -0,0 +1,72 @@
...
+    import_button = Text("//a[contains(@href, '/ansible_roles/import')]")
+    submit = Button('Submit')
+    total_imported_roles = Text("//span[contains(@class, 'pagination-pf-items-total')]")
```
mshriver, line 27: This is actually already available as `self.pagination.total_items`.

https://github.com/SatelliteQE/airgun/blob/master/airgun/widgets.py#L1463

airgun/views/ansible_role.py
```diff
@@ -0,0 +1,72 @@
...
+    table = Table(
+        './/table',
+        column_widgets={
+            'Name': Text("./a"),
```
mshriver, line 31: I know there's a long history of discussion about this table definition and its behavior.

Are you trying to define column widgets here for every column header? Usually you would only define the column that actually has an interactable widget.

The `Table` widget otherwise has properties/attributes/functions to deal with the table headers/columns, outside of `column_widgets`.

Non-blocking, but its a bit odd to see `Text` widgets being used here, the table cells are just 'normal' text.

airgun/views/ansible_role.py
```diff
@@ -0,0 +1,72 @@
...
+            'Hostgroups': Text("./a"),
+            'Hosts': Text("./a"),
+            'Imported at': Text("./a"),
+            'Actions': ActionsDropdown("./div[contains(@class, 'btn-group')]"),
+        },
+    )
+    pagination = Pagination()
+    confirm = Pf4ConfirmationDialog()
```
mshriver, line 39: This is in `BaseLoggedInView` as `dialog`, please update to use that widget where necessary.

airgun/views/ansible_variable.py
```diff
@@ -0,0 +1,116 @@
...
+from airgun.widgets import Pf4ConfirmationDialog
+from airgun.widgets import SatSelect
+
+
+class AnsibleVariablesView(BaseLoggedInView, SearchableViewMixin):
+    """Main Ansible Variables view"""
+
+    title = Text("//h1[contains(., text()='Ansible Variables')")
+    new_variable = Text("//a[contains(@href, '/ansible/ansible_variables/new')]")
+    total_variables = Text("//span[@class='pagination-pf-items-total']")
```
mshriver, line 23: Again, available through `self.pagination`, non-blocking.

airgun/views/ansible_variable.py
```diff
@@ -0,0 +1,116 @@
...
+                'Actions': Text(".//a"),
+            },
+        )
+
+        def before_fill(self, values):
+            if not self.params.is_displayed:
+                self.add_matcher.click()
+
+    @property
+    def expand_button(self):
```
mshriver, line 95: doesn't look like this is actually used anywhere. Its also slightly risky naming, as it returns an element instance, not a button widget instance.

airgun/views/dashboard.py
```diff
@@ -124,6 +124,11 @@ def fill(self, values):
                 'No. of Tasks'
             ].widget.click()
 
+    @View.nested
```
mshriver, line 127: rebase should remove this from the diff view, its already in.

airgun/widgets.py
```diff
@@ -916,31 +916,44 @@ class CustomParameter(Table):
...
+        column_widgets = kwargs.get('column_widgets') or {
             'Name': TextInput(locator=".//input[@placeholder='Name']"),
             'Value': TextInput(locator=".//textarea[@placeholder='Value']"),
             'Actions': Text(
                 locator=".//a[@data-original-title='Remove Parameter' "
                 "or @title='Remove Parameter']"
             ),
         }
-        super().__init__(parent, locator=locator, logger=logger, column_widgets=column_widgets)
+        self.name = list(column_widgets.keys())[0]
```
mshriver, line 939: This could throw `IndexError` during instantiation if column_widgets weren't passed.

Have you checked all uses of this widget class against this requirement?

I think you'll want to make this safer by having some checking, or using try/except to set these instance attributes to `None` if column_widgets doesn't have the expected length/field format.

airgun/widgets.py
```diff
@@ -916,31 +916,44 @@ class CustomParameter(Table):
...
+        self.name_key = '_'.join(self.name.lower().split(' '))
+        self.value = list(column_widgets.keys())[1]
+        self.value_key = '_'.join(self.value.lower().split(' '))
+
+        super().__init__(
```
mshriver, line 944: Good work handling the kwargs above!

This call to super() isn't quite right though - no need to try and match up kwargs with the parent class, just pass the whole thing.

```
super().__init__(parent, **kwargs)
```

mshriver (commented):
Several non-blocking comments that you may or may not want to address in this PR.

One or two blocking comments, but nothing major.

Great work here @synkd this is a great addition to airgun!

latran (commented):
Some general comments and some clarifications.  Overall looks good.  If there's some test cases written for this, I would like to see them if possible.

airgun/views/ansible_role.py
```diff
@@ -0,0 +1,70 @@
...
+from airgun.views.common import SearchableViewMixin
+from airgun.widgets import ActionsDropdown
+from airgun.widgets import Pagination
+
+
+class ImportPagination(Pagination):
+    PER_PAGE_BUTTON_DROPDOWN = ".//div[button[@id='pagination-options-menu-toggle-2']]"
+
+
+class AnsibleRolesView(BaseLoggedInView, SearchableViewMixin):
```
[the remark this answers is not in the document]
synkd: This class represents the `Configure > Roles` page (i.e. satellite.com/ansible/ansible_roles). When no Ansible roles have been imported and this page is visited, only the import button is visible. I tried to convey this in the class's docstring. 

When writing this view, I decided that it was not necessary to have separate classes for the page representing its state before and after importing roles. The only available action on this page before importing roles is importing roles, which is available via the `import_button` widget in this class. The other widgets in this class are present after roles are imported.

airgun/views/ansible_role.py
```diff
@@ -0,0 +1,70 @@
...
+        },
+    )
+    pagination = Pagination()
+
+    @property
+    def is_displayed(self):
+        return self.title.is_displayed and self.import_button.is_displayed
+
+
+class AnsibleRolesImportView(BaseLoggedInView):
```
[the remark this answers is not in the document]
synkd: Yes, this is the `Roles > Changed Ansible Roles` page, which is loaded after the `import_button` widget in `AnsibleRolesView` is clicked.

airgun/entities/ansible_role.py
```diff
@@ -0,0 +1,75 @@
...
+    def delete(self, entity_name):
+        """Delete Ansible Role from Satellite"""
+        # This method currently does not work as the Name column table header
+        # cell includes the `▲` character used for sorting that column
+        view = self.navigate_to(self, 'All')
+        view.search(entity_name)
+        view.table.row()['Actions'].widget.fill('Delete')
+        view.dialog.confirm_dialog.click()
+        view.flash.assert_no_error()
+        view.flash.dismiss()
```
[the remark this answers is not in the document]
synkd: Ack, that is good to know. When that happens, I can work with Sam (the new Ansible component owner) to refactor this  if necessary.

airgun/entities/ansible_role.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def search(self, value):
+        """Search for existing Ansible Role"""
+        view = self.navigate_to(self, 'All')
+        view.search(value)
+        return view.table.read()
+
+    def delete(self, entity_name):
+        """Delete Ansible Role from Satellite"""
+        # This method currently does not work as the Name column table header
```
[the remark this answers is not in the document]
synkd: Good point - I will change this to reflect the reason behind doing search first, then delete. I believe the Satellite UX team is working on changing this in a future release - likely by wrapping the `▲` character in a separate `span` tag from the column header.

airgun/views/ansible_variable.py
```diff
@@ -0,0 +1,116 @@
...
+class MatcherTable(CustomParameter):
+    add_new_value = Text("..//a[contains(text(),'+ Add Matcher')]")
+
+
+class MatcherActions(View):
+    """Interface table has Attribute type column that contains select field and
+    text input field."""
+
+    matcher_key = Select(".//select")
+    matcher_value = TextInput(locator=".//input[@class='matcher_value']")
```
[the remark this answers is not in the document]
synkd: Ack, in the interest of getting this PR merged, I will leave this as is for now. If a future test requirement needs to access that `delete` button, it can be added at that time.

airgun/entities/ansible_role.py
```diff
@@ -0,0 +1,79 @@
...
+        # This method currently relies on searching for a specific role as
+        # directly accessing the 'Name' column is not possible due to the
+        # presence of the `▲` character used for sorting that column in the table
+        # header cell. The Satellite UX team is planning to address this in a
+        # future release, likely by wrapping the sort character in a separate
+        # span tag from the column header text.
+        view = self.navigate_to(self, 'All')
+        view.search(entity_name)
+        view.table.row()['Actions'].widget.fill('Delete')
+        view.dialog.confirm_dialog.click()
```
mshriver, line 32: You can now drop this line.

airgun/widgets.py
synkd: @mshriver So I've tried changing the super() call to `super().__init__(parent, **kwargs) both in its existing location and with it moved to the first line in the this `__init__` method, but both are introducing errors in some of the other tests that use the `CustomParameter` widget. As an example, `tests/foreman/ui/test_domain.py::test_positive_set_parameter` fails with `TypeError: Table.__init__() got an unexpected keyword argument 'id'`. The same test passes with the current super() call. I'm wondering if a change to the `Table` widget would be necessary for this to work, but I'm a bit leery of making a change to such a widely-used widget. What do you think is the best approach here?

synkd: So I'm still having issues with this. Making the changes that we discussed has my Ansible tests still passing, but it's failing on tests that use views that instantiate the `CustomParameter` class with an `id` argument (these are the UI test modules `test_operatingsystem.py`, `test_host.py`, and `test_domain.py`). The issue seems to be that `locator` in the table widget's `__init__` method is a positional argument with no default value. However, passing a locator argument to `super().__init__` does not resolve the issue. Instead, the Ansible test now fails, and the tests that use `CustomParameter` with `id` still fail, now because the widget seems to be evaluating to a `NoneType` object. Any ideas on how to work past this?

Here are two of the variations that I tried along with links to the tracebacks that I encountered:

```
    def __init__(self, parent, **kwargs):
        """Supports initialization via ``locator=`` or ``id=``"""
        if (
            kwargs.get('locator')
            and kwargs.get('id')
            or not kwargs.get('locator')
            and not kwargs.get('id')
        ):
            raise ValueError('Please specify either locator or id')
        locator = kwargs.get('locator') or f".//table[@id='{kwargs.pop('id')}']"
        super().__init__(parent, **kwargs)
```
results:
ansible test passes
error in host test: http://rx-paste.usersys.redhat.com/view/834c2c5f


```
    def __init__(self, parent, **kwargs):
        """Supports initialization via ``locator=`` or ``id=``"""
        if (
            kwargs.get('locator')
            and kwargs.get('id')
            or not kwargs.get('locator')
            and not kwargs.get('id')
        ):
            raise ValueError('Please specify either locator or id')
        locator = kwargs.get('locator') or f".//table[@id='{kwargs.pop('id')}']"
        super().__init__(parent, locator=locator, **kwargs)

```
results:
error in ansible test: http://rx-paste.usersys.redhat.com/view/2842dfc7
error in host test: http://rx-paste.usersys.redhat.com/view/ee7b053b

PR #643, opened 2021-12-09
updated the navigation with retry

omkarkhatavkar:
### Description 
Currently we are facing navigation issue in the selenium grid 4 with Satellite. Mouse Hover Left panel on Satellite is not able catch effectively with Selenium Grid 4, could be related to DOM issue, After the browser refresh seems to helping to solve the issue. This is temporary solution as we need to move to Selenium  Grid 4.

### Result 

- Look at the Stage Jenkins Job : okhatavka-ui-testing
- pytest xdist worker count : 8

`87 failed, 360 passed, 20 skipped, 5963 warnings, 22 errors in 22990.85s (6:23:10) `

#### Other changes 

- Need update in Selenium Module Version
- Update Configuration file with new Selenium Grid URL

### DIFF
--- a/airgun/utils.py
+++ b/airgun/utils.py
@@ -1,3 +1,7 @@
+import functools
+import time
+
+
 def merge_dict(values, new_values):
     """Update dict values with new values from new_values dict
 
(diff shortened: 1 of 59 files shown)


latran (approved):
ack for temporary solution for now.

synkd (approved):
Ack to the temporary solution.

jameerpathan111 (approved):

jameerpathan111:
Merging as this has more than two ack.

airgun/utils.py
```diff
@@ -69,3 +73,22 @@ def get_widget_by_name(widget_root, widget_name):
...
+def retry_navigation(method):
+    """Decorator to invoke method one or more times, if TimedOutError is raised."""
+
+    @functools.wraps(method)
+    def retry_wrapper(*args, **kwargs):
+        attempts = 3
+        for i in range(attempts):
+            try:
+                return method(*args, **kwargs)
+            except (Exception):
```
jyejare, line 87: Dont you think Exception here is `too open` and it's not very limited to TimeoutError? 

This will end up re-executing the tests that are not failing with `TimeoutError` and taking more time to finish after error!

swadeley: I agree, in any case test should conform to description (and please check description, is it TimedOutError or TimeoutError? I think the latter.)

PR #645, opened 2021-12-17
Fix puppet-related URLs

vsedmik:
The puppet-related URLs changed with movement into a plugin in 7.0.

### DIFF
--- a/airgun/entities/configgroup.py
+++ b/airgun/entities/configgroup.py
@@ -9,7 +9,7 @@ from airgun.views.configgroup import ConfigGroupsView
 
 
 class ConfigGroupEntity(BaseEntity):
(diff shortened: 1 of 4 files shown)


jameerpathan111 (approved):

swadeley:
Hi

I don't see a 6.10 branch and seems to me this should not be in master before such a branch is made.

jameerpathan111:
@swadeley you are right, we need this change for 7.0 only.

vsedmik:
Yes, this is intended for 7.0 only

### DO NOT MERGE until 6.10 is branched!

swadeley (approved):
ACK

omkarkhatavkar (approved):
ACK

PR #648, opened 2021-12-23
Fix search for Satellite 7.0

jameerpathan111:
This PR fixes locators for Search(`searchbar`) which is causing some of the Satellite UI tests to fail.

### DIFF
--- a/airgun/widgets.py
+++ b/airgun/widgets.py
@@ -634,8 +634,10 @@ class Search(Widget):
     )
     search_field = TextInput(
         locator=(
-            ".//input[@data-autocomplete-id='searchBar' or contains(@placeholder, 'Search') or "
-            "contains(@class, 'search-input')]"
+            ".//input[@id='search' or contains(@placeholder, 'Filter') or "
+            "@ng-model='table.searchTerm' or contains(@ng-model, 'Filter') or "
+            "@data-autocomplete-id='searchBar' or contains(@placeholder, 'Search') "
+            "or contains(@class, 'search-input')]"
         )
     )
     search_button = Text(
--- a/airgun/widgets.py
+++ b/airgun/widgets.py
@@ -634,13 +634,13 @@ class Search(Widget):
     )
     search_field = TextInput(
         locator=(
-            ".//input[@id='search' or contains(@placeholder, 'Filter') or "
-            "@ng-model='table.searchTerm' or contains(@ng-model, 'Filter')]"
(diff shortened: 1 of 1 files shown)


jameerpathan111:
It'd be great If someone can confirm that this fixes search related issues for them as well.

jyejare:
@jameerpathan111 Run PRT if everything is settled from infra end :)

jameerpathan111:
@jyejare there are lot of other issues which prevents from running entire test, instead here's the output from 
things I tried:

```
(Pdb) session.activationkey.search('ak')
[{'Name': 'ak', 'Host Limit': '0 out of Unlimited', 'Environment': '', 'Content View': ''}]
(Pdb) session.syncplan.search('qe')
[{'Name': 'qe', 'Description': '', 'Original Sync Date': 'December 24, 2021, 10:35 AM', 'Sync Enabled': 'true', 'Interval': 'Hourly', 'Next Sync': 'December 24, 2021, 11:35 AM'}]
(Pdb) session.oscapcontent.search('Red Hat jre default content')
[{'Title': 'Red Hat jre default content', 'Filename': 'ssg-jre-ds.xml', 'Created at': '7 days ago', 'Actions': ['Delete']}]
(Pdb) session.host.search('host.example.com')
[{0: False, 'Power': '', 'Name': 'host.example.com', 'Recommendations': '', 'Operating system': 'RHEL Server 7.9', 'Model': 'RHEL', 'Host group': '', 'Last report': '', 'Actions': ['Clone', 'Delete']}]
```
`Activation key` and `Sync plan` entities uses old search bar and `Host` and `Scap Content` uses the new one.
So for me, it's working for both new and old search bar.

jyejare:
@jameerpathan111 Cool! Looks good to me btw and yes PRT for 7.0 will be challenging until we resolve some of the issues we are seeing in this week especially Jenkins, Gitlab and OCP outages/issues!

airgun/widgets.py
```diff
@@ -635,12 +635,13 @@ class Search(Widget):
     search_field = TextInput(
         locator=(
             ".//input[@id='search' or contains(@placeholder, 'Filter') or "
-            "@ng-model='table.searchTerm' or contains(@ng-model, 'Filter')]"
+            "contains(@ng-model, 'Filter') or @data-autocomplete-id='searchBar' "
```
omkarkhatavkar, line 638: No need to remove `ng-model="table.searchTerm"` as I can see that is still valid in the content hosts search

omkarkhatavkar (approved):

jyejare (approved):

PR #649, opened 2021-12-23
PF4 confirmation dialog for Satellite 7.0

jameerpathan111:
Satellite 7.0 has new Confirmation dialog for deleting entities. This PR adds new ConfirmationBox widget to support it.

### DIFF
--- a/airgun/widgets.py
+++ b/airgun/widgets.py
@@ -1063,7 +1063,7 @@ class ConfirmationDialog(Widget):
         do_not_read_this_widget()
 
 
-class ConfirmationBox(ConfirmationDialog):
+class Pf4ConfirmationDialog(ConfirmationDialog):
     """PF4 confirmation dialog with two buttons and close 'x' button in the
     right corner."""
 
--- a/airgun/widgets.py
+++ b/airgun/widgets.py
@@ -21,6 +21,7 @@ from widgetastic_patternfly import FlashMessages
 from widgetastic_patternfly import Kebab
 from widgetastic_patternfly import VerticalNavigation
 from widgetastic_patternfly4.ouia import BaseSelect
+from widgetastic_patternfly4.ouia import Button as PF4Button
 from widgetastic_patternfly4.ouia import ContextSelector as OUIAContextSelector
 from widgetastic_patternfly4.ouia import Dropdown
 
@@ -1060,6 +1061,16 @@ class ConfirmationDialog(Widget):
     def read(self):
         """Widgets has no fields to read"""
         do_not_read_this_widget()
+
+
+class ConfirmationBox(ConfirmationDialog):
+    """PF4 confirmation dialog with two buttons and close 'x' button in the
+    right corner."""
+
+    ROOT = '//div[@id="app-confirm-modal" or @data-ouia-component-type="PF4/ModalContent"]'
+    confirm_dialog = PF4Button('OUIA-Generated-Button-danger-1')
+    cancel_dialog = PF4Button('OUIA-Generated-Button-link-1')
+    discard_dialog = PF4Button('OUIA-Generated-Button-plain-1')
 
 
 class LCESelector(GenericLocatorWidget):


airgun/widgets.py
```diff
@@ -1062,6 +1063,16 @@ def read(self):
         do_not_read_this_widget()
 
 
+class ConfirmationBox(ConfirmationDialog):
```
omkarkhatavkar, line 1066: I would personally keep the name as Pf4ConfirmationDialog kind of pattern, So Whenever we would like to migrate completely to PF4 then it will be pretty easy to identify it. @latran @jyejare and @JacobCallahan  wanted to know thoughts about it

omkarkhatavkar: @jameerpathan111 can you please also know where are these are used/added in Satellite 7.0? add supporting code.

jameerpathan111: okay, I will try to find where we need to use this new dialog and raise a separate pr for it.

jyejare: @omkarkhatavkar @jameerpathan111 Sorry for my less knowledge here but does satellite 7.0 is the mixture of PF4 and existing PF version? 

if it's only PF4 :
    can we stop prepending PF4 altogether in airgun cause PF4 is default/only and we not to explicitely mention it ? e.g `Button` instead of `PF4Button`?
else:
    Since there are earlier version dialogs as well then we should prepend the PF4 becuase it would be easy for tomorrow to remove version < PF4 widgets/views/entities to remove.

airgun/widgets.py
```diff
@@ -1062,6 +1063,16 @@ def read(self):
...
+    """PF4 confirmation dialog with two buttons and close 'x' button in the
+    right corner."""
+
+    ROOT = '//div[@id="app-confirm-modal" or @data-ouia-component-type="PF4/ModalContent"]'
```
jyejare, line 1070: The `or` should not match with the elements that using the same `data-ouia-component-type`. So combine it with `and` instead:
```suggestion
    ROOT = '//div[@id="app-confirm-modal" and @data-ouia-component-type="PF4/ModalContent"]'
```

OR just stick with id.

airgun/widgets.py
```diff
@@ -1062,6 +1063,16 @@ def read(self):
...
+    confirm_dialog = PF4Button('OUIA-Generated-Button-danger-1')
```
jyejare, line 1071: Wonder why the identifier of confirm button has `danger` word in it, I hope you didnt mismatch the variables here for different dialogs.

airgun/widgets.py
```diff
@@ -21,6 +21,7 @@
 from widgetastic_patternfly import Kebab
 from widgetastic_patternfly import VerticalNavigation
 from widgetastic_patternfly4.ouia import BaseSelect
+from widgetastic_patternfly4.ouia import Button as PF4Button
```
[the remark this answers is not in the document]
jyejare: I see you have used it in PF4 Confirmation dialogue, so u might change it to `DialogButton` kinda naming!

jyejare (approved):
ACK with questions/comments/suggestions.

All comments are non-blocking 1

airgun/widgets.py
omkarkhatavkar: By default, we don't have PF4 everywhere in Satellite 7.0. We are migrating from PF3 to PF4. I would agree on the part if the PF4 elements are not used in combination with PF3 elements then make PF4 as default and no need to add prefix 'PF4' in that case.

airgun/widgets.py
omkarkhatavkar: I think it's better to use `or` in case of elements where you are having more robust locator strategies. even in the case, the id changed it will help to locate the element. My guess is that @jameerpathan111 was doing that only   

see e.g. 
image

airgun/widgets.py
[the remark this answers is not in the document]
jameerpathan111: @jameerpathan111 I don't think all buttons are PF4 Buttons therefore we can't use `Button` name here because it's already in use.
and for changing it to `DialogButton` is not a good idea either because it has to be general name which can be 
used with other PF4 buttons already present on the page.

airgun/widgets.py
jameerpathan111: @jyejare yes, not all buttons are PF4 buttons.

airgun/widgets.py
jameerpathan111: @jyejare I agree with omkar here.

airgun/widgets.py
jameerpathan111: nah, it's really like that :D

airgun/widgets.py
jyejare: @omkarkhatavkar I was just worried abt `data-ouia-component-type` shouldnt pick any other element! But anyways its a non blocking comment !

airgun/widgets.py
jameerpathan111: @omkarkhatavkar changed widget name to `Pf4ConfirmationDialog`.

omkarkhatavkar (approved):

PR #654, opened 2022-01-04
suppressing the exception to timeout

omkarkhatavkar:


airgun/utils.py
```diff
@@ -84,7 +84,7 @@ def retry_wrapper(*args, **kwargs):
         for i in range(attempts):
             try:
                 return method(*args, **kwargs)
-            except (Exception):
+            except (TimedOutError):
```
jyejare, line 87: Is it a TimeoutError (python in-built) or TimedOutError (Needs to be imported from where its originated , can you import it first?)

jyejare (changes requested):

airgun/utils.py
omkarkhatavkar: It should be coming from `waitfor`

jyejare (approved):

jameerpathan111 (approved):

PR #657, opened 2022-01-12
Increase timeout in get_webconsole_content

shubhamsg199:
1] Increasing timeout in `get_webconsole_content` to 10s which is required by  the test `test_positive_cockpit`


omkarkhatavkar (approved):
ACK

tstrych (approved):
ACK

PR #659, opened 2022-01-13
Update locators

shweta83:
Updated Search Locator and Locator for PopOver Widget

### DIFF
--- a/airgun/widgets.py
+++ b/airgun/widgets.py
@@ -2181,7 +2181,7 @@ class PopOverWidget(View):
     """
 
     ROOT = '.'
-    column_value = Text(".//span[contains(@class, 'editable')]")
+    column_value = Text(".//div[contains(@class,'ellipsis editable')]")
     pop_over_view = PopOverModalView()
 
     def fill(self, item):


airgun/widgets.py
```diff
@@ -635,7 +635,7 @@ class Search(Widget):
     )
     search_field = TextInput(
         locator=(
-            ".//input[@id='search' or contains(@placeholder, 'Filter') or "
+            ".//input[@id='search' or contains(@placeholder, 'Search') or "
             "@ng-model='table.searchTerm' or contains(@ng-model, 'Filter') or "
```
jameerpathan111, line 639: @shweta83 this change is not required, see line number 640. It already covers it(`contains(@placeholder, 'Search')`).

jameerpathan111 (changes requested):

airgun/widgets.py
shweta83: Yes. I had old airgun so It was wasn't there. I undo that.

latran:
@shweta83 is there a particular test case that you can associate with this.  I'm not sure the change surrounding this PR without a good context of where its usage is.

shweta83:
> @shweta83 is there a particular test case that you can associate with this. I'm not sure the change surrounding this PR without a good context of where its usage is.

@latran All the UI tests for settings component access settings page in Administer -> Settings and when we change any setting, popover widget opens.  
https://github.com/SatelliteQE/robottelo/blob/b0e7f8fe798cdd4f93f6c16237a4f0167a850dd6/tests/foreman/ui/test_settings.py#L75

latran (approved):
ACK

airgun/widgets.py
```diff
@@ -2181,7 +2181,7 @@ class PopOverWidget(View):
     """
 
     ROOT = '.'
-    column_value = Text(".//span[contains(@class, 'editable')]")
+    column_value = Text(".//div[@class='ellipsis editable']")
```
omkarkhatavkar, line 2184: image

A Better Locator would be using contains as there are empty ones. e.g. $x("//div[contains(@class,'ellipsis editable')]")

omkarkhatavkar (approved):
ACK

omkarkhatavkar (approved):

PR #663, opened 2022-01-25
Upgrading to selenium grid 4 needs upgrade in the version 

omkarkhatavkar:
$subject

> pip list | grep selenium
selenium                4.1.0
selenium-smart-locator  0.2.0
> pip list | grep pattern
widgetastic.patternfly  1.3.4
widgetastic.patternfly4 0.23.2.dev2+gcb3bf05

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,12 +21,11 @@ setup(
         'navmazing',
         'python-box',
         'pytest',
-        'selenium',
         'wait_for',
         'webdriver-kaifuku',
         'widgetastic.core',
         'widgetastic.patternfly',
-        'widgetastic.patternfly4',
+        'widgetastic.patternfly4 @ git+https://github.com/RedHatQE/widgetastic.patternfly4.git@selenium4#egg=widgetastic.patternfly4',  # noqa
     ],
     packages=find_packages(exclude=['tests*']),
     package_data={'': ['LICENSE']},


swadeley (approved):
ACK

jameerpathan111 (approved):

setup.py
```diff
@@ -21,7 +21,7 @@
         'navmazing',
         'python-box',
         'pytest',
-        'selenium',
+        'selenium==4.1.0',
```
jyejare, line 24: Currently, selenium is being installed in the latest version, but you are trying to freeze this to a specific version!  This comes an additional overhead of upgrading it every time to the suitable version!

Why not stick to the existing approach? Does `selenium` installs a different version than selenium 4 ?

jyejare (commented):
ACK pending question !

setup.py
omkarkhatavkar: This is the latest version where I have tested it with rehat image. I think this will have not any value if I guess https://github.com/RedHatQE/widgetastic.patternfly4/pull/128 PR get merged

omkarkhatavkar: > pip list | grep selenium
selenium                4.1.0
selenium-smart-locator  0.2.0
> pip list | grep pattern
widgetastic.patternfly  1.3.4
widgetastic.patternfly4 0.23.2.dev2+gcb3bf05

PR #668, opened 2022-02-14
adding the fix versions for widgetastic core and patternfly

omkarkhatavkar:


ogajduse:
I am just wondering whether https://python-poetry.org/ or https://pipenv.pypa.io/ would help our dependency management as they are using the lock file mechanism.

omkarkhatavkar:
This PR will be only valid when widgetastic PF4 moves to selenium4.

PR #670, opened 2022-02-21
new cv - create cv and search

latran:
First few working functions for the new cv.  I've tested this locally it works fine with creating cv/ccv and searching.


airgun/entities/new_contentview.py
```diff
@@ -0,0 +1,47 @@
...
+from airgun.views.new_contentview import NewContentViewTableView
+
+
+class NewContentViewEntity(BaseEntity):
+    endpoint_path = '/content_views'
+
+    def create(self, values, composite=False):
+        """Create a new content view"""
+        view = self.navigate_to(self, 'Create')
+        if composite is True:
```
omkarkhatavkar, line 16: `if composite` is enough

airgun/widgets.py
```diff
@@ -1235,6 +1322,11 @@ class EditableEntry(GenericLocatorWidget):
     save_button = Text(".//button[text()='Save']")
     cancel_button = Text(".//button[span[text()='Cancel']]")
     entry_value = Text(".//span[contains(@class, 'editable-value')]")
+    # PF4 editable entry
+    pf4_edit_button = Text("//button[@aria-label='edit name']")
```
omkarkhatavkar, line 1326: Are these PF4 elements ? I see you are using patternfly3 here ?

airgun/widgets.py
```diff
@@ -677,6 +679,91 @@ def search(self, value):
...
+
+    def after_fill(self):
+        """Ensure widgets appears after filling out searchbox"""
+        wait_for(
+            lambda: self.clear_button.is_displayed is True,
+            timeout=300,
+            delay=1,
+            logger=self.logger
+        )
+        #TODO: Might need to do sleeper or another wait_for table
```
omkarkhatavkar, line 764: Are you suppose to add anything here ?

airgun/widgets.py
```diff
@@ -677,6 +679,91 @@ def search(self, value):
...
+        """Ensure widgets appears after filling out searchbox"""
+        wait_for(
+            lambda: self.clear_button.is_displayed is True,
+            timeout=300,
+            delay=1,
+            logger=self.logger,
+        )
+
+
+class PF4Search(Search):
```
omkarkhatavkar, line 724: I think this is duplicate class

latran:
@omkarkhatavkar I'll make sure the changes at a later time and putting this on draft.  This is because I've filed BZ for UX and katello team to add ouia-ID to the cv table and elsewhere.  Once that's in a snap, I'll update this PR.

latran:
@omkarkhatavkar please give this another review.

mshriver:
@latran failing checks, please run pre-commit locally.

airgun/widgets.py
```diff
@@ -2378,3 +2422,33 @@ def fill(self, value):
 
     def read(self):
         return self.selected
+
+
+class SatPatternflyTable(BasePatternflyTable, Table):
```
mshriver, line 2427: I think you should use `PatternflyTable` here instead of trying to include both.

airgun/widgets.py
```diff
@@ -2378,3 +2422,33 @@ def fill(self, value):
...
+        parent,
+        column_widgets=None,
+        assoc_column=None,
+        rows_ignore_top=None,
+        rows_ignore_bottom=None,
+        top_ignore_fill=False,
+        bottom_ignore_fill=False,
+        logger=None,
+    ):
+        self.component_type = "PF4/Table"
```
mshriver, line 2439: This is used for OUIA type widgets, why use the non-OUIA base classes?

airgun/views/common.py
```diff
@@ -445,6 +446,38 @@ def search(self, query):
...
+    Note that class which uses this mixin should have :attr: `table` attribute.
+    """
+
+    searchbox = PF4Search()
+    welcome_message = Text("//div[@class='blank-slate-pf' or @id='welcome']")
+
+    def search(self, query):
+        """Perform search using searchbox on the page and return table
+        contents.
+        :param str query: search query to type into search field. E.g. ``foo``
```
JacobCallahan, line 462:
```suggestion

        :param str query: search query to type into search field. E.g. ``foo``
```

JacobCallahan (changes requested):
Might help with sphinx

airgun/widgets.py
latran: Ack.  Will try that.  Prior to this `self.component_id = component_id` didn't have static ouia-id so it was constantly failing in recognizing the table and so we switched over to not rely on that.  Will switch it back and retry.

latran: It seems this is also used in `ouia.py` so I have deleted mine.

airgun/entities/contentview_new.py
```diff
@@ -0,0 +1,48 @@
...
+from airgun.views.contentview_new import NewContentViewTableView
+
+
+class NewContentViewEntity(BaseEntity):
+    endpoint_path = '/content_views'
+
+    def create(self, values, composite=False):
+        """Create a new content view"""
+        view = self.navigate_to(self, 'New')
+        if composite:
```
mshriver, line 17: Why is this filled separately from the rest of `values`?   It should fill naturally as part of filling the view.

If it needs to be set before any other values are set, I would use `before_fill` for that, because it's related to the view behavior, not the entity creation behavior.

airgun/views/contentview_new.py
```diff
@@ -0,0 +1,62 @@
...
+from airgun.views.common import BaseLoggedInView
+from airgun.views.common import SearchableViewMixinPF4
+
+
+class NewContentViewTableView(BaseLoggedInView, SearchableViewMixinPF4):
+    title = Text('.//h1[contains(@data-ouia-component-id, "OUIA-Generated-Text")]')
+    create_content_view = PF4Button('create-content-view')
+    table = ExpandableTable(
+        component_id='content-views-table',
+        column_widgets={
```
mshriver, line 17: Are any of these actually widgets? If they're all Text, there should be no need to define column_widgets.

airgun/views/contentview_new.py
```diff
@@ -0,0 +1,62 @@
...
+            'Name': Text('./a'),
+            'Last Published': Text('./a'),
+            'Last task': Text('.//a'),
+            'Latest version': Text('.//a'),
+        },
+    )
+
+    @property
+    def is_displayed(self):
+        return self.browser.wait_for_element(self.create_content_view, exception=False) is not None
```
mshriver, line 27: Its not a very good pattern to use `wait_for_element` within `is_displayed`, as `is_displayed` gets wrapped by `wait_displayed`.  So you end up stacking wait conditions.

Separately its good to check multiple widgets on the page, not just one. Does the page only show the create_content_view button when there are no content views? is the table always displayed or only after the first content view is created?

airgun/views/contentview_new.py
```diff
@@ -0,0 +1,62 @@
...
+    composite = Text('//div[contains(@id, "composite")]')
+    solve_dependencies = Checkbox(id='dependencies')
+    import_only = Checkbox(id='importOnly')
+    auto_publish = Checkbox(id='auto_publish')
+    submit = PF4Button('create-content-view-form-submit')
+    cancel = PF4Button('create-content-view-form-cancel')
+
+    @property
+    def is_displayed(self):
+        return self.browser.wait_for_element(self.title, exception=False) is not None
```
mshriver, line 46: Again I would not use wait_for_element, and I'd probably include a check for title and at least one of the fields being displayed.

airgun/views/contentview_new.py
```diff
@@ -0,0 +1,62 @@
...
+
+    def after_fill(self, value):
+        """Ensure 'Create content view' button is enabled after filling out the required fields"""
+        wait_for(
+            lambda: self.submit.is_displayed is True, timeout=300, delay=1, logger=self.logger
```
mshriver, line 51: `is True` is redundant here. wait_for's fail_condition is going to be `False`, so its looping on `(self.submit.is_displayed is True) == False`, which is the same as just looping on `self.submit.is_displayed == False`, since `is_displayed` returns a boolean.

That aside, there's no need for all this. Just call `self.submit.wait_displayed()`

airgun/views/contentview_new.py
```diff
@@ -0,0 +1,62 @@
...
+        )
+
+    def wait_animation_end(self):
```
mshriver, line 54: I don't see this function getting used anywhere? I'm thinking its not necessary either, the caller can look for a the expected view to be displayed, or wait for the create view to not be displayed.

airgun/widgets.py
```diff
@@ -701,6 +701,48 @@ def search(self, value):
             self.search_button.click()
 
 
+class PF4Search(Search):
+    """PF4 Searchbar for table filtering"""
+
+    ROOT = '//div[@role="combobox" or @aria-haspopup="listbox"]'
```
mshriver, line 707: Are we sure this is a PF4 component?

Looking at the PF4 reference components, I'm not seeing any use of either of these attributes in the search box div.

The PF4 Search component is in beta still, but uses the class `pf-c-search-input` for the root div. 

The text input field uses the class `pf-c-search-input__text-input`.

airgun/widgets.py
```diff
@@ -2138,7 +2180,7 @@ def fill(self, value):
         self.clear()
         super().fill(value)
         self.browser.plugin.ensure_page_safe()
-        self.browser.send_keys_to_focused_element('\t')
```
mshriver, line 2141: This is undoing the fix applied in #718, but the relevant PR in foreman isn't in yet, why is this being backed out?

mshriver: https://github.com/SatelliteQE/airgun/pull/718/files#r904462035

mshriver (changes requested):
Some blocking work/discussions.

airgun/views/contentview_new.py
latran: The table is displayed only after the cv is created, otherwise there's no table.  And the page shows the `create_content_view` button regardless if there's a cv or not.

airgun/widgets.py
latran: adding it back in.

airgun/widgets.py
```diff
@@ -2368,13 +2410,3 @@ def _clickable_el(self):
         if not self.selected:
             locator = self.OFF_TOGGLE
         return self.browser.element(locator=locator)
-
-    def fill(self, value):
```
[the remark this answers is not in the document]
latran: I deleted the wrong thing.  Added it back in again.

airgun/entities/contentview_new.py
latran: @mshriver This is fixed with your suggestion a few weeks back.  I'm having issues testing this with the correct syntax.  I'm using this and it's not correct: `session.contentview_new.create({'name': 'sample', 'composite_view': {True}})`.  What's the correct syntax?

airgun/views/contentview_new.py
latran: This is from the old structure.  I'm not sure what how to construct the "correct" way one as the column widgets are for columns that have links, which these do.

airgun/widgets.py
latran: This definitely needs to be revisit.  Maybe we can meet and discuss.

airgun/views/contentview_new.py
mshriver: Ah! ack, the widgets are needed to make the cell content clickable, cool.

airgun/views/common.py
```diff
@@ -445,6 +446,39 @@ def search(self, query):
...
+            raise AttributeError(
+                f'Class {self.__class__.__name__} does not have attribute "table". '
+                'SearchableViewMixin only works with views, which have table for results. '
+                'Please define table or use custom search implementation instead'
+            )
+        if not self.is_searchable():
+            return None
+        self.searchbox.search(query)
+        # TODO: may need a sleeper or a wait_for
+        time.sleep(3)
```
mshriver, line 478: wait_for is always better if there's something on the table you can use.

`self.table.wait_displayed` maybe?

omkarkhatavkar (approved):

omkarkhatavkar:
docs are failing and @latran do you have test results certifying this change? code wise looks okay to me.

latran:
@omkarkhatavkar I'm running into an issue where after searching, it's not reading the table when testing today.  Will try to resolve it with Mike tomorrow.

latran:
closing this due to issues with other PRs.  Opening another cleaner one so there's no conflict when commit changes.

PR #671, opened 2022-02-22
Add method for checking whether Cloud Connector is already configured.

lhellebr:
Required by https://github.com/SatelliteQE/robottelo/pull/9387

### DIFF
--- a/airgun/views/cloud_inventory.py
+++ b/airgun/views/cloud_inventory.py
@@ -115,6 +115,9 @@ class CloudInventoryListView(BaseLoggedInView):
     obfuscate_ips = Switch('.//label[@for="rh-cloud-switcher-obfuscate_inventory_ips"]')
     exclude_packages = Switch('.//label[@for="rh-cloud-switcher-exclude_installed_packages"]')
     cloud_connector = Pf4Button(locator='//button[text()="Configure Cloud Connector"]')
+    reconfigure_cloud_connector = Pf4Button(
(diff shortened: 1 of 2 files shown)


airgun/views/cloud_inventory.py
```diff
@@ -115,6 +115,9 @@ class CloudInventoryListView(BaseLoggedInView):
     obfuscate_ips = Switch('.//label[@for="rh-cloud-switcher-obfuscate_inventory_ips"]')
     exclude_packages = Switch('.//label[@for="rh-cloud-switcher-exclude_installed_packages"]')
     cloud_connector = Pf4Button(locator='//button[text()="Configure Cloud Connector"]')
+    reconfigure_cloud_connector = Pf4Button(
```
latran, line 118: @lhellebr I'm not sure how to make the `Reconfigure Cloud Connector` button appear in `Inventory Upload` since I only see `Configure Cloud Connector` button on that screen.  Can you show me how?

Finally, why pass in a `locator`?  Why not use `OUIA-Generated-Button` if there's one?

omkarkhatavkar: @latran I think @lhellebr wanted to use more explicit locator which I think okay, I have seen pattern where OUIA id's gets shuffled. Not sure @lhellebr has also experienced that. I still not able to see the `Reconfigure` button, I guess it displayed after we configure it.

omkarkhatavkar (approved):
ACK

jyejare (approved):

PR #673, opened 2022-02-28
new host ui details

tstrych:
This is the base for new UI host details. It can be further improved by anyone. I think this is a good concept. 

This PR is introducing the first 2 Cards in Host details Overview, and also the system to move from one tab to another tab.

PR #674, opened 2022-03-01
70 UI host test fix

tstrych:
this branch is build up on #673 
before you review, please look at  #673 

the old UI host details tests are now using new UI -> dropdown -> legacy UI 
dialog = Pf4ConfirmationDialog()

this is introducing 
Robottelo [PR](https://github.com/SatelliteQE/robottelo/pull/9401)

### DIFF
--- a/airgun/entities/host.py
+++ b/airgun/entities/host.py
@@ -97,11 +97,9 @@ class HostEntity(BaseEntity):
         view = self.navigate_to(self, 'All')
         view.search(entity_name)
         view.table.row(name=entity_name)['Actions'].widget.fill('Delete')
-        alert_message = self.browser.get_alert().text
-        self.browser.handle_alert(cancel=cancel)
+        view.dialog.confirm()
         view.flash.assert_no_error()
         view.flash.dismiss()
-        return alert_message
 
     def delete_interface(self, entity_name, interface_id):
         """Delete host network interface.
@@ -330,6 +328,10 @@ class ShowHostDetails(NavigateStep):
         entity_name = kwargs.get('entity_name')
         self.parent.search(entity_name)
         self.parent.table.row(name=entity_name)['Name'].widget.click()
+        host_view = NewHostDetailsView(self.parent.browser)
+        host_view.wait_displayed()
+        host_view.dropdown.wait_displayed()
+        host_view.dropdown.item_select('Legacy UI')
 
 
 @navigator.register(HostEntity, 'Edit')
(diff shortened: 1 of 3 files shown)


airgun/entities/host.py
```diff
@@ -324,6 +320,10 @@ def step(self, *args, **kwargs):
         entity_name = kwargs.get('entity_name')
         self.parent.search(entity_name)
         self.parent.table.row(name=entity_name)['Name'].widget.click()
+        host_view = NewHostDetailsView(self.parent.browser)
+        host_view.wait_displayed()
+        host_view.dropdown.wait_displayed()
+        host_view.dropdown.item_select('Legacy UI')
```
omkarkhatavkar, line 326: `host_view.dropdown.fill('Legacy UI')` is more robust is not that working for you?

airgun/entities/host.py
```diff
@@ -324,6 +320,10 @@ def step(self, *args, **kwargs):
         entity_name = kwargs.get('entity_name')
         self.parent.search(entity_name)
         self.parent.table.row(name=entity_name)['Name'].widget.click()
+        host_view = NewHostDetailsView(self.parent.browser)
+        host_view.wait_displayed()
+        host_view.dropdown.wait_displayed()
```
omkarkhatavkar, line 325: should be part of the entities actions methods, not sure why they being added part of navigation steps

airgun/entities/host.py
tstrych: @omkarkhatavkar 
I don't understand why you are asking this question?   
PF4 dropdown - `from widgetastic_patternfly4.ouia import Dropdown` inherits from `BaseDropdown` and it has `item_select`. 
Why should I even try to use the fill method, that you are proposing?

airgun/entities/host.py
tstrych: Details navigation is in multiple places, rather than change it everywhere, I change it navigation section.

It's because we navigate via New UI to the old UI, and the entities action methods in `entities/host.py` should navigate user to old UI. 

I understand that it is not used the way it was designed but on the other hand, if it fulfill the purpose, and this  usage is something temporal, why not.

tstrych:
marking as a draft, because #673 should be merged firstly

airgun/entities/host.py
omkarkhatavkar: Fair enough I guess the pattern was fill method usually used the item select options. You are correct it is not presently part of the BaseDropdown.

omkarkhatavkar (approved):
ACK

mshriver (approved):

PR #676, opened 2022-03-04
Update cockpit new UI

peterdragun:
Adding option to use old UI which is used in RHEL7 and new UI that is used in both RHEL8 and 9.

### DIFF
--- a/airgun/entities/host.py
+++ b/airgun/entities/host.py
@@ -243,26 +243,34 @@ class HostEntity(BaseEntity):
         view.flash.assert_no_error()
         view.flash.dismiss()
 
-    def get_webconsole_content(self, entity_name):
+    def get_webconsole_content(self, entity_name, rhel_version=7):
         """Navigate to host's webconsole and return the hostname from the cockpit page
 
         :param str entity_name: The host name for which to set the parameter value.
+        :param int rhel_version: Choose UI elements based on rhel version.
         """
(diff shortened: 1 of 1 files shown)


peterdragun:
trigger: test-robottelo
pytest: tests/foreman/ui/test_host.py -k "test_positive_cockpit[rhel7] or test_positive_cockpit[rhel8]"
airgun: 676

airgun/entities/host.py
```diff
@@ -243,26 +243,34 @@ def set_puppet_class_parameter_value(self, entity_name, name, value):
         view.flash.assert_no_error()
         view.flash.dismiss()
 
-    def get_webconsole_content(self, entity_name):
+    def get_webconsole_content(self, entity_name, use_new_ui=True):
         """Navigate to host's webconsole and return the hostname from the cockpit page
 
         :param str entity_name: The host name for which to set the parameter value.
+        :param bool use_new_ui: When true: use new UI elements for RHEL8 and RHEL9.
```
tstrych, line 250: I would maybe use int instead of bool, non-blocking

tstrych (approved):
ACK

airgun/entities/host.py
peterdragun: Updated to use RHEL version instead of bool. This would be also easier to use from robottelo

pondrejk (approved):

PR #677, opened 2022-03-04
6.9.z backports 

JacobCallahan:
This is a series of cherry-picks (most of which look to have already been included) that should hopefully bring our 6.9.z branch up-to-date with what is needed to run against Selenium Grid 4. Let me know if there is anything else needed.

### DIFF
--- a/airgun/exceptions.py
+++ b/airgun/exceptions.py
@@ -1,5 +1,6 @@
 """Exceptions raised by airgun"""
 from selenium.common.exceptions import InvalidElementStateException
+from widgetastic.exceptions import *  # noqa: F403, F401
 
 
 class ReadOnlyWidgetError(Exception):
--- a/airgun/exceptions.py
+++ b/airgun/exceptions.py
@@ -1,4 +1,5 @@
 """Exceptions raised by airgun"""
+from selenium.common.exceptions import InvalidElementStateException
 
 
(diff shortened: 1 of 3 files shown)


airgun/exceptions.py
```diff
@@ -1,4 +1,6 @@
 """Exceptions raised by airgun"""
+from selenium.common.exceptions import InvalidElementStateException
+from widgetastic.exceptions import *  # noqa: F403, F401
```
jyejare, line 3: AFAIK we have imported the exceptions here from widgetastic but didnt add it into the `__all__` list so that all those exceptions could be accessed from airgun.

I think this also needs to be fixed in airgun master!

mshriver: Airgun modules can continue to import from widgetastic.exceptions directly, as its a direct dependency of the library.  

Can you be more specific about what specifically needs to be fixed here? This commit diff matches master.

mshriver (approved):

JacobCallahan:
@omkarkhatavkar mostly caution, but the diff is pretty small. I'll use this in a PRT run shortly

airgun/exceptions.py
mshriver: Non blocking comment, but still want to hear if we've got some issue here. :)

PR #679, opened 2022-03-13
Add mirroring policy to repository page

latran:
Added mirroring policy to the repository view page as well as added the file repo in repoEdit section.  Rearranged some fields so it matches with what we have in the UI for better comparison.

### DIFF
--- a/airgun/views/repository.py
+++ b/airgun/views/repository.py
@@ -58,8 +58,15 @@ class RepositoryCreateView(BaseLoggedInView):
         verify_ssl = Checkbox(id='verify_ssl_on_sync')
         upstream_username = TextInput(id='upstream_username')
         upstream_password = TextInput(id='upstream_password')
+        auth_token = TextInput(id='ansible_collection_auth_token')
+        mirroring_policy = Select(id='mirroring_policy')
+        include_tags = TextInput(id='include_tags')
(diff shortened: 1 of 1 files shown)


airgun/views/repository.py
```diff
@@ -242,6 +269,15 @@ class OstreeRepository(View):
...
             pass
 
+    @repo_content.register('file')
+    class FileRepository(View):
+        upstream_url = EditableEntry(name='Upstream URL')
+        verify_ssl = EditableEntryCheckbox(name='Verify SSL')
+        upstream_authorization = AuthorizationEntry(name='Upstream Authorization')
+        publish_via_http = EditableEntryCheckbox(name='Publish via HTTP')
+        unprotected = EditableEntryCheckbox(name='Unprotected')
+        mirroring_policy = EditableEntrySelect(name='Mirroring Policy')
```
jyejare, line 279: I see good duplication of the same properties in all types of Repository classes! I believe we should move these all to the common parent/higher class!

The comment is non-blocking and can be a future improvement comment!

jyejare (approved):

omkarkhatavkar (approved):
ACK

omkarkhatavkar:
@latran If you wanted to work on the suggestion from @jyejare create another PR. For now, I'm merging this as this is a non-blocking comment.

PR #680, opened 2022-03-14
Delete confirmation

shweta83:
Added confirmation dialog on organization delete
Removed wait_for after deletion as it can be verified in test itself to confirm the deletion of organization

### DIFF
--- a/airgun/views/organization.py
+++ b/airgun/views/organization.py
@@ -12,6 +12,7 @@ from airgun.widgets import ActionsDropdown
 from airgun.widgets import CustomParameter
 from airgun.widgets import FilteredDropdown
 from airgun.widgets import MultiSelect
+from airgun.widgets import Pf4ConfirmationDialog
 
 
 class OrganizationsView(BaseLoggedInView, SearchableViewMixin):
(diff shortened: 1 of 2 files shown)


omkarkhatavkar (approved):
ACK

airgun/views/organization.py
```diff
@@ -24,6 +25,7 @@ class OrganizationsView(BaseLoggedInView, SearchableViewMixin):
             'Actions': ActionsDropdown("./div[contains(@class, 'btn-group')]"),
         },
     )
+    dialog = Pf4ConfirmationDialog()
```
jyejare, line 28: I think this property (since its so generic and applies to all the components) can be defined at a higher(parent) class level than in a specific component class!

The comment is not blocker for this PR but can be added into future airgun improvements !

jyejare (approved):
ACK with non blocking future improvement comment!

rplevka (approved):

airgun/views/organization.py
shweta83: @jyejare Agree. I can work on it in separate PR.

PR #684, opened 2022-03-23
Fixes of puppet-related entities/views

vsedmik:
Changes:
1) Added PF4 confirmation dialog to ConfigGroups and Environments.
2) From hosts removed PuppetClasses tab (does not exist in 6.11) and added PuppetENC tam (exists on puppet-enabled Satellite when the host is assigned to a puppet capsule).
3) Minor locator updates.


airgun/views/configgroup.py
```diff
@@ -18,6 +19,7 @@ class ConfigGroupsView(BaseLoggedInView, SearchableViewMixin):
             'Actions': Text('.//a[@data-method="delete"]'),
         },
     )
+    dialog = Pf4ConfirmationDialog()
```
shweta83, line 22: @vsedmik Pf4ConfirmationDialog is common in multiple components so we moved it to https://github.com/SatelliteQE/airgun/blob/a2a8be1670196b11ad1c577fa7081c53c10bdde8/airgun/views/common.py#L44

airgun/views/puppet_environment.py
```diff
@@ -28,12 +29,13 @@ class PuppetEnvironmentTableView(BaseLoggedInView, SearchableViewMixin):
...
         column_widgets={
             'Name': Text(
-                ".//a[starts-with(@href, '/environments/') and \
+                ".//a[starts-with(@href, '/foreman_puppet/environments/') and \
                 contains(@href,'/edit')]"
             ),
             'Actions': ActionsDropdown('./div[contains(@class, "btn-group")]'),
         },
     )
+    dialog = Pf4ConfirmationDialog()
```
shweta83, line 38: Pf4ConfirmationDialog is moved to common.py so we can remove this from here

latran:
@vsedmik, I don't know how to setup puppet, but if you have some manual steps, I can look into the failed domain selection issue and also look into your views to see if there's anything missing to give this PR a proper review.

jameerpathan111:
@latran https://github.com/SatelliteQE/robottelo/blob/6b909a2af197b4efbe5e6b57c16a3cbccd3f1b9e/pytest_fixtures/component/puppet.py#L42 this could help.

vsedmik:
@latran yes, you can use the fixure @jameerpathan111 mentioned or I can provide you with already enabled SAT (or with steps to enable).

However, during the night I realized that the problem might not be in the airgun, because another drop-downs in the same dialog work as expected, so the issue might be in the domain availability for the given taxonomy or something similar, so the drop down does not contain it. I will turn the robottelo PR to draft until I get back to it.

vsedmik:
Updated (PF4 dialog removed)

airgun/entities/puppet_environment.py
```diff
@@ -38,7 +38,7 @@ def delete(self, value):
         view = self.navigate_to(self, 'All')
         view.search(value)
         view.table.row(name=value)['Actions'].widget.fill('Delete')
-        self.browser.handle_alert()
```
mshriver, line 41: This really is something that should be getting handled by `browser.handle_alert` - you see it used as a kwarg to `Widget.click` within the table row too.

The PF4 dialog used for the confirmation alert isn't being recognized and handled by `widgetastic.browser.Browser.handle_alert`, selenium doesn't recognize it as an alert.

Since we have a mix of PF4 pages right now, I'm not opposed to this being a short-term solution since its being applied to the pages that are converted.

Long term we need to discuss this with the UX team (@latran, interested?) and make an update to airgun's browser class or wt.core.

mshriver (approved):

PR #689, opened 2022-04-14
Delete Confirmation for UsersRoles

shubhamsg199:
1.  Adding Delete confirmation for Users&Roles


shweta83 (approved):

mshriver:
Like the other PRs, we can merge this in the interest of resolving test failures, but we're actively discussing with the UXD team (thanks @latran) what our 'correct' resolution to these new confirmation dialogs will be.

mshriver (approved):

PR #696, opened 2022-05-12
new host details dropdown locator fix

tstrych:
new dropdown locator

and new host ui details, each card and tab is loaded separately, last network response comes from recent job cards, so I used it to ensure page is loaded

### DIFF
--- a/airgun/views/host_new.py
+++ b/airgun/views/host_new.py
@@ -1,10 +1,10 @@
 from widgetastic.widget import Text
 from widgetastic.widget import View
 from widgetastic.widget import Widget
+from widgetastic_patternfly4 import Dropdown
 from widgetastic_patternfly4 import Tab
 from widgetastic_patternfly4.ouia import BreadCrumb
 from widgetastic_patternfly4.ouia import Button
-from widgetastic_patternfly4.ouia import Dropdown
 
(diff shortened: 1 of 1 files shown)


airgun/views/host_new.py
```diff
@@ -1,10 +1,10 @@
 from widgetastic.widget import Text
 from widgetastic.widget import View
 from widgetastic.widget import Widget
+from widgetastic_patternfly4 import Dropdown
```
mshriver, line 4: Why switch from the OUIA widget? Is it not resolving when visible?

I understand the job card location and element wait, but don't understand the need to switch back to a non-OUIA widget.

tstrych: sure :) because, a few snaps ago the schedule remote job button with dropdown was added, and it has the same  OUIA ID, and it is chosen by the widgetastic, there should be only one ID, and it can be considered as BZ

I was not able to pass any other locator than OUIA ID or OUIA title when using OUIA.

I will rather write down all places where OUIA is needed and ask developers once than create many separate BZ's, where all of them need to be cherry-picked.

mshriver (approved):

pondrejk (approved):

Issue #695, opened 2022-05-10
new host details 


tstrych:
new host details are going to be disabled for 6.11 
and will need to be re-enabled for 6.12

https://github.com/SatelliteQE/airgun/pull/674/files#diff-
95145bb3b31e9dba2c66f3318c9ca9bad16e9b35cdbe132bde7d26acc52a5582R331-R334

PR #698, opened 2022-05-18
disable new host ui details by default

tstrych:
Adressed #695.

in 6.11 the new host details are disabled by default, this PR should be re-added for 6.12

I can manually change it in 6.11 for all host modules so the change without this PR fix would work, but I want to have a 1:1 setup with customers. 

I run two tests from this [PR](https://github.com/SatelliteQE/robottelo/pull/9553 ) one using old and one that is using new host details 

```
tests/foreman/ui/test_host.py::test_positive_read_from_details_page
tests/foreman/ui/test_host.py::test_positive_read_details_page_from_new_ui[host_details_ui]
=========== 2 passed, 40 deselected, 5 warnings in 262.28s (0:04:22) ===========
```

### DIFF
--- a/airgun/entities/host_new.py
+++ b/airgun/entities/host_new.py
@@ -5,6 +5,17 @@ from airgun.views.host_new import NewHostDetailsView
 
 
 class NewHostEntity(HostEntity):
+    def create(self, values):
+        """Create new host entity"""
+        view = self.navigate_to(self, 'New')
+        view.fill(values)
+        self.browser.click(view.submit, ignore_ajax=True)
+        self.browser.plugin.ensure_page_safe(timeout='600s')
+        host_view = NewHostDetailsView(self.browser)
+        host_view.wait_displayed()
+        host_view.flash.assert_no_error()
+        host_view.flash.dismiss()
+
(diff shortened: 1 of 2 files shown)


airgun/entities/host_new.py
```diff
@@ -5,6 +5,17 @@
 
 
 class NewHostEntity(HostEntity):
+    def create(self, values):
+        """Create new host entity"""
+        view = self.navigate_to(self, 'New')
+        view.fill(values)
+        self.browser.click(view.submit, ignore_ajax=True)
+        self.browser.plugin.ensure_page_safe(timeout='600s')
+        host_view = NewHostDetailsView(self.browser)
```
tstrych, line 14: copied from host entity with, but using new View

peterdragun (approved):

swadeley:
Hi @tstrych 

> in 6.11 the new host details are disabled by default, this PR should be re-added for 6.12
> 
If you only want it after 6.11, then best leave it in draft state until after a 6.11 branch as been made.

swadeley (approved):
ACK

Issue #651, opened 2021-12-29
Use Pf4ConfirmationDialog widget to delete airgun entities


jameerpathan111:
Fix existing Airgun code for entities which use new confirmation dialog based on PF4, which is used for deleting entities.

List of UI entities using new confirmation dialog.

- [ ] Hosts
- [ ] scap content
- [ ] Architectures
- [ ] Hardware models
- [ ] Installation Media
- [ ] Operating Systems
- [ ] Job Templates
- [ ] Provisioning Templates
- [ ] Partition Tables
- [ ] Compliance Policies
- [ ] Host Groups
- [ ] Global Parameters
- [ ] Discovery Rules
- [ ] Ansible Roles
- [ ] Ansible Variables
- [ ] Capsules
- [ ] Compute Profiles
- [ ] Compute Resources
- [ ] Virt-who Configurations
- [ ] Http proxy
- [ ] Domains
- [ ] Subnets
- [ ] Realms
- [ ] Locations
- [ ] Organizations
- [ ] Users
- [ ] User Groups
- [ ] Authentication Sources
- [ ] Roles
- [ ] Bookmarks
- [ ] Webhooks
- [ ] Webhook Templates

mshriver:
I've engaged with the UXD team (@Ron-Lavi) and @latran about this PF4 confirmation dialog (which is actually a modal).

This should be addressed not individually on entity views, but within the `get_alert` browser method defined in wt.core.

Either UXD team will modify the modal component to be recognized as a JS alert, or we'll override the core widgetastic browser method so that it can find the the modal when looking for alerts.

mshriver:
Following discussion with UX, they're not going to be able to 'spoof' the browser alert for the new confirmation dialog modals.

This means we need to address this by implementing a custom `Brower.get_alert` function that can identify not only JS alerts, but confirmation modals.

https://github.com/SatelliteQE/airgun/blob/12c4698ecbfab2faa461594cc53023f21b5670f8/airgun/browser.py#L258
https://github.com/RedHatQE/widgetastic.core/blob/7b228874bdbdbe7dc29436cff718bf8ce34b7e30/src/widgetastic/browser.py#L919


`AirgunBrowser.get_alert` should be implemented, in order to override the functionality of `Browser.get_alert` in the base class.

`get_alert` is the basis for several other functions within the browser, and the overridden method should be compatible with them, returning an element that `dismiss`, `send_keys`, `accept`, and `text` can be called against.

[...]

PR #699, opened 2022-05-23
Added custom behaviour get_alert and handle_alert

omkarkhatavkar:
Resolving  #651

### DIFF
--- a/airgun/browser.py
+++ b/airgun/browser.py
@@ -16,9 +16,12 @@ from wait_for import wait_for
 from webdriver_kaifuku import BrowserManager
 from widgetastic.browser import Browser
 from widgetastic.browser import DefaultPlugin
+from widgetastic.exceptions import NoAlertPresentException
+from widgetastic.exceptions import NoSuchElementException
 
 from airgun import settings
-
+from airgun.widgets import ConfirmationDialog
+from airgun.widgets import Pf4ConfirmationDialog
 
 LOGGER = logging.getLogger(__name__)
 
@@ -409,3 +412,49 @@ class AirgunBrowser(Browser):
         self.url = current_url
         self.plugin.ensure_page_safe()
         return file_path
+
+    def check_alert(self, locator):
+        try:
+            return self.move_to_element(locator).is_displayed()
+        except NoSuchElementException:
+            self.logger.info(f'No {locator} type alert detected !')
+        return False
+
+    def get_alert(self):
+        """Returns the current alert/PF4 alert object.
+
+        Raises:
+            :py:class:`selenium.common.exceptions.NoAlertPresentException`
+        """
+        pf4_modal_locator = "//div[@data-ouia-component-type='PF4/ModalContent']"
+        modal_locator = "//div[@class='modal-content']"
+        modal_map = {pf4_modal_locator: Pf4ConfirmationDialog, modal_locator: ConfirmationDialog}
+        if not self.handles_alerts:
+            return None
+        try:
+            return self.selenium.switch_to.alert
+        except NoAlertPresentException:
+            for locator, locator_class in modal_map.items():
+                if self.check_alert(locator):
+                    return locator_class(self.browser)
+            raise
+
+    def handle_alert(
+        self,
+        cancel=False,
+        wait=30.0,
+        squash=False,
+        prompt=None,
+        check_present=False,
+    ):
+        """Extend the behaviour of widgetstatic.browser.handle_alert to handle PF4 alerts"""
+        popup = self.get_alert()
+        if isinstance(popup, (Pf4ConfirmationDialog, ConfirmationDialog)):
+            if cancel:
+                self.logger.info("  dismissing")
+                popup.cancel()
+            else:
+                self.logger.info("  accepting")
+                popup.confirm()
+        else:
+            super(AirgunBrowser, self.browser).handle_alert()
--- a/airgun/entities/location.py
+++ b/airgun/entities/location.py
@@ -26,7 +26,7 @@ class LocationEntity(BaseEntity):
         view = self.navigate_to(self, 'All')
         view.search(entity_name)
         view.table.row(name=entity_name)['Actions'].widget.fill('Delete')
-        view.dialog.confirm()
+        self.browser.handle_alert()
         view.flash.assert_no_error()
         view.flash.dismiss()
 
--- a/airgun/entities/user.py
+++ b/airgun/entities/user.py
@@ -49,7 +49,7 @@ class UserEntity(BaseEntity):
         view = self.navigate_to(self, 'All')
         view.search(entity_name)
         view.table.row(username=entity_name)['Actions'].widget.click(handle_alert=True)
-        view.dialog.confirm()
+        self.browser.handle_alert()
         view.flash.assert_no_error()
         view.flash.dismiss()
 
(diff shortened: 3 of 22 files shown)


airgun/browser.py
```diff
@@ -409,3 +410,45 @@ def save_downloaded_file(self, file_uri=None, save_path=None):
...
+
+    def handle_alert(
+        self,
+        cancel=False,
+        wait=30.0,
+        squash=False,
+        prompt=None,
+        check_present=False,
+    ):
+        """Extented the behaviour of widgetstatic.browser.handle_alert for handle PF4 alerts"""
```
swadeley, line 438:
```suggestion
        """Extend the behaviour of widgetstatic.browser.handle_alert to handle PF4 alerts"""
```

swadeley (approved):
ACK, with one typo.

shweta83 (approved):

airgun/browser.py
```diff
@@ -409,3 +410,45 @@ def save_downloaded_file(self, file_uri=None, save_path=None):
...
+        popup = self.get_alert()
+        if isinstance(popup, Pf4ConfirmationDialog):
+            if cancel:
+                self.logger.info("  dismissing")
+                popup.cancel()
+            else:
+                self.logger.info("  accepting")
+                popup.confirm()
+        else:
+            self.handle_alert(
```
mshriver, line 448: Are you intending to call an ancestor's `handle_alert`, or make this recursive? I think its the first, in which case I'd call it with `super()` instead

omkarkhatavkar: I guess I miss the trick here need to use the `super()`

mshriver: https://docs.python.org/3/library/functions.html#super

So `super()` will resolve the parent class instance, and you can be sure which class's `handle_alert` method is called.

omkarkhatavkar:
Updated Behaviour To Handle Conditions:

- Basic the browser javascript alert   
- Handle Pf3 Alert/Confirmation Modal 
- Handle Pf4 Alert/Confirmation Modal

airgun/browser.py
```diff
@@ -409,3 +412,49 @@ def save_downloaded_file(self, file_uri=None, save_path=None):
...
+        popup = self.get_alert()
+        if isinstance(popup, (Pf4ConfirmationDialog, ConfirmationDialog)):
+            if cancel:
+                self.logger.info("  dismissing")
+                popup.cancel()
+            else:
+                self.logger.info("  accepting")
+                popup.confirm()
+        else:
+            super(AirgunBrowser, self.browser).handle_alert()
```
mshriver, line 460: You can use it with no arguments here, but it's not hurting to call it with args.

airgun/browser.py
```diff
@@ -409,3 +412,49 @@ def save_downloaded_file(self, file_uri=None, save_path=None):
...
+            :py:class:`selenium.common.exceptions.NoAlertPresentException`
+        """
+        pf4_modal_locator = "//div[@data-ouia-component-type='PF4/ModalContent']"
+        modal_locator = "//div[@class='modal-content']"
+        modal_map = {pf4_modal_locator: Pf4ConfirmationDialog, modal_locator: ConfirmationDialog}
+        if not self.handles_alerts:
+            return None
+        try:
+            return self.selenium.switch_to.alert
+        except NoAlertPresentException:
```
mshriver, line 436: Great pattern here instead of trying is_displayed first.

airgun/browser.py
```diff
@@ -409,3 +412,49 @@ def save_downloaded_file(self, file_uri=None, save_path=None):
         self.url = current_url
         self.plugin.ensure_page_safe()
         return file_path
+
+    def check_alert(self, locator):
+        try:
```
mshriver, line 417: Consider moving the modal map with locators into this function, so it handles checking the potential non-browser alert element types internally, returning the relevant widget instance or raising.

Then test authors have the ability to directly validate that the alert is using a non-alert modal by calling this function to get the alert widget.

airgun/entities/user.py
```diff
@@ -49,7 +49,7 @@ def delete(self, entity_name):
         view = self.navigate_to(self, 'All')
         view.search(entity_name)
         view.table.row(username=entity_name)['Actions'].widget.click(handle_alert=True)
-        view.dialog.confirm()
+        self.browser.handle_alert()
```
mshriver, line 52: `handle_alert=True` being passed to the `click`  call above means this isn't necessary anymore.

airgun/entities/usergroup.py
```diff
@@ -44,7 +44,7 @@ def delete(self, entity_name):
         view = self.navigate_to(self, 'All')
         view.search(entity_name)
         view.table.row(name=entity_name)['Actions'].widget.click(handle_alert=True)
-        view.dialog.confirm()
+        self.browser.handle_alert()
```
mshriver, line 47: Same as above, handle_alert=True is already passed.

airgun/widgets.py
```diff
@@ -1067,9 +1067,9 @@ class Pf4ConfirmationDialog(ConfirmationDialog):
     right corner."""
 
     ROOT = '//div[@id="app-confirm-modal" or @data-ouia-component-type="PF4/ModalContent"]'
-    confirm_dialog = PF4Button('OUIA-Generated-Button-danger-1')
-    cancel_dialog = PF4Button('OUIA-Generated-Button-link-1')
-    discard_dialog = PF4Button('OUIA-Generated-Button-plain-1')
+    confirm_dialog = PF4Button(locator='//button[text()="Confirm"]')
```
mshriver, line 1070: Why remove the ouia locators?

airgun/browser.py
mshriver: (non-blocking)

mshriver (changes requested):
Couple small things to clean up or discuss, looking very strong though, thanks @omkarkhatavkar

airgun/entities/location.py
```diff
@@ -25,8 +25,7 @@ def delete(self, entity_name):
         """Delete existing location"""
         view = self.navigate_to(self, 'All')
         view.search(entity_name)
-        view.table.row(name=entity_name)['Actions'].widget.fill('Delete')
```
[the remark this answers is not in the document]
omkarkhatavkar: that was removed while I was testing. need to be present

airgun/widgets.py
omkarkhatavkar: These locators are not stable and change when you repeat the step. e.g. when you try to delete the entity 2 times the locator id changes. I thought this would be more brittle than that.

airgun/browser.py
omkarkhatavkar: there is not displayed method for checking alerts, these only exists in the web element

airgun/browser.py
omkarkhatavkar: The reason was designed like that, we basically do not know which type of confirmation modal it will be on the webpage. I do want to continue the loop even after the exception. As per my thinking, `get_alert` will anytime return the right alert object either javascript alert of pf3/4 alert

airgun/browser.py
omkarkhatavkar: I tried my head scratched for more time, but the `super().handle_alert()` did not work for me it throws about not being able to find the handle_alert() method

airgun/widgets.py
mshriver: I think you can address this with the UXD team to get static ids

omkarkhatavkar: Just checked with the UX team https://github.com/theforeman/foreman/commit/4bcd67ddb14abad4c9b44e213e19e1dfda7c8b7e here is the updated PR land in 6.12

swadeley (approved):
ACK

PR #710, opened 2022-06-06
Replace text() function

shweta83:
Replace `text()` function with `normalize-space(.)` to standardize use in xpath locators.
 Issue  https://github.com/SatelliteQE/airgun/issues/591

### DIFF
--- a/airgun/views/discoveredhosts.py
+++ b/airgun/views/discoveredhosts.py
@@ -147,14 +147,17 @@ class DiscoveredHostsActionDialog(BaseLoggedInView):
 class DiscoveredHostsAutoProvisionDialog(DiscoveredHostsActionDialog):
     """Discovered hosts Auto Provision action dialog view"""
 
-    title = Text("//h4[text()='Auto Provision - The following hosts are about to be changed']")
+    title = Text(
+        "//h4[normalize-space(.)='Auto Provision - The following hosts are about to be changed']"
+    )
 
 
 class DiscoveredHostsAssignOrganizationDialog(DiscoveredHostsActionDialog):
     """Discovered hosts Assign Organization action dialog view"""
 
     title = Text(
-        "//h4[text()='Assign Organization - The following hosts are about to be changed']"
+        "//h4[normalize-space(.)='Assign Organization - "
+        "The following hosts are about to be changed']"
     )
     organization = Select(id='organization_id')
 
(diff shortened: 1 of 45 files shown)


mshriver:
@shweta83 how much of the robottelo UI test folder have you executed against this? We can set up a Jenkins job to make it easier since this could be very slow for you to run locally.

The element selection of the XPATH expression becomes a bit more important when using normalize-space(.) as it will gather and concatenate all text from child nodes of the contextual element being selected. `text()` will collect only the text from the selected element.

So its possible the use of `normalize-space(.)` will grab more text than was being collected before, causing test behavior/validation to fail.  These can be really hard to spot on visual inspection/review.  Easy to fix once we find them by adjusting the XPATH to select a more specific element or by using `normalize-space(text())`

airgun/views/discoveredhosts.py
```diff
@@ -147,42 +147,47 @@ def is_displayed(self):
...
+    )
 
 
 class DiscoveredHostsAssignOrganizationDialog(DiscoveredHostsActionDialog):
     """Discovered hosts Assign Organization action dialog view"""
 
     title = Text(
-        "//h4[text()='Assign Organization - The following hosts are about to be changed']"
+        "//h4[normalize-space(.)='Assign Organization -"
+        "The following hosts are about to be changed']"
```
mshriver, line 160: Looks like this is missing a space between `-` and `The`.

airgun/views/hardware_model.py
```diff
@@ -10,13 +10,13 @@
 
 
 class DeleteHardwareModelDialog(ConfirmationDialog):
-    confirm_dialog = Text(".//button[text()='Delete']")
-    cancel_dialog = Text(".//button[text()='Cancel']")
+    confirm_dialog = Text(".//button[normalize-space(.)='Delete']")
+    cancel_dialog = Text(".//button[normalize-space(.)/='Cancel']")
```
mshriver, line 14: Extra slash here?

airgun/views/host.py
```diff
@@ -754,42 +759,49 @@ def is_displayed(self):
 
 class HostsAssignOrganization(HostsActionCommonDialog):
     title = Text(
-        "//h4[text()='Assign Organization - The following hosts are about to be changed']"
+        "//h4[normalize-space(.)='Assign Organization -"
+        "The following hosts are about to be changed']"
```
mshriver, line 763: Missing space

airgun/views/hardware_model.py
shweta83: Added by mistake. Removed the slash

shweta83:
> @shweta83 how much of the robottelo UI test folder have you executed against this? We can set up a Jenkins job to make it easier since this could be very slow for you to run locally.
> 
> The element selection of the XPATH expression becomes a bit more important when using normalize-space(.) as it will gather and concatenate all text from child nodes of the contextual element being selected. `text()` will collect only the text from the selected element.
> 
> So its possible the use of `normalize-space(.)` will grab more text than was being collected before, causing test behavior/validation to fail. These can be really hard to spot on visual inspection/review. Easy to fix once we find them by adjusting the XPATH to select a more specific element or by using `normalize-space(text())`

@mshriver I totally agree. Though I tested the changes for most of the files but it would be easier if we have Jenkins job to test the changes so that we can fix it if any validation failure occurs.

mshriver (approved):

PR #711, opened 2022-06-10
fixed the bookmark search in airgun

omkarkhatavkar:
This will fix the `session.host.create_bookmark({'name': name, 'query': query, 'public': True})`, As there are mixture of the PF4 search and PF3 Search. We were failing this part. This will fix both conditions. 

@jaryn Can you please test with this PR? it worked for me over Satellite 6.11 and 6.10


jaryn:
The tests went much further with your patch! Thanks!

swadeley (approved):
ACK

jyejare (approved):

mshriver (approved):

PR #712, opened 2022-06-13
Fixing DiscoveryPlugin failures

Griffin-Sullivan:
In the new UI, the Discovery Rules page has a different Create button depending on if there are any rules yet or not. My updates made a simple change to address this. I also added a `wait_for` during an Auto Provision dialog step in Discovered Hosts, since one of my tests was failing due to the dialog not submitting properly.

### DIFF
--- a/airgun/entities/discoveredhosts.py
+++ b/airgun/entities/discoveredhosts.py
@@ -118,6 +118,13 @@ class DiscoveredHostsEntity(BaseEntity):
             self, 'Select Action', action_name=action_name, entities_list=entities_list
         )
         view.fill(values)
+        wait_for(
+            lambda: view.submit.disabled,
+            fail_condition=True,
+            timeout=10,
+            delay=0.2,
+            logger=view.logger,
+        )
         view.submit.click()
         view.flash.assert_no_error()
         view.flash.dismiss()
--- a/airgun/entities/discoveryrule.py
+++ b/airgun/entities/discoveryrule.py
@@ -1,4 +1,5 @@
 from navmazing import NavigateToSibling
+from widgetastic.exceptions import NoSuchElementException
 
 from airgun.entities.base import BaseEntity
 from airgun.navigation import NavigateStep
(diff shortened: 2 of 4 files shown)


airgun/entities/discoveredhosts.py
```diff
@@ -119,6 +119,14 @@ def apply_action(self, action_name, entities_list, values=None):
         )
         view.fill(values)
         view.submit.click()
+        wait_for(
```
mshriver, line 122: The create button behavior fix makes sense - and explains why it was failing in CI differently than locally.

This wait_for I'm not as sure about as a mechanism to retry clicking submit.

1. Why does the first submit.click() fail silently? Previously I've seen this happen with a disabled button, as click can run without exception/warning logged when the element is visible.  If this is happening, it would be better to address it by waiting for `view.submit.enabled` before calling `view.submit.click()` (without retry).

2. If there is some other explanation (or none found) as to why the first submit is not working, this wait_for needs parameters adjusted. 
    - Ten seconds is way too long to wait between retry. 
    - I would also consider some adjustments to the pattern.  For clarity of the wait_for behavior I try to stick to patterns like a non-inverse boolean func, `lambda: view.submit.is_displayed`, with `fail_condition=True` being used to indicate what return causes wait_for to continue. This way the function's signature "reads" more like: When the submit button is displayed, fail the check and run fail_func.
    - The `fail_func` kwarg needs to pass the function method, in your case you're actually calling the function and passing its result as `fail_func`. There's a sort of protection against calling `None()` in wait_for, so you don't see a TypeError.

mshriver (changes requested):
View/navigation updates look solid, discoveredhosts form behavior needs some discussion.

Great work debugging through the browser behavior!

airgun/entities/discoveredhosts.py
mshriver: Glad to hear/see the submit button check works here and resolves the behavior, that's a far better approach to address the form behavior.

PR #715, opened 2022-06-20
Close action dropdown after reading

### LINKED ISSUE #714: Actions dropdown clicking on wrong elements
After merging the #711 all Action dropdowns are opened to read all options, but on dropdowns in a table with multiple rows this is causing the dropdown from the first line to cover the dropdown from the second row. The dropdown from the first row is never closed, so instead of opening the dropdown from the second row, we are clicking on the option from the first dropdown. For E.g. see the Hosts view :
image

This caused multiple tests fail in Hosts component

peterdragun:
fixes #714

### DIFF
--- a/airgun/widgets.py
+++ b/airgun/widgets.py
@@ -591,13 +591,23 @@ class ActionsDropdown(GenericLocatorWidget):
             except NoSuchElementException:
                 self.pf4_drop_down.click()
 
+    def close(self):
+        """Closes dropdown list"""
+        if self.is_open:
+            try:
+                self.dropdown.click()
+            except NoSuchElementException:
+                self.pf4_drop_down.click()
+
     @property
     def items(self):
         """Returns a list of all dropdown items as strings."""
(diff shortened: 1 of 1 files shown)


airgun/widgets.py
```diff
@@ -595,9 +595,11 @@ def open(self):
     def items(self):
         """Returns a list of all dropdown items as strings."""
         self.open()
-        return [
+        items = [
             self.browser.text(el) for el in self.browser.elements(self.ITEMS_LOCATOR, parent=self)
         ]
+        self.dropdown.click()  # close the dropdown
```
omkarkhatavkar, line 601: I think it is better to write the `self.close` here similar to what we have open, and migrate the dropdown click to handle that. 
One another thing, the dropdown could be pf4 as well. make modifications that it handles both.

omkarkhatavkar (approved):

jyejare (approved):

PR #716, opened 2022-06-20
fix the locator issue in action drop down

omkarkhatavkar:
- Remove the redundant `self.open` as `self.items` has that step
- Fixed the correct locator here (https://github.com/SatelliteQE/airgun/pull/711/files#diff-b6d36691d2c9b246c1a2f7fff90473fe0098b4a6846694417e9acc246e35ea30R569) because of this line commit  

@vsedmik Can you please test your changes as well as 1 bookmark test passes with this change.  (from my side it looks okay)

### DIFF
--- a/airgun/widgets.py
+++ b/airgun/widgets.py
@@ -566,8 +566,8 @@ class ActionsDropdown(GenericLocatorWidget):
     )
     pf4_drop_down = Dropdown('OUIA-Generated-Dropdown-2')
     button = Text(
-        "//*[self::button or self::span][contains(@class, 'btn')]"
-        "[contains(@aria-label, 'search button')]"
+        ".//*[self::button or self::span][contains(@class, 'btn') or "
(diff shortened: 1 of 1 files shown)


vsedmik (approved):
Tested with tests/foreman/ui/test_smartclassparameter.py::test_positive_update_matcher_from_attribute and it passed, so seems good to me.

airgun/widgets.py
```diff
@@ -602,7 +602,6 @@ def items(self):
     def select(self, item):
         """Selects item from dropdown."""
         if item in self.items:
-            self.open()
```
peterdragun, line 605: Please leave this open as this will be necessary after merging #715

peterdragun (changes requested):

peterdragun (approved):

omkarkhatavkar:
merging this as multiple ack's

PR #717, opened 2022-06-22
New Host UI Ansible Work

sambible:
Figured I'd put this up earlier rather than later, and just leave it in Draft as I work on this. 

This is my first Airgun PR, and I'm going to be flailing for a while, definitely appreciate eyes on this early and often - with any suggestions you think make sense!


airgun/views/host_new.py
```diff
@@ -93,7 +93,23 @@ class RepositorySets(Tab):
 
     @View.nested
     class Ansible(Tab):
-        pass
+        ROOT = './/div[contains(@class, "ansible-host-details")]'
+
+        @View.nested
+        class RolesTab(Tab):
+            ROOT = ""
```
mshriver, line 100: No need to set this, the `Tab` class implements a locator based on the name of the class

mshriver: https://github.com/RedHatQE/widgetastic.patternfly4/blob/master/src/widgetastic_patternfly4/tabs.py#L6

So, the 'right' way to do it is to set `TAB_NAME` within the tab class, or to name the tab class according to the displayed tab in the UI (comment on TAB_NAME).

No need to set ROOT then, its defined in the `Tab` class and is based on `TAB_NAME`.

The other tabs in this module are using ROOT directly, which isn't ideal, but maybe is necessary if the DOM is messed up? I'd try with TAB_NAME first, and if that works try to apply it to all `Tab` based classes in the view.

latran:
This will be a challenge since the tile cards are something new, but you're lucky that `Card` is one of the supported components of PF4.  I would highly advise writing some RFE BZs to get the ouia-id into those cards for automation if you're planning on reading each card.

https://www.patternfly.org/v4/developer-resources/open-ui-automation

sambible:
Accidentally closed this some how, but reopening it/adding a new commit

airgun/views/host_new.py
```diff
@@ -264,12 +266,84 @@ class traces(Tab):
...
+                },
+            )
+
+        @View.nested
+        class inventory(SatTab):
+            TAB_NAME = 'Inventory'
+            ROOT = ""
+
+        @View.nested
+        class jobs(SatTab):
```
synkd, line 318: In chat, @mayurilahane pointed out that representing these three states in a single view class could cause issues with `is_displayed`. Are you planning to keep this class as is or split it into separate classes?

airgun/views/host_new.py
```diff
@@ -294,7 +368,19 @@ class InstallPackagesView(View):
     cancel = Button('Cancel')
 
 
+class EditAnsibleRolesView(View):
+    """Edit Ansible Roles Modal"""
+
+    ROOT = ''
+    #No current representation for this Widget in Widgetastic
```
synkd, line 375: Are you planning to write this widget in upstream Widgetastic?

synkd (commented):
Glad to see you making progress on this! Have you written any tests that use this view and entity as a POC?

airgun/views/host_new.py
sambible: I feel like I will have to, if this is going to be automated in the same was as everything else. Perhaps there's an option to reach out to dev and get some assistance.

airgun/views/host_new.py
```diff
@@ -264,12 +266,84 @@ class traces(Tab):
...
+        ROOT = './/div[contains(@class, "ansible-host-details")]'
+        TAB_NAME = 'Ansible'
+
+        @View.nested
+        class roles(SatTab):
+            TAB_NAME = 'Roles'
+            assignedRoles = Text('.//a[contains(@href, "roles/all")]')
+            edit = Button(locator='.//button[@aria-label="edit ansible roles"]')
+            table = PatternflyTable(
+                #No OUIA ID
```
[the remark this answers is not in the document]
sambible: Those are auto generated OUIA IDs, and change when you load more tables :/ 

Can't be used for reliable automation. https://bugzilla.redhat.com/show_bug.cgi?id=2060099 similar case to this here.

sambible:
Tests are coming today hopefully, in a seperate PR - saw something that @latran was doing with his PR that looked neat with having it reference my Airgun PR as the source of airgun in the testing, so I'll try that

airgun/views/host_new.py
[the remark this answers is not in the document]
sambible: Maybe I don't need specifically an ID for the exact table, since there's only on on the page? Not sure.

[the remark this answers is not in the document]
synkd: That's what I was thinking. The same ID is used for all three tables, but only one of them should be present on the page at a given time. I would give it a shot and consider falling back to using `class` as a locator if you hit issues using `data-ouia-component-id`.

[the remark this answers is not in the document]
mayurilahane: Hey @synkd can you share the link to UI?

airgun/views/host_new.py
```diff
@@ -264,12 +266,84 @@ class traces(Tab):
...
 
     @View.nested
-    class ansible(Tab):
+    class insights(Tab):
         pass
 
     @View.nested
-    class insights(Tab):
-        pass
+    class ansible(SatTab):
```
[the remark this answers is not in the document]
mayurilahane: As i have seen before, initials should be capitalized.

airgun/views/host_new.py
mayurilahane: also, you will need to add "page refresh" as a prerequisite to the second view 
ex: https://gitlab.cee.redhat.com/insights-qe/iqe-ros-plugin/-/blob/master/iqe_ros/entities/__init__.py#L147

mayurilahane: also, consider page refresh time

airgun/views/host_new.py
```diff
@@ -264,12 +266,84 @@ class traces(Tab):
...
+                    'Description': Text('.//a'),
+                    'Schedule': Text('./span'),
+                    'Next Run': Text('./span'),
+                    4: Dropdown(locator='.//div[contains(@class, "pf-c-dropdown")]'),
+                }
+            )
+
+            #Only displayed on Refresh when there are previously executed jobs
+            previousJobsTable = PatternflyTable(
+                #No OUIA ID
```
mayurilahane, line 337: it has ouaid "table-row1/2/3"

airgun/views/host_new.py
```diff
@@ -264,12 +266,84 @@ class traces(Tab):
...
+
+            #Mutually Exclusive with the above button
+            scheduledJobsTable = PatternflyTable(
+                #No OUIA ID
+                locator='',
+                column_widgets={
+                    'Description': Text('.//a'),
+                    'Schedule': Text('./span'),
+                    'Next Run': Text('./span'),
+                    4: Dropdown(locator='.//div[contains(@class, "pf-c-dropdown")]'),
```
mayurilahane, line 331: this locator is not locating the dropdown for "cancel"

airgun/views/host_new.py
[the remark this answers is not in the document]
peterdragun: It was suggested by @mshriver to use lowercase names for classes so the navigation will look like `host_new.ansible.variables`. You can see this pattern through the airgun, if you are creating a new widget - the class name should be capitalized; but here as a nested view, it should be lowercase.

sambible:
Big time gap here, but this commit will change several of the classes, rework Jobs to function properly with 3 different nested views, and also finally get the subtab navigation working and allowing the first of the tests to pass.

sambible:
https://github.com/SatelliteQE/robottelo/pull/10155 here's a link to the Test PR, which has also been updated. Going to do a cross branch test trigger here.

airgun/views/host_new.py
```diff
@@ -294,7 +396,32 @@ class InstallPackagesView(View):
...
 class ModuleStreamDialog(Pf4ConfirmationDialog):
 
     confirm_dialog = Button(locator='.//button[@aria-label="confirm-module-action"]')
     cancel_dialog = Button(locator='.//button[@aria-label="cancel-module-action"]')
+
+
+class RecurringJobDialog(Pf4ConfirmationDialog):
+
+    confirm_dialog = Button(locator='.//button[@data-ouia-component-id="btn-modal-confirm"]')
+    cancel_dialog = Button(locator='.//button[@data-ouia-component-id="btn-modal-cancel"]')
```
peterdragun, line 427:
```suggestion
    confirm_dialog = OUIAButton("btn-modal-confirm")
    cancel_dialog = OUIAButton("btn-modal-cancel")
```

airgun/views/host_new.py
```diff
@@ -264,12 +264,114 @@ class traces(Tab):
...
+
+            @property
+            def is_displayed(self):
+                return (
+                    self.schedule.is_displayed
+                    or self.jobs.is_displayed
+                    or self.previous.is_displayed
+                )
+
+            class schedule(Tab):
```
peterdragun, line 329: I think you are missing the `@View.nested` decorator on these classes :thinking:

sambible: Seems like it, yep

sambible:
Would like to get this merged soon, want to get everything in since Ansible is no longer on me.

peterdragun (changes requested):
If you want to get this merged soon, then you should add my suggestion, that's definitely a blocker for me.

peterdragun (approved):

omkarkhatavkar:
@sambible Do you have passing tests for this?

sambible:
https://github.com/SatelliteQE/robottelo/pull/10155 Test for this Airgun work, nailgun PR has been merged so all should be good to go.

sambible:
Test in Sister PR is passing, lgtm

airgun/views/host_new.py
```diff
@@ -254,12 +254,117 @@ class traces(Tab):
...
+                    'Ansible role': Text('./span'),
+                    'Type': Text('./span'),
+                    # the next field can also be a form group
+                    'Value': Text('./span'),
+                    'Source attribute': Text('./span'),
+                    # The next 2 buttons are hidden by default, but appear in this order
+                    5: Button(locator='.//button[@aria-label="Cancel editing override button"]'),
+                    6: Button(locator='.//button[@aria-label="Submit override button"]'),
+                    # Clicking this button hides it, and displays the previous 2
+                    7: Button(locator='.//button[@aria-label="Edit override button"]'),
```
Griffin-Sullivan, line 296: Should we have dev make this easier for automation like a better id to select by instead of titles (since those can be changed)? We can address this in another PR if you think we need to. In terms of automation coverage I imagine you wouldn't really be using these buttons much anyway?

Griffin-Sullivan (approved):
Good stuff. Let's get another review and merge, so we can take on the robottelo PR.

airgun/views/host_new.py
sambible: It's always better for dev to have specific IDs for things, since they make it much more likely to not fall apart at small changes in the future. As far as automation coverage is concerned, no, you wouldn't.

synkd (approved):
Nice job getting the test passing with this PR!

omkarkhatavkar (approved):

Gauravtalreja1:
@sambible ah you missed `6.13.z` label, Can you cherrypick this to 6.13.z manually?

PR #718, opened 2022-06-22
Fixing AutoCompleteTextInput blur

Griffin-Sullivan:
The `foreman_autocomplete` component has different behavior in 6.11 where once you finish typing the search, the dropdown stays open even if you run `<element>.blur()`. Replacing this with a tab makes sure the element is out of focus. Tested on 6.10 and 6.11 and it seems to work. Waiting for https://github.com/SatelliteQE/airgun/pull/716 to be merged and I will test this change on the DiscoveryPlugin pipeline.

### DIFF
--- a/airgun/widgets.py
+++ b/airgun/widgets.py
@@ -2138,7 +2138,7 @@ class AutoCompleteTextInput(TextInput):
         self.clear()
         super().fill(value)
         self.browser.plugin.ensure_page_safe()
-        self.browser.execute_script('arguments[0].blur();', self.__element__())
+        self.browser.send_keys_to_focused_element('\t')
         return self.value != old_value
 
 


airgun/widgets.py
```diff
@@ -2128,7 +2128,7 @@ def fill(self, value):
         self.clear()
         super().fill(value)
         self.browser.plugin.ensure_page_safe()
-        self.browser.execute_script('arguments[0].blur();', self.__element__())
+        self.browser.send_keys_to_focused_element('\t')
```
omkarkhatavkar, line 2131: is this with all `search box, we might have some search which still uses the old one? Do you think we could also add the backword compatibility here?

mshriver: I think this fill is kind of weirdly implemented in the first place.

Putting the `clear()` call in `before_fill` and this call in `after_fill` would be a better way to implement it - not even implementing `AutoCompleteInputText.fill` at all. 

Wt.core's fill method will automatically call `before_fill` and `after_fill` when `fill` is called on this widget.

Also thinking asserting against the field changing should not be within this widget's fill behavior. That's starting to try and test the component implementation every time we use it.

Griffin-Sullivan: ok I can change it around. I also just found an issue with Discovery Rule read_all so going to include that here. I agree the assertion didn't really make sense here.

mshriver (approved):

PR #719, opened 2022-06-28
Remove unnecessary handle_alert

shubhamsg199:
Removing unnecessary handle_alert which  is causing User end_to_end test to fail


omkarkhatavkar (approved):
looks like I missed this one, thanks @shubhamsg199

PR #720, opened 2022-07-01
Add content tab to new host UI

peterdragun:
Also part of this PR is replacing Cards locators that were based on their position in UI with locators based on names. As those elements are dynamically generated we cannot rely on their position.

All locators were tested and seem to work without any issue, except those ones that are disabled/not presented in 6.11.

### DIFF
--- a/airgun/entities/host_new.py
+++ b/airgun/entities/host_new.py
@@ -1,5 +1,3 @@
-import time
-
 from airgun.entities.host import HostEntity
 from airgun.navigation import NavigateStep
 from airgun.navigation import navigator
@@ -30,24 +28,24 @@ class NewHostEntity(HostEntity):
     def get_packages(self, entity_name, search=""):
         """Filter installed packages on host"""
         view = self.navigate_to(self, 'NewDetails', entity_name=entity_name)
-        view.Content.Packages.select()
-        view.Content.Packages.searchbar.fill(search)
+        view.content.packages.select()
+        view.content.packages.searchbar.fill(search)
         # wait for filter to apply
-        time.sleep(1)
-        return view.Content.Packages.read()
+        self.browser.plugin.ensure_page_safe()
+        return view.content.packages.read()
 
     def install_package(self, entity_name, package):
         """Installs package on host using the installation modal"""
         view = self.navigate_to(self, 'NewDetails', entity_name=entity_name)
         view.wait_displayed()
-        view.Content.Packages.select()
-        view.Content.Packages.dropdown.wait_displayed()
-        view.Content.Packages.dropdown.item_select('Install packages')
+        view.content.packages.select()
+        view.content.packages.dropdown.wait_displayed()
+        view.content.packages.dropdown.item_select('Install packages')
         view = InstallPackagesView(self.browser)
         view.wait_displayed()
         view.searchbar.fill(package)
         # wait for filter to apply
-        time.sleep(2)
+        self.browser.plugin.ensure_page_safe()
         view.select_all.click()
         view.install.click()
 
@@ -55,11 +53,11 @@ class NewHostEntity(HostEntity):
         """Apply `action` to selected package based on the `package_name`"""
         view = self.navigate_to(self, 'NewDetails', entity_name=entity_name)
         view.wait_displayed()
-        view.Content.Packages.searchbar.fill(package_name)
+        view.content.packages.searchbar.fill(package_name)
         # wait for filter to apply
-        time.sleep(1)
-        view.Content.Packages.table.wait_displayed()
-        view.Content.Packages.table[0][5].widget.item_select(action)
+        self.browser.plugin.ensure_page_safe()
+        view.content.packages.table.wait_displayed()
+        view.content.packages.table[0][5].widget.item_select(action)
         view.flash.assert_no_error()
         view.flash.dismiss()
 
@@ -67,17 +65,19 @@ class NewHostEntity(HostEntity):
         """List errata based on type and return table"""
         view = self.navigate_to(self, 'NewDetails', entity_name=entity_name)
         view.wait_displayed()
-        view.Content.Errata.select()
-        view.Content.Errata.type_filter.fill(type)
-        return view.read(widget_names="Content.Errata.table")
+        view.content.errata.select()
+        view.content.errata.type_filter.fill(type)
+        return view.read(widget_names="content.errata.table")
 
     def apply_erratas(self, entity_name, search):
         """Apply errata on selected host based on errata_id"""
         view = self.navigate_to(self, 'NewDetails', entity_name=entity_name)
         view.wait_displayed()
-        view.Content.Errata.searchbar.fill(search)
-        view.Content.Errata.select_all.click()
-        view.Content.Errata.apply.click()
+        view.content.errata.searchbar.fill(search)
+        # wait for filter to apply
+        self.browser.plugin.ensure_page_safe()
+        view.content.errata.select_all.click()
+        view.content.errata.apply.click()
         view.flash.assert_no_error()
         view.flash.dismiss()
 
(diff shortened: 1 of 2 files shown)


airgun/views/host_new.py
```diff
@@ -63,27 +70,108 @@ class Overview(Tab):
...
+
+            security_advisory = Text('.//a[contains(@href, "type=security")]')
+            bug_fixes = Text('.//a[contains(@href, "type=bugfix")]')
+            enhancements = Text('.//a[contains(@href, "type=enhancement")]')
+
+        @View.nested
+        class TotalRisksCard(Card):
+            ROOT = './/article[.//div[text()="Total risks"]]'
+
+            low = Text('.//*[@id="legend-labels-0"]/*')
```
peterdragun, line 98: the correct locator here would be `'.//text[@id="legend-labels-0"]/tspan'`, any suggestions why this is not working? With wildcards, everything is working as expected. Can wildcard stay here as this is matched by id?

omkarkhatavkar: regarding `'.//text[@id="legend-labels-0"]/tspan'` my guess would be `text` is not a valid tag.

tstrych (approved):
LGTM, nice work :rocket: 
Please link the PR once you have it, let us see that the locators work :smiley:

airgun/views/host_new.py
peterdragun: tag `text` is probably not the issue here (at least not the only one) as even the `'.//*[@id="legend-labels-0"]/tspan'` is not working.

airgun/views/host_new.py
```diff
@@ -63,27 +70,108 @@ class Overview(Tab):
...
         @View.nested
         class DetailsCard(Card):
-            ROOT = '(.//article[contains(@class, "pf-c-card")])[1]'
+            ROOT = './/article[.//div[text()="Details"]]'
             details = HostDetailsCard()
 
         @View.nested
         class HostStatusCard(Card):
-            ROOT = '(//article[contains(@class, "pf-c-card")])[2]'
+            ROOT = './/article[.//span[text()="Host Status"]]'
```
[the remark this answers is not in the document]
peterdragun: Yeah, probably the IDs would be better here :+1:, just wanted to make it work with dynamic cards.

airgun/views/host_new.py
omkarkhatavkar: I looked at the locator, it seems to be valid how you have written as it is the react component, there seems to be not located standard way of xpath.

airgun/entities/host_new.py
```diff
@@ -23,6 +26,81 @@ def get_details(self, entity_name, widget_names=None):
...
+        """Installs package on host using the installation modal"""
+        view = self.navigate_to(self, 'NewDetails', entity_name=entity_name)
+        view.Content.Packages.select()
+        view.Content.Packages.dropdown.wait_displayed()
+        view.Content.Packages.dropdown.item_select('Install packages')
+        view = InstallPackagesView(self.browser)
+        view.wait_displayed()
+        view.searchbar.fill(package)
+        # wait for filter to apply
+        time.sleep(2)
```
omkarkhatavkar, line 48: I'm not against/harsh the hard sleep in a single place, but that is the consistent multiple places we really need to think about what causes such behavior/pattern. Can you use the before/after_fill to handle such cases (use the wait_for)?

omkarkhatavkar (approved):
ACK, pending comment

airgun/entities/host_new.py
peterdragun: So the problem here is that basically after filling the filter the table is still visible with old data. So I am not sure for what should I wait, as the table is visible and everything seems to be "correct". Maybe I can use after_fill but still, it will need to contain the sleep. I have tried avoiding using sleep, but haven't figured out a better way, if you have some ideas let me know.

airgun/entities/host_new.py
```diff
@@ -23,6 +27,85 @@ def get_details(self, entity_name, widget_names=None):
         view = self.navigate_to(self, 'NewDetails', entity_name=entity_name)
         return view.read(widget_names=widget_names)
 
+    def get_packages(self, entity_name, search=""):
+        """Filter installed packages on host"""
+        view = self.navigate_to(self, 'NewDetails', entity_name=entity_name)
+        view.Content.Packages.select()
```
mshriver, line 33: `Content` and `Packages` are `Tab` classes, this `select()` is redundant, `Tab` implements `child_widget_accessed` and clicks itself. You should be able to directly call `searchbar.fill`

peterdragun: Yeah I thought so, but without this line, it was actually not able to locate everything on that page for some reason

airgun/entities/host_new.py
mshriver: The `ensure_page_safe` checks should be handling this automatically in the browser plugin.

Something about the page should trigger the checks in this function, letting us know that the page isn't ready to interact with yet:
https://github.com/SatelliteQE/airgun/blob/986ff09d6a578414d5f4099cd776082a9f1d77a9/airgun/browser.py#L192

airgun/views/host_new.py
```diff
@@ -63,35 +77,145 @@ class Overview(Tab):
...
+            searchbar = TextInput(locator='.//input[contains(@class, "pf-m-search")]')
+            status_filter = Dropdown(locator='.//div[@aria-label="select Status container"]/div')
+            upgrade = Button('Upgrade')
+            dropdown = Dropdown(locator='.//div[button[@aria-label="bulk_actions"]]')
+
+            table = PatternflyTable(
+                component_id="host-packages-table",
+                column_widgets={
+                    0: Checkbox(locator='.//input[@type="checkbox"]'),
+                    'Package': Text('./parent::td'),
```
mshriver, line 133: Are these clickable cells?

peterdragun: They could be if the package is installed from synced repo on satellite

airgun/views/host_new.py
```diff
@@ -63,35 +77,145 @@ class Overview(Tab):
...
         @View.nested
         class RecentJobsCard(Card):
-            ROOT = '(//article[contains(@class, "pf-c-card")])[5]'
+            ROOT = './/article[.//div[text()="Recent jobs"]]'
             is_table_loaded = './/ul[@aria-label="recent-jobs-table"]'
 
     @View.nested
     class Content(Tab):
-        pass
+        ROOT = './/div'
```
mshriver, line 117: I'd suggest not setting ROOT for tab widgets. Its overriding the ParametrizedLocators that use TAB_NAME (or the class name), and is likely causing the behavior where you're having to manually select tabs.

airgun/entities/host_new.py
mshriver: Started convo about this in chat as it could take some debug - starting with suggestion to not set ROOT for tab classes.

airgun/views/host_new.py
peterdragun: I will try it, thanks for the suggestion :+1:

mshriver: We did some debugging this morning, and found that the sub-tabs here can't resolve their elements when nested. A workaround is to override `Tab.child_widget_accesssed` on the subtab classes, and first call `self.parent.select()` before calling `self.select()`.

peterdragun: @mshriver As the workaround does not work, are you okay with using the ROOT?

latran (approved):
Really nice work here!  acking, pending some tests that uses these if available.

peterdragun:
@latran tests can be found in linked PR to robotello: https://github.com/SatelliteQE/robottelo/pull/9822

airgun/views/host_new.py
peterdragun: Adding BZ link and TODO

mshriver (approved):
I still think some of these locators are a bit problematic, but this is an excellent starting point.

Thanks for digging into the tab behavior with me and identifying the PF4 integration bug on the page.

PR #723, opened 2022-07-13
Adding Webhooks CRUD

Griffin-Sullivan:
Basic Create, Update, Read, and Delete functions added for Administer > Webhooks. There are no unique ids for the AutoCompleteTextInputs on the Create page, so I am searching for them based on their neighboring `<label>` elements right now. I will reach out to UX to get some unique ids for these, so we don't have to worry about new updates to the DOM. Because the Webhook create, update, and read is based on a popup window (there's no separate page for these actions), we have to do some waiting for elements to be loaded in the popup which is handled by `wait_for_popup`. There are two checks because the popup and it's child elements all load asynchronously of each other.

### DIFF
--- a/airgun/entities/webhook.py
+++ b/airgun/entities/webhook.py
@@ -1,0 +1,134 @@
+from navmazing import NavigateToSibling
+from widgetastic.exceptions import NoSuchElementException
+
+from airgun.entities.base import BaseEntity
+from airgun.navigation import NavigateStep
+from airgun.navigation import navigator
+from airgun.utils import retry_navigation
+from airgun.views.webhook import DeleteWebhookConfirmationView
+from airgun.views.webhook import WebhookCreateView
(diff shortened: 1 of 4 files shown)


omkarkhatavkar:
@Griffin-Sullivan You can run PRT with this PR in robottelo to get the test result as well.

airgun/entities/webhook.py
```diff
@@ -0,0 +1,134 @@
...
+class DeleteWebhook(NavigateStep):
+    """Search for Webhook and confirm deletion in dialog.
+
+    Args:
+        entity_name: name of the Webhook
+    """
+
+    VIEW = DeleteWebhookConfirmationView
+
+    def prerequisite(self, *args, **kwargs):
```
mshriver, line 128: `NavigateToSibling`

mshriver (approved):

PR #724, opened 2022-07-14
Skipping ensure_page_safe on timeout. 

lhellebr:
Workaround for e.g. https://bugzilla.redhat.com/show_bug.cgi?id=2106022

### DIFF
--- a/airgun/browser.py
+++ b/airgun/browser.py
@@ -12,7 +12,6 @@ from urllib.parse import unquote
 import yaml
 from box import Box
 from selenium import webdriver
-from wait_for import TimedOutError
 from wait_for import wait_for
 from webdriver_kaifuku import BrowserManager
 from widgetastic.browser import Browser
@@ -237,20 +236,11 @@ class AirgunBrowserPlugin(DefaultPlugin):
         }
         '''
 
-    success_on_timeout = False
-
     def ensure_page_safe(self, timeout='30s'):
         """Ensures page is fully loaded.
         Default timeout was 10s, this changes it to 30s.
-        If self.success_on_timeout is True, the function doesn't raise an exception
-        and continues as if the page was safe instead. This can be used to bypass
-        some bugs, e.g. https://bugzilla.redhat.com/show_bug.cgi?id=2106022
         """
-        try:
-            super().ensure_page_safe()
-        except TimedOutError as e:
-            if not self.success_on_timeout:
-                raise e
+        super().ensure_page_safe(timeout)
 
     def before_click(self, element, locator=None):
         """Invoked before clicking on an element. Ensure page is fully loaded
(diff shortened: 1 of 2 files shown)


airgun/browser.py
```diff
@@ -236,11 +237,20 @@ class AirgunBrowserPlugin(DefaultPlugin):
...
+        If self.success_on_timeout is True, the function doesn't raise an exception
+        and continues as if the page was safe instead. This can be used to bypass
+        some bugs, e.g. https://bugzilla.redhat.com/show_bug.cgi?id=2106022
         """
-        super().ensure_page_safe(timeout)
+        try:
+            super().ensure_page_safe()
+        except TimedOutError as e:
+            if not self.success_on_timeout:
+                raise e
```
mshriver, line 253: No need to capture the exception object here, you can drop `as e` and `raise e`, just calling `raise` will propagate the exception object.

lhellebr:
I wanted to be simple here because I don't see this being used often but you're right, let's make things properly. I'll change the PR alongside your example.

lhellebr:
Fixed, the test run failed with an unrelated error.
`selenium.common.exceptions.NoSuchElementException: Message: Could not find an element Locator(by='xpath', locator="//div[@id='scsi_controllers']")`

mshriver (approved):
Looks great! Thanks for taking the time to make this enhancement instead of working around in the impacted test only.

latran:
@lhellebr Can you provide a test case that causes the failure?

lhellebr:
`tests/foreman/ui/test_computeresource_vmware.py::test_positive_access_vmware_with_custom_profile`
Perhaps others but this is my reproducer

latran (approved):
acking.  I will investigate your unrelated error and see what's going on.

lhellebr:
> acking. I will investigate your unrelated error and see what's going on.

Don't, I'm working on it right now. Some locator changes.

airgun/browser.py
```diff
@@ -236,11 +238,30 @@ class AirgunBrowserPlugin(DefaultPlugin):
...
     def ensure_page_safe(self, timeout='30s'):
         """Ensures page is fully loaded.
         Default timeout was 10s, this changes it to 30s.
+        If self.ignore_ensure_page_safe_timeout is True, the function doesn't raise an exception
+        and continues as if the page was safe instead. This can be used to bypass
+        some bugs, e.g. https://bugzilla.redhat.com/show_bug.cgi?id=2106022
         """
-        super().ensure_page_safe(timeout)
+        try:
+            super().ensure_page_safe()
```
lhellebr, line 261: I somehow dropped the timeout parameter from a function all. This will need to be fixed in a separate PR.

PR #725, opened 2022-07-14
Fixing PF4ConfirmationDialog

Griffin-Sullivan:
Was seeing some errors in 6.12 since these PF4 Buttons were changed in the DOM a little (might be worth getting static ids for these?). Switched out the type of widget here so we could locate with locators and button text. I used `aria-label` for the close button so that it could be used in 6.11 also. Although I think most people will be using the cancel button instead.

### DIFF
--- a/airgun/widgets.py
+++ b/airgun/widgets.py
@@ -1134,9 +1134,9 @@ class Pf4ConfirmationDialog(ConfirmationDialog):
     right corner."""
 
     ROOT = '//div[@id="app-confirm-modal" or @data-ouia-component-type="PF4/ModalContent"]'
-    confirm_dialog = PF4Button('OUIA-Generated-Button-danger-1')
-    cancel_dialog = PF4Button('OUIA-Generated-Button-link-1')
-    discard_dialog = PF4Button('OUIA-Generated-Button-plain-1')
+    confirm_dialog = PF4Button('btn-modal-confirm')
+    cancel_dialog = PF4Button('btn-modal-cancel')
+    discard_dialog = PF4Button('OUIA-Generated-Modal-small-1-ModalBoxCloseButton')
 
 
 class LCESelector(GenericLocatorWidget):


omkarkhatavkar (approved):

airgun/widgets.py
```diff
@@ -1134,9 +1134,9 @@ class Pf4ConfirmationDialog(ConfirmationDialog):
     right corner."""
 
     ROOT = '//div[@id="app-confirm-modal" or @data-ouia-component-type="PF4/ModalContent"]'
-    confirm_dialog = PF4Button('OUIA-Generated-Button-danger-1')
-    cancel_dialog = PF4Button('OUIA-Generated-Button-link-1')
-    discard_dialog = PF4Button('OUIA-Generated-Button-plain-1')
+    confirm_dialog = PF4Button('Confirm')
+    cancel_dialog = PF4Button('Cancel')
+    discard_dialog = PF4Button('locator=//button[contains(@aria-label, "Close")]')
```
jyejare, line 1139:
```suggestion
    discard_dialog = PF4Button(locator='//button[contains(@aria-label, "Close")]')
```
??

jyejare (changes requested):

airgun/widgets.py
```diff
@@ -20,8 +20,8 @@
 from widgetastic_patternfly import FlashMessages
 from widgetastic_patternfly import Kebab
 from widgetastic_patternfly import VerticalNavigation
+from widgetastic_patternfly4.button import Button as PF4Button
```
mshriver, line 23: All of these dialog buttons should have static ouiaID's.

https://github.com/theforeman/foreman/commit/4bcd67ddb14abad4c9b44e213e19e1dfda7c8b7e

I don't think this is going the right direction.

mshriver (changes requested):
These buttons should have static ouiaID's already

airgun/widgets.py
mshriver: ```<button aria-disabled="false" class="pf-c-button pf-m-danger" type="button" data-ouia-component-type="PF4/Button" data-ouia-safe="true" data-ouia-component-id="btn-modal-confirm">Confirm</button>```

The buttons on the PF4 confirmation dialog modal have data-ouia-component-id's set, you should be able to use them.  Their inclusion is probably what appears to have 'broken' the modal. 

image

mshriver: I pinged in the UX channel about the close button, its not as simple to set an ouiaID on because we'll have to define the entire header. There is an ouia component ID for the close button though.

mshriver (approved):

PR #730, opened 2022-07-27
fix locators for hardware models tests

peterdragun:
Targeted mainly for 6.12, but should not cause any issues on 6.11 as locators are the same for creating host and search is just adding another option.

### DIFF
--- a/airgun/entities/hardware_model.py
+++ b/airgun/entities/hardware_model.py
@@ -48,7 +48,7 @@ class HardwareModelEntity(BaseEntity):
         view.table.row(name=entity_name)['Actions'].widget.click()
         view.delete_dialog.confirm()
         if err_message:
-            view.flash.assert_message(err_message)
+            view.flash.assert_message(f"Danger alert: {err_message}")
             view.flash.dismiss()
             view.delete_dialog.cancel()
         else:
--- a/airgun/views/host.py
+++ b/airgun/views/host.py
@@ -199,7 +199,7 @@ class HostInterface(View):
 class HostsView(BaseLoggedInView, SearchableViewMixin):
     title = Text("//h1[normalize-space(.)='Hosts']")
     export = Text(".//a[contains(@class, 'btn')][contains(@href, 'hosts.csv')]")
-    new = Text("//a[contains(normalize-space(.),'Create Host')]")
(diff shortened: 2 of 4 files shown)


airgun/views/host.py
```diff
@@ -199,7 +199,7 @@ def is_displayed(self):
 class HostsView(BaseLoggedInView, SearchableViewMixin):
     title = Text("//h1[normalize-space(.)='Hosts']")
     export = Text(".//a[contains(@class, 'btn')][contains(@href, 'hosts.csv')]")
-    new = Text("//a[contains(normalize-space(.),'Create Host')]")
+    new = Text(".//div[@id='rails-app-content']//a[contains(normalize-space(.),'Create Host')]")
```
peterdragun, line 202: The previous locator was matching also "Create Host" from the nav bar that was actually not visible, this will make sure that the only locator on the main page is matched. Tested for empty host view and also with some hosts.

omkarkhatavkar (approved):
ACK, pending test result.

peterdragun:
Test result:
image

airgun/entities/hardware_model.py
```diff
@@ -48,7 +48,7 @@ def delete(self, entity_name, err_message=''):
         view.table.row(name=entity_name)['Actions'].widget.click()
         view.delete_dialog.confirm()
         if err_message:
-            view.flash.assert_message(err_message)
+            view.flash.assert_message(f"Danger alert: {err_message}")
```
shweta83, line 51: @peterdragun Is this change applicable for 6.11.z as well? If not, we can add 6.12 label to this PR.

omkarkhatavkar, line 51: Why are hardcoding here? It should be coming from the test?

peterdragun: Yeah, it is, as PF4 notifications are also presented in 6.11. As mentioned changes were tested on both 6.12 and 6.11, to make sure that everything works.

peterdragun: This part of the notification is actually not visible, but the text is here to indicate the error. I think it makes more sense to have it hardcoded here than in the test.

JacobCallahan (approved):

PR #735, opened 2022-08-11
use new host UI by default

### LINKED ISSUE #695: new host details

peterdragun:
Fixes #695
After resolving BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2106691 we are using the new host UI by default. This would allow all tests that were using the old UI to run without any test changes.

Test result:
image

### DIFF
--- a/airgun/entities/host.py
+++ b/airgun/entities/host.py
@@ -26,6 +26,7 @@ from airgun.views.host import HostsJobInvocationStatusView
 from airgun.views.host import HostsUnassignCompliancePolicy
 from airgun.views.host import HostsView
 from airgun.views.host import RecommendationListView
+from airgun.views.host_new import NewHostDetailsView
 
 
 class HostEntity(BaseEntity):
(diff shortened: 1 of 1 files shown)


tstrych (approved):
LGTM, thanks

airgun/entities/host.py
```diff
@@ -327,6 +328,10 @@ def step(self, *args, **kwargs):
         entity_name = kwargs.get('entity_name')
         self.parent.search(entity_name)
         self.parent.table.row(name=entity_name)['Name'].widget.click()
+        host_view = NewHostDetailsView(self.parent.browser)
```
mshriver, line 331: VIEW for this class is still set to `HostDetailsView`, I don't think your change is fully addressing using the new view by default.

peterdragun: This change should not use the New host details UI, the only change here is that default navigation would be to the new UI where we need to click on the `Legacy UI` button, so the rest of the view will actually still be the `HostDetailsView`

Griffin-Sullivan: So we don't want to test the NewHostDetailsView yet? We are just using this to workaround for automation until that's ready?

peterdragun: This is workaround for those test that we already have implemented for old UI. I have separate PRs for the new host details page. As the UI changed so much it does not make sense to rewrite old testes.

Griffin-Sullivan (approved):

swadeley (approved):
ACK

PR #738, opened 2022-08-19
Fixing locator for login page footer

adarshdubey-star:
Locator for login page footer is updated which is causing test failures around components.

### DIFF
--- a/airgun/views/login.py
+++ b/airgun/views/login.py
@@ -7,10 +7,9 @@ from widgetastic.widget import View
 class LoginView(View, ClickableMixin):
     username = TextInput(id="login_login")
     password = TextInput(id="login_password")
-    login_text = Text('//*[@id="login-footer-text"]')
+    login_text = Text(".//footer[contains(@class,'login-pf-page-footer')]")
     logo = Text('//img[@alt="logo"]')
     submit = Text('//button[@type="submit"]')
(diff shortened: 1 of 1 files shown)


airgun/views/login.py
```diff
@@ -7,7 +7,7 @@
 class LoginView(View, ClickableMixin):
     username = TextInput(id="login_login")
     password = TextInput(id="login_password")
-    login_text = Text('//*[@id="login-footer-text"]')
+    login_text = Text(".//footer[contains(@class,'login-pf-page-footer')]")
     logo = Text('//img[@alt="logo"]')
     submit = Text('//button[@type="submit"]')
     version = Text('//*[@id="version"]')
```
peterdragun, line 13: Any reason not to delete `version` if you stopped using it and there isn't any element with this id?

shweta83: @peterdragun Version is removed from UI and it is moved as a part of footer so we are good to remove this @adarshdubey-star

shweta83 (approved):
Ack

airgun/views/login.py
adarshdubey-star: Yes, I was going to remove it just wanted to have an ack to remove it. @peterdragun

peterdragun (approved):

omkarkhatavkar (approved):

PR #741, opened 2022-08-19
[Draft] Add content source page and option in new host details page

spusateri:
New view and entity for the 6.12 content source changing feature

Will need to do some additional testing and fix the job invocation behavior

### DIFF
--- a/airgun/views/host.py
+++ b/airgun/views/host.py
@@ -684,10 +684,11 @@ class HostsChangeContentSourceView(BaseLoggedInView):
     content_view = Select("//select[@id=change_cs_content_view_select]")
     content_source = Select("//select[@id=change_cs_content_source_select]")
     change_button = Button("//button[@id=generate_btn]")
+    run_job_link = Link('//a[normalize-space(.)="Run job invocation"]')
 
     @property
     def is_displayed(self):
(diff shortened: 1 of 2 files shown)


airgun/views/host.py
```diff
@@ -679,6 +679,18 @@ def is_displayed(self):
...
+class HostsChangeContentSourceView(BaseLoggedInView):
+    environment = Select("//select[@id=change_cs_environment_select]")
+    content_view = Select("//select[@id=change_cs_content_view_select]")
+    content_source = Select("//select[@id=change_cs_content_source_select]")
+    change_button = Button("//button[@id=generate_btn]")
+    run_job_link = Link('//a[normalize-space(.)="Run job invocation"]')
+
+    @property
+    def is_displayed(self):
+        return self.browser.wait_for_element(self.change_button, exception=False) is not None
```
latran, line 691: Better to do an assert on `self.change_button.is_displayed()` here instead of doing a `wait_for_element` or another element on the page if that is what is expected to be displayed.

latran (changes requested):

omkarkhatavkar:

ogajduse:
Stale PR, closing.

PR #743, opened 2022-09-05
Ensure cockpit page is safe before locating elements

peterdragun:
Tests in linked robotello PR: SatelliteQE/robottelo#10011

### DIFF
--- a/airgun/entities/host.py
+++ b/airgun/entities/host.py
@@ -259,6 +259,7 @@ class HostEntity(BaseEntity):
 
         # switch to the last opened tab,
         self.browser.switch_to_window(self.browser.window_handles[-1])
+        self.browser.plugin.ensure_page_safe()
         self.browser.wait_for_element(locator='//div[@id="content"]/iframe', exception=True)
         # the remote host content is loaded in an iframe, let's switch to it
         self.browser.switch_to_frame(locator='//div[@id="content"]/iframe')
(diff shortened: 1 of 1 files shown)


airgun/entities/host.py
```diff
@@ -259,14 +259,15 @@ def get_webconsole_content(self, entity_name, rhel_version=7):
...
+        self.browser.plugin.ensure_page_safe()
         self.browser.wait_for_element(locator='//div[@id="content"]/iframe', exception=True)
         # the remote host content is loaded in an iframe, let's switch to it
         self.browser.switch_to_frame(locator='//div[@id="content"]/iframe')
 
         self.browser.wait_for_element(
             locator=f'//{hostname_element}[@id="{hostname_id}"]', exception=True, visible=True
         )
-        hostname_button_view = self.browser.selenium.find_elements_by_id(hostname_id)
+        hostname_button_view = self.browser.selenium.find_elements("id", hostname_id)
```
Griffin-Sullivan, line 270: This is possibly out of scope for the PR, but what exactly is the element being found? The variable name `hostname_button_view` makes it sound like we would be changing the view, but based on the code it seems like you are searching for an element by id and then getting the text from it. If my assumption is correct something like `hostname_button` makes more sense.

peterdragun: The reason here was probably just to indicate that this button is not from Satellite but from a cockpit view. I can see that it could indicate a change of the view itself, but the view was already changed by clicking on the `web console` button.

omkarkhatavkar (approved):

synkd (approved):

Griffin-Sullivan (approved):

PR #762, opened 2022-09-21
Remove publish via http field since it's no longer valid

latran:
This field was removed from previous sat version.  Need to check with dev if "unprotected" was its replacement or not.


latran:
Confirmed with dev that "Publish via HTTP" is replaced with "Unprotected."

See [PR#10094](https://github.com/SatelliteQE/robottelo/pull/10094) for test run result

airgun/views/repository.py
```diff
@@ -274,7 +270,6 @@ class FileRepository(View):
         upstream_url = EditableEntry(name='Upstream URL')
         verify_ssl = EditableEntryCheckbox(name='Verify SSL')
         upstream_authorization = AuthorizationEntry(name='Upstream Authorization')
-        publish_via_http = EditableEntryCheckbox(name='Publish via HTTP')
         unprotected = EditableEntryCheckbox(name='Unprotected')
```
Griffin-Sullivan, line 273: Are file and yum repos the only repository types that support unprotected?

Griffin-Sullivan: Just checked 6.12 and they are the only repo types with unprotected as an option.

Griffin-Sullivan (approved):

swadeley (approved):
ACK

PR #765, opened 2022-09-23
Publish cv and read cvv table for new cv UI

latran:
Adding the ability to publish a cv and read the cvv table after it's done.

### DIFF
--- a/airgun/views/contentview_new.py
+++ b/airgun/views/contentview_new.py
@@ -2,16 +2,24 @@ from widgetastic.widget import Checkbox
 from widgetastic.widget import Text
 from widgetastic.widget import TextInput
 from widgetastic.widget import View
-from widgetastic_patternfly4.ouia import Button as PF4Button
+from widgetastic_patternfly4 import Switch
+from widgetastic_patternfly4.ouia import BreadCrumb
+from widgetastic_patternfly4.ouia import Button
+from widgetastic_patternfly4.ouia import Dropdown
 from widgetastic_patternfly4.ouia import ExpandableTable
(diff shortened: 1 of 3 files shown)


airgun/widgets.py
```diff
@@ -2034,6 +2034,53 @@ def read(self):
         return self.progress
 
 
+class ProgressBarPF4(ProgressBar):
```
latran, line 2037: I keep hitting this error: ```selenium.common.exceptions.NoSuchElementException: Message: Could not find an element Locator(by='xpath', locator='.//div[contains(@class,"progress-bar")]')```.  It seems sometimes the automation recognizes that it's on the progress bar modal page and other times it just doesn't recognize it and failing.  I'm not sure what the best approach for capturing that progressbar modal is.  Any help is appreciated.

@mshriver @omkarkhatavkar

airgun/views/contentview_new.py
```diff
@@ -61,3 +69,99 @@ def is_displayed(self):
...
+    @View.nested
+    class repos(SatTab):
+        pass
+
+    @View.nested
+    class filters(SatTab):
+        pass
+
+
+class NewContentViewPublishView(BaseLoggedInView):
```
latran, line 139: This needs to use `baseModal` instead of the `BasedLoggedInView`.  The modal for publish uses both `wizard` and modal so it can be confusing to write something that makes sense here.

omkarkhatavkar:

sambible:
Closing as this work has been picked up and finished as part of more recent PRs.

PR #766, opened 2022-09-26
Fix OUIA IDs in hosts

peterdragun:
Some of the IDs were changed to static ones. 
There are still dynamic IDs in the registration form, as there was a new option for pull provider (included in this PR), the ID for LCE has increased.


peterdragun:
Not sure why checks are failing on doc creation, this PR is not changing anything related to docs and I am not able to find that part of docstring that it is failing on.

latran (approved):

omkarkhatavkar (approved):
ACK

PR #775, opened 2022-10-24
Fixes for Robottelo Role/Variable UI tests

sambible:
Couple of changes here to fix some role/Variable UI tests - including removing some "failsafe" checks that always failed

### DIFF
--- a/airgun/entities/ansible_role.py
+++ b/airgun/entities/ansible_role.py
@@ -37,12 +37,7 @@ class AnsibleRolesEntity(BaseEntity):
     def imported_roles_count(self):
         """Return the number of Ansible roles currently imported into Satellite"""
         view = self.navigate_to(self, 'All')
-        # Before any roles have been imported, no table or pagination widget are
-        # present on the page
-        if not view.pagination.is_displayed:
-            return 0
-        else:
-            return int(view.total_imported_roles.read())
+        return int(view.total_imported_roles.read())
 
     def import_all_roles(self):
         """Import all available roles and return the number of roles
--- a/airgun/entities/ansible_role.py
+++ b/airgun/entities/ansible_role.py
@@ -37,7 +37,12 @@ class AnsibleRolesEntity(BaseEntity):
     def imported_roles_count(self):
         """Return the number of Ansible roles currently imported into Satellite"""
         view = self.navigate_to(self, 'All')
-        return int(view.total_imported_roles.read())
+        # Before any roles have been imported, no table or pagination widget are
+        # present on the page
+        if not view.pagination.is_displayed:
+            return 0
+        else:
+            return int(view.total_imported_roles.read())
 
     def import_all_roles(self):
         """Import all available roles and return the number of roles
--- a/airgun/widgets.py
+++ b/airgun/widgets.py
@@ -1588,7 +1588,7 @@ class Pagination(Widget):
     and current page index/overall amount of pages. Mainly used with Table widget.
     """
 
-    ROOT = ".//form[contains(@class, 'content-view-pf-pagination')]"
+    ROOT = ".//foreman-react-component[contains(@name, 'Pagination')]"
     # Kattelo views use per_page with select, foreman use a per_page with Button DropDown.
     PER_PAGE_BUTTON_DROPDOWN = ".//div[button[@id='pagination-row-dropdown']]"
     PER_PAGE_SELECT = ".//select[contains(@ng-model, 'per_page')]"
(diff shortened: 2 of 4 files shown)


airgun/entities/ansible_role.py
```diff
@@ -37,12 +37,7 @@ def delete(self, entity_name):
     def imported_roles_count(self):
         """Return the number of Ansible roles currently imported into Satellite"""
         view = self.navigate_to(self, 'All')
-        # Before any roles have been imported, no table or pagination widget are
-        # present on the page
-        if not view.pagination.is_displayed:
```
omkarkhatavkar, line 42: why are you removing this check?

sambible: At least when I've been debugging the test, It's not properly respecting the actual pagination, and the test runs find without it.

synkd: I think the comment here provides valuable context, and I think that this property should return `0` if no roles have been imported. Without this check, the property returns an error if no roles have been imported. Consequently, I would request that you not remove these lines.

However, you're right that the check is not respecting the actual pagination. `view.pagination.is_displayed` returns false even when the pagination widget is present on the actual page. It looks to me like the pagination implementation has changed since I wrote this entity originally. I think the move here should be to address that discrepancy rather than remove this check. 

Looking at the `Pagination` widget [1], it's using `.//span[contains(@class, 'pagination-pf-items-total')]` as the locator for `total_items` (the same locator that you've changed to `//span[@class='pf-c-options-menu__toggle-text']/b[2]` in this PR) and `.//form[contains(@class, 'content-view-pf-pagination')]` as the `ROOT` locator for the widget. Looking at the pagination section in the HTML on the Ansible Roles page after importing roles on Satellite 6.12, I don't see either of those classes in use. The pagination widget is now in a `<foreman-react-component>` element instead of a `<form>` element. Looking at Satellite 6.11, the pagination widget HTML matches the locators currently defined in the Airgun `Pagination` widget.

So we need to refactor the `Pagination` widget for Satellite 6.12. Doing so may be outside the scope of this PR, so if you want to find an alternate way to check for the presence of the pagination widget as a temporary workaround, I think that would be acceptable as well.

[1] https://github.com/SatelliteQE/airgun/blob/007fe8bac68a980cd0ba5555ea535afda93b05df/airgun/widgets.py#L1586

synkd (changes requested):

sambible:
Posted a comment in Automation channel, but went ahead and included the necessary pagination widget changes here. I agree with @synkd that the widget needs updating as a whole though, I don't think it should work at all in current state.

omkarkhatavkar (approved):

sambible:
Updated the locators for more of the Pagination widget. I will admit, I can't find anywhere using the page controls though.

sambible:
There will be additional changes needed to fix various tests (Errata, content_host, settings, etc) that are currently failing, but this should pave the way to make that na easier task.

jyejare (approved):

jyejare:
@synkd Need your ACK/NACK here !

airgun/widgets.py
```diff
@@ -1588,17 +1588,17 @@ class Pagination(Widget):
...
     PER_PAGE_SELECT = ".//select[contains(@ng-model, 'per_page')]"
-    first_page_button = Text(".//li[a[span[contains(@class, 'angle-double-left')]]]")
-    previous_page_button = Text(".//li[a[span[contains(@class, 'angle-left')]]]")
-    next_page_button = Text(".//li[a[span[contains(@class, 'angle-right')]]]")
-    last_page_button = Text(".//li[a[span[contains(@class, 'angle-double-right')]]]")
+    first_page_button = Button(".//div[button[@data-action='first']]")
+    previous_page_button = Button(".//div[button[@data-action='previous']]")
+    next_page_button = Button(".//div[button[@data-action='next']]")
+    last_page_button = Button(".//div[button[@data-action='last']]")
     page = TextInput(locator=".//input[contains(@class, 'pagination-pf-page')]")
```
synkd, line 1599:
```suggestion
    page = TextInput(locator=".//input[contains(@class, 'pf-c-form-control')]")
```

airgun/widgets.py
```diff
@@ -1588,17 +1588,17 @@ class Pagination(Widget):
     and current page index/overall amount of pages. Mainly used with Table widget.
     """
 
-    ROOT = ".//form[contains(@class, 'content-view-pf-pagination')]"
+    ROOT = ".//foreman-react-component[contains(@name, 'Pagination')]"
     # Kattelo views use per_page with select, foreman use a per_page with Button DropDown.
     PER_PAGE_BUTTON_DROPDOWN = ".//div[button[@id='pagination-row-dropdown']]"
```
synkd, line 1593:
```suggestion
    PER_PAGE_BUTTON_DROPDOWN = ".//div[button[@id='paginationoptions-menu-toggle-3']]"
```

synkd, line 1593:
```suggestion
    PER_PAGE_BUTTON_DROPDOWN = ".//div[ul[@class='pf-c-options-menu__menu']]"
```
This is now represented as an unordered list with class `pf-c-options-menu__menu`, and each menu option is a list item with e.g. `data-action="per-page-5"`. I don't think it's necessary to add locators for the menu options, but this suggestion should make it easier to identify what needs to be added if someone needs to access those elements in the future.

synkd (changes requested):

synkd (approved):

PR #802, opened 2023-01-27
Updated Activation keys field location on global registration page

shweta83:
Activation keys field location on Global registration page has been moved to general tab from advanced tab hence updated the location.

Changes caused by [BZ 2059829](https://bugzilla.redhat.com/show_bug.cgi?id=2059829)

### DIFF
--- a/airgun/views/host.py
+++ b/airgun/views/host.py
@@ -502,6 +502,8 @@ class HostRegisterView(BaseLoggedInView):
         linux_host_init_link = Link('//a[normalize-space(.)="Linux host_init_config default"]')
         capsule = FormSelect('OUIA-Generated-FormSelect-default-5')
         insecure = Checkbox(id='reg_insecure')
+        activation_keys = BaseMultiSelect('OUIA-Generated-Select-typeaheadmulti-1')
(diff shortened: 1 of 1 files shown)


Griffin-Sullivan (approved):
Seems like it should work. Any results to share? I noticed that `OUIA-Generated-Select-typeaheadmulti-1` looks like it would change dynamically. However, after messing around in the UI it seems like the Select for AKs on this page always has that data-ouia-component-id. Only thing that could be a concern is if another Select is added with that id in the future.

shweta83:
> Seems like it should work. Any results to share? I noticed that `OUIA-Generated-Select-typeaheadmulti-1` looks like it would change dynamically. However, after messing around in the UI it seems like the Select for AKs on this page always has that data-ouia-component-id. Only thing that could be a concern is if another Select is added with that id in the future.

OUIA id is static(though it doesn't look like it) and we can update the selector in future if another selector is added. I am going to raise a PR in robottelo today for the changes in tests. You can hold the merge for this PR until robottelo PR is raised and then I will run PRT on the PR.

Gauravtalreja1 (approved):

Gauravtalreja1 (approved):

PR #819, opened 2023-03-14
new job-inv page

pondrejk:
Updated the remote execution view to accommodate the new job wizard

### DIFF
--- a/airgun/views/job_invocation.py
+++ b/airgun/views/job_invocation.py
@@ -1,19 +1,18 @@
 from wait_for import wait_for
-from widgetastic.widget import Checkbox
-from widgetastic.widget import ConditionalSwitchableView
 from widgetastic.widget import Text
 from widgetastic.widget import TextInput
 from widgetastic.widget import View
 from widgetastic_patternfly import BreadCrumb
-from widgetastic_patternfly import Button
+from widgetastic_patternfly4 import Button
+from widgetastic_patternfly4 import Radio
+from widgetastic_patternfly4.ouia import Select
 
 from airgun.views.common import BaseLoggedInView
 from airgun.views.common import SatTab
 from airgun.views.common import SatTable
 from airgun.views.common import SearchableViewMixin
+from airgun.views.common import WizardStepView
 from airgun.widgets import ActionsDropdown
-from airgun.widgets import FilteredDropdown
-from airgun.widgets import RadioGroup
 
 
 class JobInvocationsView(BaseLoggedInView, SearchableViewMixin):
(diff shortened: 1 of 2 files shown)


airgun/views/job_invocation.py
adarshdubey-star, line 1: Nonblocking: can we use single quotes when equating parameters with there respective Id [ id ='dummy']
i.e - time_span = TextInput(id='time-span')

airgun/views/job_invocation.py
```diff
@@ -1,19 +1,19 @@
...
 
 from airgun.views.common import BaseLoggedInView
 from airgun.views.common import SatTab
 from airgun.views.common import SatTable
 from airgun.views.common import SearchableViewMixin
 from airgun.widgets import ActionsDropdown
-from airgun.widgets import FilteredDropdown
-from airgun.widgets import RadioGroup
+
+# from widgetastic_patternfly4.ouia import Select
```
omkarkhatavkar, line 16: commented code?

airgun/views/job_invocation.py
```diff
@@ -28,125 +28,124 @@ def is_displayed(self):
...
+        job_category = Select('OUIA-Generated-Select-single-1')
+        job_template = Select('OUIA-Generated-Select-typeahead-1')
 
-    @template_content.register('Puppet Run Once - SSH Default')
-    class RunPuppetForm(View):
-        puppet_options = TextInput(id='puppet_options')
+        def __init__(self, parent, logger=None):
+            """Expand the selected wizard step"""
+            View.__init__(self, parent, logger=logger)
+            self.expander.click()
```
omkarkhatavkar, line 40: This can go to some base class no need to repeat that expander everytime e.g. 
```python 
class WizerdView(View):
    def __init__(self, parent, logger=None):
        """Expand the selected wizard step"""
        View.__init__(self, parent, logger=logger)
        self.expander.click()


class category_and_template(WizerdView):
    expander = Text(".//button[contains(.,'Category and Template')]")
    job_category = Select('OUIA-Generated-Select-single-1')
    job_template = Select('OUIA-Generated-Select-typeahead-1')```

**Review by omkarkhatavkar — APPROVED** [review]
ACK with smaller suggestion

File: airgun/views/job_invocation.py [hunk 3]
```diff
@@ -28,125 +27,89 @@ def is_displayed(self):
...
-        command = TextInput(id='command')
-
-    @template_content.register('Power Action - SSH Default')
-    class RestartHostForm(View):
-        action = FilteredDropdown(id='s2id_action')
 
-    @template_content.register('Puppet Run Once - SSH Default')
-    class RunPuppetForm(View):
-        puppet_options = TextInput(id='puppet_options')
+    class category_and_template(WizardStepView):
```
Gauravtalreja1, line 31: Can we please use the python naming convention here for class names, using either camelcase class names and no underscores in names? same for other classes
```suggestion
    class CategoryAndTemplate(WizardStepView):
```

Gauravtalreja1 (changes requested):

airgun/views/job_invocation.py
lhellebr: Also, for consistency with other non-class view items, it may be better to keep it as it is. Because in robottelo, you just view it as a "values" dict and it would be awful to have some keys lowercase and some uppercase. Hard decision.

lhellebr (approved):
Please provide a passing test run result. Otherwise, LGTM.

airgun/views/job_invocation.py
```diff
@@ -28,125 +27,89 @@ def is_displayed(self):
...
+        expander = Text(".//button[contains(.,'Category and Template')]")
+        job_category = Select('OUIA-Generated-Select-single-1')
+        job_template = Select('OUIA-Generated-Select-typeahead-1')
 
-    @template_content.register('Module Action - SSH Default')
-    class RunModuleForm(View):
-        action = FilteredDropdown(id='s2id_action')
+    class target_hosts_and_inputs(WizardStepView):
+        expander = Text(".//button[contains(.,'Target hosts and inputs')]")
+        # command action
```
omkarkhatavkar, line 38: In my opinion, these comments seem redundant and provide no added value. Instead, we could remove them and rely on the variable to determine the object type it represents.

airgun/views/job_invocation.py
pondrejk: done, thanks!

airgun/views/job_invocation.py
pondrejk: done, thanks

lhellebr:
That's one possible solution. Another would be a PR to Widgetastic. Yet another would be using a custom javascript that does the critical part without Widgetastic. Anything goes.

But I also think this is a valid bug - the documentation explicitly states that developers should not append to `document.body`, they should rather use `parent` because otherwise they can cause accessibility issues - which is exactly what happened.

Gauravtalreja1 (approved):

PR #826, opened 2023-04-04
NewHostEntity::get_details() wait check added

pnovotny:
Added `view.wait_displayed()` to `NewHostEntity::get_details()`
so it is ensured the host detail page is fully loaded.

### DIFF
--- a/airgun/entities/host_new.py
+++ b/airgun/entities/host_new.py
@@ -27,6 +27,7 @@ class NewHostEntity(HostEntity):
         will be read.
         """
         view = self.navigate_to(self, 'NewDetails', entity_name=entity_name)
+        view.wait_displayed()
         self.browser.plugin.ensure_page_safe()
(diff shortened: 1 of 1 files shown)


pnovotny:
trigger: test-robottelo
pytest: -k test_positive_read_details_page_from_new_ui tests/foreman/ui/test_host.py

lhellebr (approved):
I see a pattern of using both `ensure_page_safe` and `wait_displayed` at the same place - when waiting for the view to be loaded. Is it necessary? Probably yes since you're doing this fix. Maybe it would be worth investigating why exactly we need both and perhaps create a function that calls both for us.

Anyway, ACK, this only makes `get_details` do the same thing as other methods.

adarshdubey-star (approved):
Ack

pondrejk (approved):

PR #828, opened 2023-04-04
change locator for `Name` in `HostsView`

mstrlc:
Changed locator for `Name` in `views/host.py/HostsView`. Before, the locator was matching incorrectly to the value in Reccommendations column instead of Name because of similar values.

Fixed by adding condition that Name's locator should not include `'Insights'`, which is the Reccomendations locator.

In Hosts, Reccomendations value points to `'(hostname)/#Insights'` and Name value points to `'(hostname)'`.

### DIFF
--- a/airgun/views/host.py
+++ b/airgun/views/host.py
@@ -207,7 +207,9 @@ class HostsView(BaseLoggedInView, SearchableViewMixin):
         './/table',
         column_widgets={
             0: Checkbox(locator=".//input[@class='host_select_boxes']"),
-            'Name': Text("//a[contains(@href, '/new/hosts')]"),
+            'Name': Text(
+                "//a[contains(@href, '/new/hosts/') and not(contains(@href, 'Insights'))]"
+            ),
             'Recommendations': Text("./a"),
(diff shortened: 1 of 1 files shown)


adarshdubey-star:
@mstrlc try running pre-commit locally after a rebase and push again with changes

airgun/views/host.py
```diff
@@ -206,7 +206,7 @@ class HostsView(BaseLoggedInView, SearchableViewMixin):
         './/table',
         column_widgets={
             0: Checkbox(locator=".//input[@class='host_select_boxes']"),
-            'Name': Text("//a[contains(@href, '/new/hosts')]"),
+            'Name': Text(".//a[contains(@href, '/new/hosts/') and not(contains(@href, 'Insights'))]"),
```
adarshdubey-star, line 209: Looks functional but it will be good to have only one precise locator if we can achieve that.

Griffin-Sullivan (approved):
After poking around the UI it looks like this is your only solution for now. You can also ask the dev team to add an id for this locator if you'd like a cleaner solution. Please run pre-commit locally and push changes.

omkarkhatavkar (approved):

PR #830, opened 2023-04-07
Fix Pagination for Content Hosts

sambible:
Pagination wasn't working for Content Hosts, so this replaces the non functional Pagination widget with some more specific locators. Necessary for https://github.com/SatelliteQE/robottelo/pull/11193

### DIFF
--- a/airgun/views/contenthost.py
+++ b/airgun/views/contenthost.py
@@ -106,7 +106,9 @@ class ContentHostsView(BaseLoggedInView, SearchableViewMixin):
             'Installable Updates': InstallableUpdatesCellView(),
         },
     )
-    pagination = Pagination()
+
+    total_items = Text("//span[@class='pagination-pf-items-total ng-binding']")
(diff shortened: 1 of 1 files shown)


airgun/views/contenthost.py
```diff
@@ -106,7 +106,9 @@ class ContentHostsView(BaseLoggedInView, SearchableViewMixin):
             'Installable Updates': InstallableUpdatesCellView(),
         },
     )
-    pagination = Pagination()
```
jyejare, line 109: Any way we can fix the pagination for all the airgun entitites? 
- By using the unique id for `total_items` and `pages` through pagination function directly ?
- If the unique ID is not present, shouldnt we file a bug to the DEVs that the ID is changing just for Content Host entity ?

The changing ID has been consistent issue in Satellite UI.

jyejare (commented):
ACK pending question.

jyejare (approved):

LadislavVasina1 (approved):
Provided locators work. ACK

PR #833, opened 2023-04-26
Extension of NewHostUI entities

LadislavVasina1:
I have implemented new entities that allow us to do several actions on the new host details page: 
* Edit system purpose
* Add host to host collection
* Remove host from host collection
* Add new parameter
* Edit parameter
* Delete parameter
* Remediate with insights

I have also made some view extension that was needed, to create those entities.

### DIFF
--- a/airgun/entities/host_new.py
+++ b/airgun/entities/host_new.py
@@ -648,10 +648,44 @@ class NewHostEntity(HostEntity):
             view.insights.select_all_one_page.click()
             view.insights.select_all_pages.click()
         else:
-            view.insights.recommendations_table.sort_by('name', 'ascending')
-            view.insights.search_bar.fill(recommendation_to_remediate)
-            view.wait_displayed()
-            view.insights.recommendations_table[0][0].widget.click()
+            view.insights.recommendations_table.sort_by('Recommendation', 'ascending')
+
+            if type(recommendation_to_remediate) is list:
+                if not recommendation_to_remediate:
+                    raise ValueError('List of recommendations cannot be empty!')
+                for recommendation in recommendation_to_remediate:
+                    view.insights.click()
+                    # Excape double quotes in the recommendation
+                    recommendation = recommendation.replace('"', '\\"')
+                    recommendation = f'title = "{recommendation}"'
+                    view.insights.search_bar.fill(recommendation, enter_timeout=3)
+                    view.wait_displayed()
+                    self.browser.plugin.ensure_page_safe()
+                    time.sleep(3)
+                    try:
+                        # Click the checkbox of the first recommendation
+                        view.insights.recommendations_table[0][0].widget.click()
+                    except IndexError:
+                        raise IndexError(
+                            f'Recommendation {recommendation} not found on {entity_name}, '
+                            'thus cannot be remediated.'
+                        )
+            else:
+                # Excape double quotes in the recommendation
+                recommendation_to_remediate = recommendation_to_remediate.replace('"', '\\"')
+                recommendation_to_remediate = f'title = "{recommendation_to_remediate}"'
+                view.insights.search_bar.fill(recommendation_to_remediate)
+                view.wait_displayed()
+                self.browser.plugin.ensure_page_safe()
+                time.sleep(3)
+                try:
+                    # Click the checkbox of the first recommendation
+                    view.insights.recommendations_table[0][0].widget.click()
+                except IndexError:
+                    raise IndexError(
+                        f'Recommendation {recommendation_to_remediate} not found '
+                        f'on {entity_name}, thus cannot be remediated.'
+                    )
         view.insights.remediate.click()
         view = RemediationView(self.browser)
         view.remediate.click()
--- a/airgun/entities/host_new.py
+++ b/airgun/entities/host_new.py
@@ -90,13 +90,14 @@ class NewHostEntity(HostEntity):
 
         Args:
             entity_name: Name of the host
-            host_collection_name: Name of the host collection we want to add the host to
+            host_collection_name: Name  or list of the host collection we want to add the host to
             add_to_all_collections: If True, host will be added to all host collections
 
         Raises:
             ValueError: If specific hostColName is set to be removed and add all is set to True
                         or if no parameters are passed.
             ValueError: If host is already assigned to selected host collection.
+            ValueError: If there is empty list provided for host_collection_name.
             ValueError: If there are no host collections left for addition.
             ValueError: Given host collection name is not found in host collections.
         """
@@ -130,14 +131,27 @@ class NewHostEntity(HostEntity):
             raise ValueError('No host collections found or left for addition!')
 
         if not add_to_all_collections:
-            view.searchbar.fill("name = " + host_collection_name, enter_timeout=5)
-            view.wait_displayed()
-            self.browser.plugin.ensure_page_safe()
-            if view.host_collection_table.row_count == 0:
-                raise ValueError(f'{host_collection_name} not found in host collections!')
-            time.sleep(3)
-            # Select the host collection via checkbox in the table
-            view.host_collection_table[0][0].widget.click()
+            if type(host_collection_name) is list:
+                if not host_collection_name:
+                    raise ValueError('host_collection_name list is empty!')
+                for host_col in host_collection_name:
+                    view.searchbar.fill("name = " + host_col, enter_timeout=2)
+                    view.wait_displayed()
+                    self.browser.plugin.ensure_page_safe()
+                    if view.host_collection_table.row_count == 0:
+                        raise ValueError(f'{host_col} not found in host collections!')
+                    time.sleep(3)
+                    # Select the host collection via checkbox in the table
+                    view.host_collection_table[0][0].widget.click()
+            else:
+                view.searchbar.fill("name = " + host_collection_name, enter_timeout=2)
+                view.wait_displayed()
+                self.browser.plugin.ensure_page_safe()
+                if view.host_collection_table.row_count == 0:
+                    raise ValueError(f'{host_collection_name} not found in host collections!')
+                time.sleep(3)
+                # Select the host collection via checkbox in the table
+                view.host_collection_table[0][0].widget.click()
         else:
             view.select_all.click()
         view.add.click()
@@ -150,13 +164,14 @@ class NewHostEntity(HostEntity):
 
         Args:
             entity_name: Name of the host
-            host_collection_name: Name of the host collection we want to remove host from
+            host_collection_name: Name or list of the host collection we want to remove host from
             remove_from_all_collections: If True, host will be removed from all host collections
 
         Raises:
             ValueError: If specific hostColName is set to be removed & remove all is set to True
                             or if no parameters are passed.
             ValueError: If host is not assigned to any host collection.
+            ValueError: If there is empty list provided for host_collection_name.
             ValueError: If given host col name we want to remove is not assigned to host.
         """
 
@@ -181,16 +196,29 @@ class NewHostEntity(HostEntity):
         self.browser.plugin.ensure_page_safe()
 
         if not remove_from_all_collections:
-            view.searchbar.fill("name = " + host_collection_name, enter_timeout=5)
-            view.wait_displayed()
-            self.browser.plugin.ensure_page_safe()
-            if not view.host_collection_table.is_displayed:
-                raise ValueError(
-                    f"{host_collection_name} not assigned to host, thus can't remove it!"
-                )
-            time.sleep(3)
-            # Select the host collection via checkbox in the table
-            view.host_collection_table[0][0].widget.click()
+            if type(host_collection_name) is list:
+                if not host_collection_name:
+                    raise ValueError('host_collection_name list is empty!')
+                for host_col in host_collection_name:
+                    view.searchbar.fill("name = " + host_col, enter_timeout=2)
+                    view.wait_displayed()
+                    self.browser.plugin.ensure_page_safe()
+                    if not view.host_collection_table.is_displayed:
+                        raise ValueError(f"{host_col} not assigned to host, thus can't remove it!")
+                    time.sleep(3)
+                    # Select the host collection via checkbox in the table
+                    view.host_collection_table[0][0].widget.click()
+            else:
+                view.searchbar.fill("name = " + host_collection_name, enter_timeout=2)
+                view.wait_displayed()
+                self.browser.plugin.ensure_page_safe()
+                if not view.host_collection_table.is_displayed:
+                    raise ValueError(
+                        f"{host_collection_name} not assigned to host, thus can't remove it!"
+                    )
+                time.sleep(3)
+                # Select the host collection via checkbox in the table
+                view.host_collection_table[0][0].widget.click()
         else:
             view.select_all.click()
         view.remove.click()
--- a/airgun/entities/host_new.py
+++ b/airgun/entities/host_new.py
@@ -17,6 +17,10 @@ from airgun.views.host_new import RemediationView
 from airgun.views.job_invocation import JobInvocationCreateView
 
 
+global available_param_types
+available_param_types = ['string', 'boolean', 'integer', 'real', 'array', 'hash', 'yaml', 'json']
+
+
 class NewHostEntity(HostEntity):
     def create(self, values):
         """Create new host entity"""
@@ -58,7 +62,7 @@ class NewHostEntity(HostEntity):
 
         if not any([role, sla, usage, release_ver, add_ons]):
             raise ValueError(
-                'At least one of the role, sla, usage, release_ver, add_ons ' 'must be provided!'
+                'At least one of the role, sla, usage, release_ver, add_ons must be provided!'
             )
         view = self.navigate_to(self, 'NewDetails', entity_name=entity_name)
         view.wait_displayed()
@@ -86,25 +90,22 @@ class NewHostEntity(HostEntity):
 
         Args:
             entity_name: Name of the host
-            host_collection_name: Name of the host collection we want to remove host from
+            host_collection_name: Name of the host collection we want to add the host to
             add_to_all_collections: If True, host will be added to all host collections
 
         Raises:
-            ValueError: If specific hostColName is set to be removed and add all is set to True.
-            ValueError: If no parameters are passed.
+            ValueError: If specific hostColName is set to be removed and add all is set to True
+                        or if no parameters are passed.
             ValueError: If host is already assigned to selected host collection.
             ValueError: If there are no host collections left for addition.
             ValueError: Given host collection name is not found in host collections.
         """
 
-        if host_collection_name and add_to_all_collections:
+        if (host_collection_name and add_to_all_collections) or (
+            (host_collection_name is None) and (add_to_all_collections is False)
+        ):
             raise ValueError(
                 'Either host_collection_name or add_to_all_collections must be provided!'
-            )
-        if (host_collection_name is None) and (add_to_all_collections is False):
-            raise ValueError(
-                'host_collection_name None and add_to_all_collections False '
-                'is invalid combination!'
             )
 
         view = self.navigate_to(self, 'NewDetails', entity_name=entity_name)
@@ -135,6 +136,7 @@ class NewHostEntity(HostEntity):
             if view.host_collection_table.row_count == 0:
                 raise ValueError(f'{host_collection_name} not found in host collections!')
             time.sleep(3)
+            # Select the host collection via checkbox in the table
             view.host_collection_table[0][0].widget.click()
         else:
             view.select_all.click()
@@ -152,20 +154,17 @@ class NewHostEntity(HostEntity):
             remove_from_all_collections: If True, host will be removed from all host collections
 
         Raises:
-            ValueError: If specific hostColName is set to be removed & remove all is set to True.
-            ValueError: If no parameters are passed.
+            ValueError: If specific hostColName is set to be removed & remove all is set to True
+                            or if no parameters are passed.
             ValueError: If host is not assigned to any host collection.
             ValueError: If given host col name we want to remove is not assigned to host.
         """
 
-        if (host_collection_name is not None) and remove_from_all_collections:
+        if ((host_collection_name is not None) and remove_from_all_collections) or (
+            (host_collection_name is None) and (remove_from_all_collections is False)
+        ):
             raise ValueError(
-                'Either host_collection_name or remove_from_all_collections must be provided!'
-            )
-        if (host_collection_name is None) and (remove_from_all_collections is False):
-            raise ValueError(
-                'host_collection_name None and remove_from_all_collections False '
-                'is invalid combination!'
+                'Either host_collection_name or add_to_all_collections must be provided!'
             )
 
         view = self.navigate_to(self, 'NewDetails', entity_name=entity_name)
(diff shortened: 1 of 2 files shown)


airgun/entities/host_new.py
```diff
@@ -32,6 +38,163 @@ def get_details(self, entity_name, widget_names=None):
...
+            release_ver: Release version to be assigned
+            add_ons: Add-ons to be assigned
+
+        Raises:
+            ValueError: If no parameters are passed.
+        """
+
+        if not any([role, sla, usage, release_ver, add_ons]):
+            raise ValueError(
+                'At least one of the role, sla, usage, release_ver, add_ons ' 'must be provided!'
```
pondrejk, line 61: Is it intended to have two strings here, or formatting glitch?

airgun/entities/host_new.py
```diff
@@ -272,15 +435,146 @@ def get_parameters(self, entity_name):
...
         view = self.navigate_to(self, 'NewDetails', entity_name=entity_name)
         view.wait_displayed()
         self.browser.plugin.ensure_page_safe()
+        view.parameters.click()
+        view.parameters.searchbar.fill(parameter_name)
+        view.wait_displayed()
+        # Check if parameter with given name does not exist. If it does, raise ValueError.
+        if view.parameters.parameters_table.row_count != 0:
+            if view.parameters.parameters_table[0][0].text == parameter_name:
+                raise ValueError(f'Parameter with name {parameter_name} already exists')
```
pondrejk, line 468: from looking at the param search, this could yield multiple resutls if the searched param_name is a substring of other names. This could lead to false negative since you just check the first result here

airgun/entities/host_new.py
```diff
@@ -272,15 +435,146 @@ def get_parameters(self, entity_name):
...
+        """
+
+        if not any([new_parameter_name, new_parameter_type, new_parameter_value]):
+            raise ValueError(
+                'At least one of the new_parameter_name, new_parameter_type, new_parameter_value '
+                'must be provided!'
+            )
+        # Sanitize input
+        new_parameter_type = new_parameter_type.lower()
+        available_types = ['string', 'boolean', 'integer', 'real', 'array', 'hash', 'yaml', 'json']
```
pondrejk, line 509: could we have this defined somewhere just once, so it can be reused across functions?

airgun/entities/host_new.py
```diff
@@ -272,15 +435,146 @@ def get_parameters(self, entity_name):
...
+        self.browser.plugin.ensure_page_safe()
+        view.parameters.click()
+        view.parameters.searchbar.fill(parameter_to_change)
+        view.wait_displayed()
+        if view.parameters.parameters_table.row_count == 0:
+            raise ValueError(
+                f'Parameter {parameter_to_change} not found on {entity_name}, '
+                'thus cannot be edited.'
+            )
+        elif view.parameters.parameters_table[0][0].text != parameter_to_change:
```
pondrejk, line 524: same comment about the multiple results as above

airgun/entities/host_new.py
pondrejk: image

airgun/entities/host_new.py
```diff
@@ -272,15 +435,146 @@ def get_parameters(self, entity_name):
...
+            raise ValueError(
+                f'Parameter {parameter_to_change} not found on {entity_name}, '
+                'thus cannot be edited.'
+            )
+
+        view.parameters.searchbar.fill(new_parameter_name)
+        view.wait_displayed()
+        if view.parameters.parameters_table.row_count != 0:
+            if view.parameters.parameters_table[0][0].text == new_parameter_name:
```
pondrejk, line 533: here as well

airgun/entities/host_new.py
pondrejk: maybe you could request a normal key=value search at this tab, as it is on the insights tab, that would probably solve the issue as well, up to you

airgun/entities/host_new.py
LadislavVasina1: Formating glitch will look into that.

airgun/entities/host_new.py
```diff
@@ -32,6 +38,163 @@ def get_details(self, entity_name, widget_names=None):
...
+        """
+
+        if not any([role, sla, usage, release_ver, add_ons]):
+            raise ValueError(
+                'At least one of the role, sla, usage, release_ver, add_ons ' 'must be provided!'
+            )
+        view = self.navigate_to(self, 'NewDetails', entity_name=entity_name)
+        view.wait_displayed()
+        view.overview.system_purpose.edit_system_purpose.click()
+        view = EditSystemPurposeView(self.browser)
```
lhellebr, line 66: Why don't we use normal navigation here? I.e. `view = self.navigate_to(...` and something defined with `@navigator.register`?

airgun/entities/host_new.py
```diff
@@ -32,6 +38,163 @@ def get_details(self, entity_name, widget_names=None):
...
+        view.wait_displayed()
+        if role:
+            view.role.fill(role)
+        if sla:
+            view.sla.fill(sla)
+        if usage:
+            view.usage.fill(usage)
+        if release_ver:
+            view.release_version.fill(release_ver)
+        if add_ons:
```
lhellebr, line 77: Do we need to be so explicit with the conditions here? Can't we just call `fill` on the whole form with `fill` ignoring everything that's not defined? Isn't that what we do elsewhere?

airgun/entities/host_new.py
```diff
@@ -32,6 +38,163 @@ def get_details(self, entity_name, widget_names=None):
...
+
+    def add_host_to_host_collection(
+        self, entity_name, host_collection_name=None, add_to_all_collections=False
+    ):
+        """
+        Function that adds host to host collection
+
+        Args:
+            entity_name: Name of the host
+            host_collection_name: Name of the host collection we want to remove host from
```
lhellebr, line 89:
```suggestion
            host_collection_name: Name of the host collection we want to add the host to
```

airgun/entities/host_new.py
```diff
@@ -32,6 +38,163 @@ def get_details(self, entity_name, widget_names=None):
...
+            ValueError: Given host collection name is not found in host collections.
+        """
+
+        if host_collection_name and add_to_all_collections:
+            raise ValueError(
+                'Either host_collection_name or add_to_all_collections must be provided!'
+            )
+        if (host_collection_name is None) and (add_to_all_collections is False):
+            raise ValueError(
+                'host_collection_name None and add_to_all_collections False '
```
lhellebr, line 106: This message is too general. Can't we use these two conditions (both specified or neither specified) in one condition and in either case, just show `Either host_collection_name or add_to_all_collections must be provided!`?

airgun/entities/host_new.py
```diff
@@ -32,6 +38,163 @@ def get_details(self, entity_name, widget_names=None):
...
+
+        view = self.navigate_to(self, 'NewDetails', entity_name=entity_name)
+        view.wait_displayed()
+        self.browser.plugin.ensure_page_safe()
+        # Handle the case where there are no host collections assigned
+        if view.overview.host_collections.no_host_collections.is_displayed:
+            view.overview.host_collections.add_to_host_collection.click()
+        else:
+            if (
+                host_collection_name
```
lhellebr, line 118: Shouldn't we take a list instead of just one collection?

airgun/entities/host_new.py
```diff
@@ -32,6 +38,163 @@ def get_details(self, entity_name, widget_names=None):
...
+                in view.overview.host_collections.assigned_host_collections.read()
+            ):
+                raise ValueError(f'{host_collection_name} already assigned to host!')
```
lhellebr, line 121: Food for thought, is that really an error?

airgun/entities/host_new.py
LadislavVasina1: key=value that is present on the insights tab is not working here sadly

airgun/entities/host_new.py
```diff
@@ -32,6 +38,163 @@ def get_details(self, entity_name, widget_names=None):
...
+            view.overview.host_collections.kebab_menu.item_select('Add host to collections')
+
+        view = ManageHostCollectionModal(self.browser)
```
lhellebr, line 124: Same as comment somewhere above, can't we use standard navigation?

airgun/entities/host_new.py
```diff
@@ -32,6 +38,163 @@ def get_details(self, entity_name, widget_names=None):
...
+            raise ValueError('No host collections found or left for addition!')
+
+        if not add_to_all_collections:
+            view.searchbar.fill("name = " + host_collection_name, enter_timeout=5)
+            view.wait_displayed()
+            self.browser.plugin.ensure_page_safe()
+            if view.host_collection_table.row_count == 0:
+                raise ValueError(f'{host_collection_name} not found in host collections!')
+            time.sleep(3)
+            view.host_collection_table[0][0].widget.click()
```
lhellebr, line 138: What is that? I suppose a checkbox? Either address better or if unable, comment.

lhellebr (changes requested):
First set of comments, there will be more

airgun/entities/host_new.py
LadislavVasina1: hmm. I was thinking about this problem with the substring. But when I was testing that if I search for the parameter (and it existed) satellite puts that result as first every time.

airgun/entities/host_new.py
```diff
@@ -32,6 +38,163 @@ def get_details(self, entity_name, widget_names=None):
...
+            ValueError: If given host col name we want to remove is not assigned to host.
+        """
+
+        if (host_collection_name is not None) and remove_from_all_collections:
+            raise ValueError(
+                'Either host_collection_name or remove_from_all_collections must be provided!'
+            )
+        if (host_collection_name is None) and (remove_from_all_collections is False):
+            raise ValueError(
+                'host_collection_name None and remove_from_all_collections False '
```
lhellebr, line 167: Same as in the function above, why complicate things?

airgun/entities/host_new.py
```diff
@@ -32,6 +38,163 @@ def get_details(self, entity_name, widget_names=None):
...
+                'is invalid combination!'
+            )
+
+        view = self.navigate_to(self, 'NewDetails', entity_name=entity_name)
+        view.wait_displayed()
+        self.browser.plugin.ensure_page_safe()
+        # Handle the case where there are no host collections assigned
+        view.overview.click()
+        if view.overview.host_collections.no_host_collections.is_displayed:
+            raise ValueError('No host collections assigned to host, thus nothing to remove!')
```
lhellebr, line 177: Is that an error?

airgun/entities/host_new.py
LadislavVasina1: update: Satellite puts it first when you have table ordering set to ascending

airgun/entities/host_new.py
```diff
@@ -32,6 +38,163 @@ def get_details(self, entity_name, widget_names=None):
...
+        if not remove_from_all_collections:
+            view.searchbar.fill("name = " + host_collection_name, enter_timeout=5)
+            view.wait_displayed()
+            self.browser.plugin.ensure_page_safe()
+            if not view.host_collection_table.is_displayed:
+                raise ValueError(
+                    f"{host_collection_name} not assigned to host, thus can't remove it!"
+                )
+            time.sleep(3)
+            view.host_collection_table[0][0].widget.click()
```
lhellebr, line 193: Same as above, what is that element?

airgun/entities/host_new.py
```diff
@@ -32,6 +38,163 @@ def get_details(self, entity_name, widget_names=None):
...
+
+        if not remove_from_all_collections:
+            view.searchbar.fill("name = " + host_collection_name, enter_timeout=5)
+            view.wait_displayed()
+            self.browser.plugin.ensure_page_safe()
+            if not view.host_collection_table.is_displayed:
+                raise ValueError(
+                    f"{host_collection_name} not assigned to host, thus can't remove it!"
+                )
+            time.sleep(3)
```
lhellebr, line 192: Is this stable enough? If not, you can `wait_for()`

airgun/entities/host_new.py
```diff
@@ -272,15 +435,146 @@ def get_parameters(self, entity_name):
...
+        Raises:
+            ValueError: If parameter type is not valid
+            ValueError: If parameter with the same name already exists
+        """
+
+        # Sanitize input
+        parameter_type = parameter_type.lower()
+        available_types = ['string', 'boolean', 'integer', 'real', 'array', 'hash', 'yaml', 'json']
+        if parameter_type not in available_types:
+            raise ValueError(f'Parameter type {parameter_type} is not valid')
```
lhellebr, line 457: `Valid options are...`

airgun/entities/host_new.py
lhellebr: Then you may have to cycle through all the results, checking for equality.

airgun/entities/host_new.py
```diff
@@ -272,15 +435,146 @@ def get_parameters(self, entity_name):
...
+            if view.parameters.parameters_table[0][0].text == parameter_name:
+                raise ValueError(f'Parameter with name {parameter_name} already exists')
+
         view.parameters.add_parameter.click()
-        view.parameters.new_parameter_name.fill(parameter_name)
-        view.parameters.new_parameter_type.fill(parameter_type)
-        view.parameters.new_parameter_value.fill(parameter_value)
+        view.parameters.parameter_name_input.fill(parameter_name)
+        view.parameters.parameter_type_input.fill(parameter_type)
+        view.parameters.parameter_value_input.fill(parameter_value)
```
lhellebr, line 473: Same idea as above with calling `fill` for this whole thing

airgun/entities/host_new.py
```diff
@@ -272,15 +435,146 @@ def get_parameters(self, entity_name):
...
+                raise ValueError(
+                    f'Cannot rename {parameter_to_change} to {new_parameter_name}. '
+                    'This parameter already exists.'
+                )
+
+        view.parameters.searchbar.fill(parameter_to_change)
+        view.wait_displayed()
+        view.parameters.parameters_table[0][4].widget.click()
+        view.wait_displayed()
+        if new_parameter_name:
```
lhellebr, line 543: `fill` for the whole form?

airgun/entities/host_new.py
```diff
@@ -272,15 +435,146 @@ def get_parameters(self, entity_name):
...
+        view.wait_displayed()
+        self.browser.plugin.ensure_page_safe()
+        view.parameters.click()
+        view.parameters.searchbar.fill(parameter_name)
+        view.wait_displayed()
+        if view.parameters.parameters_table.row_count == 0:
+            raise ValueError(
+                f'Parameter {parameter_name} not found on {entity_name}, thus cannot be deleted.'
+            )
+        view.parameters.parameters_table[0][5].widget.item_select('Delete')
```
lhellebr, line 573: Also a problem with multiple results, right?

airgun/entities/host_new.py
```diff
@@ -314,6 +608,45 @@ def get_insights(self, entity_name):
...
+            ValueError: If both recommendation_to_remediate and remediate_all are provided
+        """
+
+        if (recommendation_to_remediate is not None) and remediate_all:
+            raise ValueError(
+                'Either recommendation_to_remediate or remediate_all must be provided!'
+            )
+        if (recommendation_to_remediate is None) and (remediate_all is False):
+            raise ValueError(
+                'recommendation_to_remediate None and remediate_all False is invalid combination!'
```
lhellebr, line 633: Simplifiable as above

airgun/entities/host_new.py
```diff
@@ -314,6 +608,45 @@ def get_insights(self, entity_name):
...
 
+    def remediate_with_insights(
+        self, entity_name, recommendation_to_remediate=None, remediate_all=False
+    ):
+        """
+        Function that can remediate all or one recommendation with insights.
+
+        Args:
+            entity_name: Name of the host on which recommendations are to be remediated
+            recommendation_to_remediate: Name of the recommendation to be remediated
```
lhellebr, line 619: Can we accept a list?

airgun/entities/host_new.py
```diff
@@ -314,6 +608,45 @@ def get_insights(self, entity_name):
...
+        self.browser.plugin.ensure_page_safe()
+        if remediate_all:
+            view.insights.select_all_one_page.click()
+            view.insights.select_all_pages.click()
+        else:
+            view.insights.search_bar.fill(recommendation_to_remediate)
+            view.wait_displayed()
+            view.insights.recommendations_table[0][0].widget.click()
+        view.insights.remediate.click()
+        view = RemediationView(self.browser)
```
lhellebr, line 647: Standard navigation as above?

lhellebr (commented):
This surely required a lot of work! Overally a nice commit :)

airgun/entities/host_new.py
LadislavVasina1: There is another line in that error saying that it is an invalid combination.

I wanted to do that as one condition but I also didn't want it to look like this:
```python
if (host_collection_name and add_to_all_collections) or ((host_collection_name is None) and (add_to_all_collections is False)):
    pass
```

so I chose this approach.

airgun/entities/host_new.py
lhellebr: What if
1) The searched term is something's suffix? I can imagine that works too since the search is only prefix-based?
2) You search for `text`, you find `textsuffix` and delete it in the deletion method? You really should check for equality.

airgun/entities/host_new.py
lhellebr: The above looks nice to me, if correctly split in lines and if `pass` is replaced by `raise...`

airgun/entities/host_new.py
LadislavVasina1: Good question. Probably not, but I think that the user of that function has the intention of adding the host to the new host collection and then checking if the host was actually added to that collection. So he should be notified that it didn't go according to his intentions in my opinion.

airgun/entities/host_new.py
LadislavVasina1: I can split it then.

with `pass` in that comment I was just lazy :D

airgun/entities/host_new.py
LadislavVasina1: I will add comment

airgun/entities/host_new.py
LadislavVasina1: Check box, will comment.

airgun/entities/host_new.py
lhellebr: Yes, you're right, especially since this is used for testing - we should be careful and explicit about those things.

airgun/entities/host_new.py
LadislavVasina1: I have tested it several times and it is working.

LadislavVasina1: I have added this as an extra precaution because it was sometimes reloading search result two times in a row.

airgun/entities/host_new.py
LadislavVasina1: image

airgun/entities/host_new.py
LadislavVasina1: I had same mindset as explained here https://github.com/SatelliteQE/airgun/pull/833#discussion_r1177804185
I think that the user of that function should know that function is not doing what he wants.

airgun/entities/host_new.py
LadislavVasina1: I have chosen this approach for implementing modal views like this because it is already implemented in the same way multiple times in `/view/host_new.py`. So I wanted to be consistent.

airgun/entities/host_new.py
LadislavVasina1: Same idea about implementing modals as explained here https://github.com/SatelliteQE/airgun/pull/833#discussion_r1177843673

airgun/entities/host_new.py
LadislavVasina1: It sure is a good idea. I can try to add that.

airgun/entities/host_new.py
lhellebr: And this paradigm is present in so many places already that I think it's not a blocker.

But the `navigator` paradigm is IMO at the core of airgun and should be followed where possible.

LadislavVasina1: Ok, I will try to take that in mind.

airgun/entities/host_new.py
lhellebr: More like print `available_types` as part of the error message.

airgun/entities/host_new.py
LadislavVasina1: @lhellebr After investigating for a while I have found out that this form is not implemented as a PF form (I couldn't even find basic PF Form implementation in widgetastic though) but as an individual PF FormSelects which needs to be filled one by one. I have also tried to look for some implementation of these forms by someone else, but everything I have found (i.e.  `views/host.py -> class general()`) is done in the same way as mine is.

airgun/entities/host_new.py
LadislavVasina1: Similar explanation as here: https://github.com/SatelliteQE/airgun/pull/833#discussion_r1179050031 
It is not implemented as form... and widgetastic does not support forms.

airgun/entities/host_new.py
LadislavVasina1: Similar explanation as here: https://github.com/SatelliteQE/airgun/pull/833#discussion_r1179050031 
It is not implemented as form... and widgetastic does not support forms.

airgun/entities/host_new.py
lhellebr: You're the expert in new hosts interface so I'll leave it up to you but look at this and consider something similar: https://github.com/SatelliteQE/airgun/blob/master/airgun/views/subscription.py#L108 and invocation https://github.com/SatelliteQE/airgun/blob/master/airgun/entities/subscription.py#L129

lhellebr (approved):

airgun/entities/host_new.py
```diff
@@ -130,14 +131,27 @@ def add_host_to_host_collection(
...
+                for host_col in host_collection_name:
+                    view.searchbar.fill("name = " + host_col, enter_timeout=2)
+                    view.wait_displayed()
+                    self.browser.plugin.ensure_page_safe()
+                    if view.host_collection_table.row_count == 0:
+                        raise ValueError(f'{host_col} not found in host collections!')
+                    time.sleep(3)
+                    # Select the host collection via checkbox in the table
+                    view.host_collection_table[0][0].widget.click()
+            else:
```
lhellebr, line 146: These are very repetitive, can't you use a function for that?

airgun/entities/host_new.py
```diff
@@ -181,16 +196,29 @@ def remove_host_from_host_collection(
...
+                for host_col in host_collection_name:
+                    view.searchbar.fill("name = " + host_col, enter_timeout=2)
+                    view.wait_displayed()
+                    self.browser.plugin.ensure_page_safe()
+                    if not view.host_collection_table.is_displayed:
+                        raise ValueError(f"{host_col} not assigned to host, thus can't remove it!")
+                    time.sleep(3)
+                    # Select the host collection via checkbox in the table
+                    view.host_collection_table[0][0].widget.click()
+            else:
```
lhellebr, line 211: Here as well

pondrejk (approved):

vsedmik (approved):

pnovotny (approved):

PR #843, opened 2023-05-05
Add trigger-robottelo-workflow for updating robottelo image on quay

jameerpathan111:
Problem statement: Currently, the Robottelo image is only updated when a Robottelo PR is merged, which means that it won't be updated if there are any changes merged in Airgun/Nailgun. This can cause issues when running automation as it might not have the latest code changes from these repositories.

Solution: We can add a new [repository_dispatch](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#repository_dispatch) trigger that [listens for any changes](https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#create-a-repository-dispatch-event) in the Airgun/Nailgun repositories. When any PR is merged in those repos, the trigger will run the workflow that updates the Robottelo image. This ensures that the Robottelo image is always up-to-date with the latest code changes.

See https://github.com/SatelliteQE/robottelo/pull/11428


JacobCallahan (approved):

.github/workflows/trigger_robottelo_workflow.yml
```diff
@@ -0,0 +1,20 @@
...
+    steps:
+      - name: Trigger workflow in robottelo repo
+        run: |
+          curl -L \
+            -X POST \
+            -H "Accept: application/vnd.github+json" \
+            -H "Authorization: Bearer ${{ secrets.CHERRYPICK_PAT }}" \
+            -H "X-GitHub-Api-Version: 2022-11-28" \
+            https://api.github.com/repos/SatelliteQE/robottelo/dispatches \
+            -d '{"event_type": "airgun-pr-merged", "client_payload":{"tag": "'"${GITHUB_REF##*/}"'"}}'
```
jyejare, line 20:
```suggestion
            -d '{"event_type": "airgun-pr-merged", "client_payload":{"branch": "'"${GITHUB_REF##*/}"'"}}'
```

.github/workflows/trigger_robottelo_workflow.yml
```diff
@@ -0,0 +1,20 @@
...
+      - master
+      - 6.*.z
+
+jobs:
+  trigger-robottelo-workflow:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Trigger workflow in robottelo repo
+        run: |
+          curl -L \
```
jyejare, line 14: You could also use readymade Github Action instead of a curl command:

https://github.com/peter-evans/repository-dispatch 

Not a blocker comment!

jyejare (approved):
ACK with nitpick and non-blocking improvement comment!

omkarkhatavkar (approved):
ACK

.github/workflows/trigger_robottelo_workflow.yml
jameerpathan111: I think we don't need to use github action as we don't have much code here anyway.
but I'll check it out later and evaluate it again.

.github/workflows/trigger_robottelo_workflow.yml
jameerpathan111: Updated it in both repos.

jameerpathan111:
@jyejare raised some concerns with using [repository_dispatch](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#repository_dispatch) and has asked me to look into [workflow_dispatch](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch) instead. Moving this to draft till I do further research on this.

JacobCallahan:
@jameerpathan111 what's the status of this PR?

jyejare:
@JacobCallahan This was now moved to Q3 Planning ! 

BTW @jameerpathan111 if you dont find to work on this please let us know and we can handover this to someone interested and has time for!

jameerpathan111:
@jyejare @JacobCallahan I have updated the PRs and now using workflow_dispatch :) Tested it with my dummy repos.

ogajduse (approved):

JacobCallahan (approved):

PR #852, opened 2023-06-12
[6.14 RFE] New Host and AK views and entities

ColeHiggins2:
New views and entities for Host and Activation Keys

Host:
- Added bulk override of repository sets
- Added new Repository Type field

Activation Keys:
- Get repository data for repository sets
- Added repository type field added to table

### DIFF
--- a/airgun/entities/activationkey.py
+++ b/airgun/entities/activationkey.py
@@ -36,7 +36,7 @@ class ActivationKeyEntity(BaseEntity):
         view = self.navigate_to(self, 'Edit', entity_name=entity_name)
         return view.read(widget_names=widget_names)
 
-    def get_repos(self, entity_name, repo_type="All", widget_names=None):
+    def get_repos(self, entity_name, repo_type="All"):
         """Read all values for repository sets on activation keys"""
         view = self.navigate_to(self, 'Edit', entity_name=entity_name)
         view.repository_sets.repo_type.select_by_visible_text(repo_type)
--- a/airgun/entities/activationkey.py
+++ b/airgun/entities/activationkey.py
@@ -36,6 +36,12 @@ class ActivationKeyEntity(BaseEntity):
         view = self.navigate_to(self, 'Edit', entity_name=entity_name)
         return view.read(widget_names=widget_names)
 
+    def get_repos(self, entity_name, repo_type="All", widget_names=None):
+        """Read all values for created activation key entity"""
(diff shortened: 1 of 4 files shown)


airgun/entities/activationkey.py
```diff
@@ -36,6 +36,12 @@ def read(self, entity_name, widget_names=None):
         view = self.navigate_to(self, 'Edit', entity_name=entity_name)
         return view.read(widget_names=widget_names)
 
+    def get_repos(self, entity_name, repo_type="All", widget_names=None):
```
omkarkhatavkar, line 39: widget_names is not unused

omkarkhatavkar (changes requested):
Please remove the get_repos method and update the read to handle this

airgun/entities/activationkey.py
ColeHiggins2: The normal read does work here... However, you are not able to select ```repo_type``` when you call ```activationkey.read```. We are specifically testing that we can filter via custom repo_type. If we were to modify the normal read on activation keys, we may have to modify other tests out of this scope... Additionally, ```widget_names='repository_sets'``` is not the only thing we use ```activationkey.read``` for. In my mind, it does not make sense to modify read with ```repo_type``` when other widgets can also be called.

LadislavVasina1: I agree with @ColeHiggins2 on the purpose of `get_repos` function, but I would still remove `widget_names` as suggested by Omkar parameter as it is unused.

omkarkhatavkar: this can be done easily without modifying the tests as well. 

```
def read(self, entity_name, widget_names=None, values=None):
        """Read all values for created activation key entity"""
        view.fill(values)
```
where as the `values={"repository_sets.repo_type" : "All"}`  neverless, I'm okay with this change. You have to remove the `widget_names`

Griffin-Sullivan (approved):
Ack pending @omkarkhatavkar's recent comment

airgun/entities/activationkey.py
omkarkhatavkar: non-blocking, please remove the unused `widget_names`

omkarkhatavkar (approved):
ACK pending one thing now

omkarkhatavkar:
@ColeHiggins2 please resolve the conflicts.

omkarkhatavkar (approved):

PR #869, opened 2023-06-25
Add Virtcard entity

sambible:
Adds support for the new Virtualization card added in 6.14. This view will only work for the version of the the card that's returned when the Virtualization platform of the host is RHV, as when other Virtualization platforms are used (vmware, etc) the card will have different fields, since it just displays the info that it recieves from the hypervisor. Not ideal, but I don't have a good way to automatically generate virtualized hosts of other types, so this should cover the basic functionality at least.

### DIFF
--- a/airgun/views/host_new.py
+++ b/airgun/views/host_new.py
@@ -344,6 +344,20 @@ class NewHostDetailsView(BaseLoggedInView):
         class networking_interface(Card):
             pass
 
+        @View.nested
+        class virtualization(Card):
+            ROOT = './/article[contains(@data-ouia-component-id, "card-template-Virtualization")]'
+
+            name = Text('.//div[contains(@class, "pf-c-description-list__group")][1]//dd')
+            core_per_socket = Text(
+                './/div[contains(@class, "pf-c-description-list__group")][2]//dd'
+            )
+            sockets = Text('.//div[contains(@class, "pf-c-description-list__group")][3]//dd')
+            memory = Text('.//div[contains(@class, "pf-c-description-list__group")][4]//dd')
+            display = Text('.//div[contains(@class, "pf-c-description-list__group")][5]//dd')
+            disk = Text('.//div[contains(@class, "pf-c-description-list__group")][6]//dd')
+            running_on = Text('.//div[contains(@class, "pf-c-description-list__group")][13]//dd')
+
     @View.nested
     class content(Tab):
         # TODO Setting ROOT is just a workaround because of BZ 2119076,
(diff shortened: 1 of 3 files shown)


sambible:
Turns out VMware is going to be the version of this card that i move forward with in CI, so I've changed the locators to reflect that. VMware is substantially longer, but the card location is the same. Just different contents. I'm going to post an IP here later today that will lead you to a Satellite where you can check out this card in action. Stay tuned.-

sambible:
https://github.com/SatelliteQE/robottelo/pull/11748

airgun/views/host_new.py
```diff
@@ -344,6 +344,34 @@ class networking_interfaces(Card):
...
+            public_ip_address = Text(
+                './/div[contains(@class, "pf-c-description-list__group")][4]//dd'
+            )
+            mac_address = Text('.//div[contains(@class, "pf-c-description-list__group")][5]//dd')
+            cpus = Text('.//div[contains(@class, "pf-c-description-list__group")][6]//dd')
+            cores_per_socket = Text(
+                './/div[contains(@class, "pf-c-description-list__group")][7]//dd'
+            )
+            firmware = Text('.//div[contains(@class, "pf-c-description-list__group")][8]//dd')
+            hypervisor = Text('.//div[contains(@class, "pf-c-description-list__group")][9]//dd')
```
jyejare, line 363: The XPath is very common in all of them (Expected) apart from their integer position value.

Could we simply generate the paths instead of repeating them through looping or any other better mechanism ?  

Not a blocking comment but could shorten the code and easy to maintain.

jyejare (approved):
ACK with improvement comment that could be done across airgun!

sambible:
https://ip-10-0-167-201.rhos-01.prod.psi.rdu2.redhat.com/new/hosts/automation-robottelo-vmware-vm.cfxufkg7e4#/Details

This is a link to a view of a host that has this card showing, if anyone wants to inspect it themselves.

airgun/views/host_new.py
LadislavVasina1: @sambible I think that my PR could be an inspiration on generating locators https://github.com/LadislavVasina1/airgun/blob/70689c1f2597c778df5fab62c9ce33beb89a7b69/airgun/entities/host_new.py#L427-L445

sambible:
Change org/location to Any as well, and it should show up

sambible:
Seemed to work. I'm going to file an issue to revisit this once we figure out why there are issues with CI being wonky. 
https://github.com/SatelliteQE/airgun/issues/873
I'll revisit this sometime after we are further along in 6.14

Griffin-Sullivan (approved):
Changes look good. Do you want to merge this or implement @jyejare's comment?

sambible:
Prefer merge, this is something I'd like to get into CI quickly. I'll circle back and consider @LadislavVasina1 work in another PR as a base perhaps.

JacobCallahan (approved):

PR #870, opened 2023-06-27
Update Global Registration OUIA IDs

shweta83:
OUIA IDs of Global Registration are updated and it is causing failures in stream. This change only impacts stream.

### DIFF
--- a/airgun/views/host.py
+++ b/airgun/views/host.py
@@ -497,8 +497,8 @@ class HostCreateView(BaseLoggedInView):
 
 
 class HostRegisterView(BaseLoggedInView):
-    generate_command = PF4Button('OUIA-Generated-Button-primary-1')
-    cancel = PF4Button('OUIA-Generated-Button-link-1')
+    generate_command = PF4Button('registration_generate_btn')
+    cancel = PF4Button('registration-cancel-button')
     registration_command = TextInput(locator="//input[@aria-label='Copyable input']")
 
     @View.nested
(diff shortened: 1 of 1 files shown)


lhellebr (approved):
Are these changes in product permanent and intentional? If yes and if the tests pass, ACK.

sambible (approved):
ACK pending some results showing this working in regular Satellite as well. I will say, this is far superior to the locators we were using previously, assuming these ouia IDs are in the product. Those previous one only worked because nothing else would ever use those same tags above those elements in the page :confused:

shweta83:
> Are these changes in product permanent and intentional? If yes and if the tests pass, ACK.

Since these changes are only in stream so I cannot say in confidence if it is permanent but this change will help with the PRT running on master and E2E test cases as well.

shweta83:
> ACK pending some results showing this working in regular Satellite as well. I will say, this is far superior to the locators we were using previously, assuming these ouia IDs are in the product. Those previous one only worked because nothing else would ever use those same tags above those elements in the page 😕

If you mean regular as 6.14.0 then No, this change currently impacts stream only.

omkarkhatavkar:
@shweta83 can you run some tests here using PRT?

omkarkhatavkar:
trigger: test-robottelo
pytest: tests/foreman/ui/test_host.py -k test_positive_global_registration_form

omkarkhatavkar:
PRT passed!

PR #875, opened 2023-06-28
[6.14.z] Add Virtcard entity

Satellite-QE:
Cherrypick of PR: https://github.com/SatelliteQE/airgun/pull/869

Adds support for the new Virtualization card added in 6.14. This view will only work for the version of the the card that's returned when the Virtualization platform of the host is RHV, as when other Virtualization platforms are used (vmware, etc) the card will have different fields, since it just displays the info that it recieves from the hypervisor. Not ideal, but I don't have a good way to automatically generate virtualized hosts of other types, so this should cover the basic functionality at least.

### DIFF
--- a/airgun/entities/computeresource.py
+++ b/airgun/entities/computeresource.py
@@ -91,9 +91,13 @@ class ComputeResourceEntity(BaseEntity):
         if vm['Power'].widget.read() == 'On':
             vm['Actions'].widget.click(handle_alert=True)
 
-    def vm_import(self, entity_name, vm_name, hostgroup, location):
+    def vm_import(self, entity_name, vm_name, hostgroup, location, org=None, name=None):
         """Imports the specified VM"""
         view = self.navigate_to(self, 'VMImport', entity_name=entity_name, vm_name=vm_name)
+        if name:
+            view.fill({'host.name': name})
+        if org:
+            view.fill({'host.organization': org})
         view.fill({'host.hostgroup': hostgroup, 'host.location': location})
         view.submit.click()
 
--- a/airgun/views/host_new.py
+++ b/airgun/views/host_new.py
@@ -344,6 +344,34 @@ class NewHostDetailsView(BaseLoggedInView):
         class networking_interface(Card):
             pass
 
+        @View.nested
+        class virtualization(Card):
+            ROOT = './/article[contains(@data-ouia-component-id, "card-template-Virtualization")]'
+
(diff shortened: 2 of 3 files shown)


airgun/entities/computeresource.py
```diff
@@ -91,9 +91,13 @@ def vm_poweroff(self, entity_name, vm_name):
...
             vm['Actions'].widget.click(handle_alert=True)
 
-    def vm_import(self, entity_name, vm_name, hostgroup, location):
+    def vm_import(self, entity_name, vm_name, hostgroup, location, org=None, name=None):
         """Imports the specified VM"""
         view = self.navigate_to(self, 'VMImport', entity_name=entity_name, vm_name=vm_name)
+        if name:
+            view.fill({'host.name': name})
+        if org:
+            view.fill({'host.organization': org})
```
omkarkhatavkar, line 100: Maybe I was not able to comment on the original PR, I feel there is not need to pass the extra parameters, you can pass everything from the test itself. 

e.g. 

`python 
values = {'host.name': name, 'host.hostgroup': hostgroup, 'host.location': location}}
session.ComputeResource().vm_import(entity_name, values)
`

airgun/views/host_new.py
```diff
@@ -344,6 +344,34 @@ class networking_interfaces(Card):
         class networking_interface(Card):
             pass
 
+        @View.nested
+        class virtualization(Card):
+            ROOT = './/article[contains(@data-ouia-component-id, "card-template-Virtualization")]'
+
+            datacenter = Text('.//div[contains(@class, "pf-c-description-list__group")][1]//dd')
+            cluster = Text('.//div[contains(@class, "pf-c-description-list__group")][2]//dd')
```
omkarkhatavkar, line 352: I'm not a fan of using hard number locators, these are so fragile that in case if any new field/label is added, this will disrupt all other locators. I see this pattern is been used across the whole class which I really think needs to improve in Airgun.

e.g . `$x("//div[@class='pf-c-description-list__group']//span[text()='Host group']/parent::*/following-sibling::*")`

you also could have looked at the existing custom card, https://github.com/SatelliteQE/airgun/blob/88a9f225088ca84ec865e10926522eb2fbf58ea8/airgun/views/host_new.py#L74 nicely return all the attributes and not need to add all the locators.

sambible: Good suggestion, I prefer this too. I didn't see that existing utility. I'll make an issue, and get a fix raised for this in master, and then merged into 6.13.z - lets merge this, and then I'll fix it shortly.

PR #876, opened 2023-06-29
Implementation of views and entities for new "Alternate Content Sources" page

LadislavVasina1:
Assingment `SAT-17105`

In this PR I have created new files `views/acs.py` and `entities/acs.py` to support new page in the Satellite called `Alternate Content Sources`.

The `views/acs.py` file contains the whole description of the new ACS page in the Satellite including the creation wizard.

The `entities/acs.py` file contains a wide spectrum of functions that enables users to perform actions over this ACS page.

i.e.:
- Create a new ACS
- Delete ACS
- Refresh ACS
- Get info on existing ACS
- Edit all parts of existing ACS
-  ...

### DIFF
--- a/airgun/entities/acs.py
+++ b/airgun/entities/acs.py
@@ -158,10 +158,7 @@ class AcsEntity(BaseEntity):
             )
 
         self.browser.plugin.do_refresh()
-        if action == 'Delete':
-            time.sleep(10)
-        elif action == 'Refresh':
-            time.sleep(5)
+        wait_for(lambda: view.title.is_displayed, timeout=10, delay=1)
 
     def refresh_all_acs(self):
         """Function that refreshes all ACS items"""
--- a/airgun/entities/acs.py
+++ b/airgun/entities/acs.py
@@ -17,14 +17,6 @@ from airgun.views.acs import RowDrawer
 
 
 class AcsEntity(BaseEntity):
-    def do_refresh(self):
-        """
-        Refresh the page as it is needed to refresh page to see changes after deleting an ACS
-        """
-
-        self.browser.refresh()
-        self.browser.plugin.ensure_page_safe()
-
     def wait_for_content_table(self, view):
         wait_for(lambda: view.acs_drawer.content_table.is_displayed, timeout=10, delay=1)
 
@@ -39,7 +31,7 @@ class AcsEntity(BaseEntity):
         Raises:
             ValueError: If row_id and acs_name are not specified or both specified
             ValueError: If no ACS is found
-            ValueEroor: If given ACS name does not exist
+            ValueError: If given ACS name does not exist
         """
 
         if (row_id is None and acs_name is None) or (row_id is not None and acs_name is not None):
@@ -62,9 +54,10 @@ class AcsEntity(BaseEntity):
             view.acs_drawer.content_table[0][1].widget.click()
         self.browser.plugin.ensure_page_safe()
         view.wait_displayed()
-        x = RowDrawer(self.browser).read()
+        result = RowDrawer(self.browser).read()
+        result['details']['last_refresh'] = RowDrawer(self.browser).last_refresh.text
         self.close_details_side_panel()
-        return x
+        return result
 
     def get_all_acs_info(self):
         """
@@ -75,6 +68,9 @@ class AcsEntity(BaseEntity):
         view = self.navigate_to(self, 'ACS')
         view.wait_displayed()
         self.browser.plugin.ensure_page_safe()
+        if view.acs_drawer.clear_search_btn.is_displayed:
+            view.acs_drawer.clear_search_btn.click()
+        wait_for(lambda: view.acs_drawer.content_table.is_displayed, timeout=10, delay=1)
         acs_table = view.acs_drawer.content_table.read()
         for i, row in enumerate(acs_table):
             row['details'] = self.get_row_drawer_content(row_id=i)
@@ -108,7 +104,6 @@ class AcsEntity(BaseEntity):
 
         for acs in acs_name:
             view.acs_drawer.search_bar.fill(f'name = {acs}')
-            time.sleep(3)
             if not view.acs_drawer.content_table.is_displayed:
                 raise ValueError(f'ACS {acs} not found!')
             # Check the checkbox of ACS item
@@ -123,8 +118,8 @@ class AcsEntity(BaseEntity):
         if action == 'Refresh':
             view.acs_drawer.clear_search_btn.click()
 
-        self.do_refresh()
-        time.sleep(10)
+        self.browser.plugin.do_refresh()
+        wait_for(lambda: view.acs_drawer.content_table.is_displayed, timeout=10, delay=1)
 
     def refresh_acs(self, acs_name):
         """Function that refreshes ACS item(s)"""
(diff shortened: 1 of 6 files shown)


airgun/entities/acs.py
```diff
@@ -0,0 +1,823 @@
...
+
+
+class AcsEntity(BaseEntity):
+    def do_refresh(self):
+        """
+        Refresh the page as it is needed to refresh page to see changes after deleting an ACS
+        """
+
+        self.browser.refresh()
+        self.browser.plugin.ensure_page_safe()
```
omkarkhatavkar, line 26: This is not specific to `AcsEntity` and can be moved elsewhere.  I would suggest moving this to `AirgunBrowserPlugin ` if this needs to be done often or I may update the  `refresh` method with an additional call of the plugin.ensure_page_safe()

pondrejk (approved):
did my best to go through it thoroughly, nice work

airgun/views/acs.py
```diff
@@ -0,0 +1,447 @@
...
+
+            products_list = ItemsList(locator='.//ul[contains(@class, "pf-c-list")]')
+
+    def read(self):
+        """Reads the ACS drawer and returns its content as a dictionary."""
+
+        keys = ['details', 'capsules', 'url_and_subpaths', 'credentials', 'products']
+        result = dict.fromkeys(keys)
+        # Sleep ensures that all stack items are loaded and test does not fail
+        time.sleep(3)
```
omkarkhatavkar, line 385: add the is_displayed to main view so these sleep is not required

airgun/views/acs.py
```diff
@@ -0,0 +1,447 @@
...
+        # Expand and read each stack item
+        for key in result:
+            if key == 'details':
```
omkarkhatavkar, line 388: you can avoid this completely if the naming of the item widget is the same as the key here.

airgun/views/acs.py
```diff
@@ -0,0 +1,447 @@
...
+        return result
+
+
+class AlternateContentSourcesView(BaseLoggedInView):
+    """Class that describes view of the Alternate Content Sources page."""
+
+    title = Text('//h1[contains(., "Alternate Content Sources")]')
+    error_message = Text('//div[contains(@aria-label, "Danger Alert")]')
+
+    @View.nested
```
omkarkhatavkar, line 419: I do not think, this is a `nested view`. looking at the UI I feel it is another All entity page view.

airgun/views/acs.py
```diff
@@ -0,0 +1,447 @@
...
+from airgun.widgets import ItemsList
+
+
+class EditModal(View):
+    """Class representing the Edit modal header"""
+
+    title = Text('.//h1')
+    close_button = OUIAButton('acs-edit-details-modal-ModalBoxCloseButton')
+
+    error_message = Text('//div[contains(@aria-label, "Danger Alert")]')
```
omkarkhatavkar, line 33: edit_button, cancel_button can be also included here, move this class to the common widget.py. So this can be further used for other entities

airgun/views/acs.py
```diff
@@ -0,0 +1,447 @@
...
+    ROOT = '//div[@data-ouia-component-id="acs-edit-details-modal"]'
+
+    name = OUIATextInput('acs-edit-name-field')
+    description = TextInput(locator='.//textarea[@id="acs_description_field"]')
+
+    edit_button = OUIAButton('edit-acs-details-submit')
+    cancel_button = OUIAButton('edit-acs-details-cancel')
+
+
+class DualListSelector(EditModal):
```
omkarkhatavkar, line 48: move this to common place

airgun/entities/acs.py
```diff
@@ -0,0 +1,823 @@
...
+
+@navigator.register(AcsEntity, 'ACS')
+class OpenAcsPage(NavigateStep):
+    """Navigate to the ACS page"""
+
+    VIEW = AlternateContentSourcesView
+
+    @retry_navigation
+    def step(self, *args, **kwargs):
+        self.view.menu.select('Content', 'Alternate Content Sources')
```
omkarkhatavkar, line 823: failed to see other navigations?

airgun/entities/acs.py
```diff
@@ -0,0 +1,823 @@
...
+
+            # Credentials
+            if verify_ssl:
+                view.credentials.verify_ssl_toggle.fill(True)
+                if ca_cert is not None:
+                    view.credentials.select_ca_cert.fill(ca_cert)
+            if manual_auth:
+                view.credentials.manual_auth_radio_btn.fill(True)
+                view.credentials.username.fill(username)
+                view.credentials.password.fill(password)
```
omkarkhatavkar, line 791: I dont think we need to call explicit `fill` methods

airgun/entities/acs.py
```diff
@@ -0,0 +1,823 @@
...
+
+        if add_all and remove_all:
+            raise ValueError('add_all and remove_all cannot be True at the same time!')
+        if (add_all or remove_all) and (
+            products_to_add is not None or products_to_remove is not None
+        ):
+            raise ValueError(
+                'add_all or remove_all cannot be True at the same time with '
+                'options_to_add or options_to_remove!'
+            )
```
omkarkhatavkar, line 548: I personally feel that this is an overhead for Airgun to handle these validations, Consider a case where someone wanted to test Airgun UI for invalid data, it won't happen because of validation.

omkarkhatavkar (changes requested):
Overall I like some of the approaches you made in the views but there should be a much better way you can handle the code for the entities. you can also add the custom fill in the view itself if you wanted to manage how you have managed the read.  

Like iterating over the data coming from the Robottelo in the form of layers and using that fill the view (nested ones)

airgun/entities/acs.py
```diff
@@ -0,0 +1,823 @@
...
+        if manual_auth is True and (username is None or password is None):
+            raise ValueError('username and password cannot be None when manual_auth is True!')
+
+        if rhui_type is True and manual_auth is True:
+            raise ValueError('manual_auth cannot be True when rhui_type is True!')
+
+        view = self.navigate_to(self, 'ACS')
+        view.wait_displayed()
+        self.browser.plugin.ensure_page_safe()
+        time.sleep(5)
```
omkarkhatavkar, line 689: please do not use the hard sleep

airgun/entities/acs.py
LadislavVasina1: After a discussion, I will use a flag to enable/disable those checks.

vsedmik (commented):
Views looks neat, still need to go through half of the entities but so far so good.
Great work anyway! :heart:

airgun/entities/acs.py
```diff
@@ -0,0 +1,832 @@
...
+        view.acs_drawer.select_all.click()
+        view.acs_drawer.kebab_menu.item_select(action)
+        if view.error_message.is_displayed:
+            raise ValueError(
+                f'Error while taking an action on all ACS: {view.error_message.read()}'
+            )
+
+        self.browser.plugin.do_refresh()
+        if action == 'Delete':
+            time.sleep(10)
```
omkarkhatavkar, line 162: can we please add the wait_for work around here, hard sleep should be avoided

omkarkhatavkar (changes requested):
I see hard sleep in most places, we have to use the wait_for. please update those things.

omkarkhatavkar (approved):

PR #877, opened 2023-06-29
Bump GHA python version check to 3.10 and 3.11

LadislavVasina1:
As our automation uses 3.11 at this moment, let's bump the version we check in the airgun from [3.8, 3.9] to [3.10, 3.11]

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


Griffin-Sullivan (approved):
Thanks for spotting this! Might need someone more familiar with the checks to verify we didn't miss something here.

ogajduse (approved):
LGTM. It does make sense to keep n-1 support, where `n = the version of Python that we use in automation`.

omkarkhatavkar:
trigger: test-robottelo

PR #887, opened 2023-07-04
fix in audit log search

pondrejk:
The problem is that with 6.14, the suggestions dropdown stays open even after the search is submitted, so we need to change focus from the search field to be able to expand the result. I didn't find a smarter way to do it other than clicking on the title. All audit log UI tests were failing due to this, with this change:

```
pytest tests/foreman/ui/test_audit.py -k test_positive_add_event
============================================================ test session starts ============================================================


tests/foreman/ui/test_audit.py .                                                                                                      [100%]


================================================ 1 passed, 5 deselected, 1 warning in 33.27s ==============================================
```

### DIFF
--- a/airgun/views/audit.py
+++ b/airgun/views/audit.py
@@ -49,4 +49,5 @@ class AuditsView(BaseLoggedInView):
 
     def search(self, query):
         self.searchbox.search(query)
+        self.title.click()  # to exit the search field
         return self.entry.read()


jyejare (approved):

jameerpathan111 (approved):

omkarkhatavkar:
@mstrlc see this problem, this could be handled in the search box class(implementation) itself. This is a search box that needs to click on outside to get rid of that drop-downs.

The same problem could be somewhere else as well in other entities.

PR #913, opened 2023-07-31
update Search widgets to PF4 class

mstrlc:
This PR updates views to use the common `SearchableViewMixinPF4` which uses the new `PF4Search` widget. Should help simplify the Search locators across different views.

**Changed files:**
- `airgun/`
	- `widgets.py`
- `airgun/views/`
	- `common.py`
	- `ansible_variable.py`
	- `architecture.py`
	- `audit.py`
	- `bookmark.py`
	- `cloud_insights.py`
	- `computeprofile.py`
	- `domain.py`
	- `hardware_model.py`
	- `host.py`
	- `hostgroup.py`
	- `job_template.py`
	- `location.py`
	- `media.py`
	- `modulestream.py`
	- `organization.py`
	- `os.py`
	- `oscapcontent.py`
	- `partitiontable.py`
	- `provisioning_template.py`
	- `report_template.py`
	- `role.py`
	- `settings.py`
	- `task.py`
	- `user.py`
	- `usergroup.py`

PR #914, opened 2023-08-01
Use PF4 Search in Webhooks

lhellebr:
Switches Webhooks search to PF4. Now, the search doesn't work. It will work after this PR and https://github.com/SatelliteQE/airgun/pull/913 are merged.


pondrejk (approved):

Gauravtalreja1 (approved):
Ack, just a qq, is PF4 search valid for all sat version you added labels for or just stream?

omkarkhatavkar (approved):

omkarkhatavkar:
@lhellebr are you confirm that this change is on all satellite versions?  or better to run the PRT and if it fails tells that which version it is not compatible

PR #915, opened 2023-08-02
change Navigation to Widgetastic PF4

mstrlc:
The Satellite vertical navigation component was changed, which broke navigating between views in UI tests. This PR uses the Navigation component from Widgetastic PatternFly4 library to fix this.


mstrlc:
The PRT passes for the test, previously here https://github.com/SatelliteQE/airgun/pull/913 it failed because of `NoSuchElementException`, this PR should fix this also for other tests.

Gauravtalreja1 (approved):

PR #919, opened 2023-08-07
Change entity routes to accomodate PF4 navigation with multiple levels

mstrlc:
In the old navigation, all items were visible and clickable after selecting first navigation level. In the new implementation of the PF4 Navigation, two levels must be clicked to see the destination.

This PR updates the routes so the Navigation works.

PR #930, opened 2023-08-11
Update puppet entity routes for PF4 navigation

Gauravtalreja1:
Missed during https://github.com/SatelliteQE/airgun/pull/919

### DIFF
--- a/airgun/entities/configgroup.py
+++ b/airgun/entities/configgroup.py
@@ -55,7 +55,7 @@ class ShowAllConfigGroups(NavigateStep):
 
     @retry_navigation
     def step(self, *args, **kwargs):
-        self.view.menu.select('Configure', 'Config Groups')
+        self.view.menu.select('Configure', 'Puppet ENC', 'Config Groups')
 
 
 @navigator.register(ConfigGroupEntity, 'New')
--- a/airgun/entities/puppet_class.py
+++ b/airgun/entities/puppet_class.py
@@ -51,7 +51,7 @@ class ShowAllPuppetClasses(NavigateStep):
 
     @retry_navigation
     def step(self, *args, **kwargs):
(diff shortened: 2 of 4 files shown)


Gauravtalreja1:
trigger: test-robottelo
pytest: tests/foreman/ui/test_puppetenvironment.py::test_positive_end_to_end

omkarkhatavkar (approved):

airgun/entities/host.py
```diff
@@ -308,7 +308,7 @@ class ShowAllHosts(NavigateStep):
 
     @retry_navigation
     def step(self, *args, **kwargs):
-        self.view.menu.select('Hosts', 'All Hosts')
+        self.view.menu.select('Hosts', 'Hosts', 'All Hosts')
```
mstrlc, line 311: I don't think it's needed to have 'Hosts' twice here (same for 'Configure'), as 'All Hosts' is directly under 'Hosts', only 2 levels down

Gauravtalreja1: yes, I know it might be skipped intensionally earlier, but to keep this pattern of paths consistent thought better to have this, do you think it could fail the tests?

omkarkhatavkar: @Gauravtalreja1 can you please test this?

mstrlc: Probably will not cause failures, but I feel the additional step in navigation adds unnecessary delay to tests as these navigation routes are used rather frequently.

Gauravtalreja1: @mstrlc yes, it shouldn't cause any failures, and I don't think it should add any significant delay but considering the entire testsuite that's good point, will revert this changes

Gauravtalreja1: Updated, thank you

mstrlc (approved):

PR #933, opened 2023-08-16
Handle no title case for search

sambible:
Adding a small clause here to allow pages that don't have a title attribute to continue using search. Only one I know of for sure that's this way, are the vm_import tests within computeresources, but there might be others. 

This seems like a no op for most pages, and allows others to continue to use search.


sambible:
gonna just put another update that I need to do for virt card in here as well, this will dynamically handle the fields on the card. Planning to probably rework this as part of the airgun improvement work soon, and refactor this stuff into a more general place, but I need to get this test back to passing in CI to move the epic to Release Pending

airgun/views/host_new.py
```diff
@@ -337,30 +337,28 @@ class networking_interface(Card):
...
+                    raise AttributeError(
+                        'Each label should have one value, therefore length should be equal. '
+                        f'Length of labels: {len(labels)} isnt equal to length of {len(values)}, '
+                        'Please double check xpaths.'
+                    )
+                for key, value in zip(labels, values):
+                    value = self.browser.text(value)
+                    key = self.browser.text(key).replace(' ', '_').lower()
+                    items[key] = value
+                return items
```
omkarkhatavkar, line 361: @sambible I see same here https://github.com/SatelliteQE/airgun/blob/857f8c1df2e848f40a5fb84fbcdb115bc650adf0/airgun/views/host_new.py#L64 please override rather adding similar code, One thing you can also do to check if this exists in other entity places if it exists push in widgets.py

airgun/views/common.py
```diff
@@ -440,7 +440,8 @@ def search(self, query):
         if not self.is_searchable():
             return None
         self.searchbox.search(query)
-        self.title.click()
+        if hasattr(self, 'title'):
+            self.title.click()
```
omkarkhatavkar, line 444: if you see this behaviour in many search widgets? if not I think it better to move and fix in a particular view only. This could fail other tests as this is a common location.

airgun/entities/host_new.py
```diff
@@ -480,6 +480,8 @@ def get_virtualization(self, entity_name):
         view = self.navigate_to(self, 'NewDetails', entity_name=entity_name)
         view.wait_displayed()
         self.browser.plugin.ensure_page_safe()
+        view.details.virtualization.read()
+        time.sleep(3)
```
omkarkhatavkar, line 484: I think this is unnecessary, can you please explain why you need hard sleep?

airgun/views/host_new.py
```diff
@@ -338,29 +338,7 @@ class networking_interface(Card):
...
-                './/div[contains(@class, "pf-c-description-list__group")][10]//dd'
-            )
-            overall_status = Text(
-                './/div[contains(@class, "pf-c-description-list__group")][11]//dd'
-            )
-            annotation_notes = Text(
-                './/div[contains(@class, "pf-c-description-list__group")][12]//dd'
-            )
-            running_on = Text('.//div[contains(@class, "pf-c-description-list__group")][13]//dd')
+            details = HostDetailsCard()
```
omkarkhatavkar, line 341: are you missing the code for this ?

airgun/entities/host_new.py
sambible: The trick is, the page only loads after calling view.details.virtualization.read(), and I need to wait at least 1 or 2 seconds for the card to load; it noticibly takes a few seconds before it pops up. So the workaround is to navigate there and try to read, wait, and then actually return a read

omkarkhatavkar: I understand you need to add the sync login to the read() method.  Override the read to add that sync logic using wait for

mstrlc (approved):
Looks good to me, now that `wait_for` is used.

vsedmik (approved):
Looks good to me, also the PRT for the related test passed (#3671).

PR #938, opened 2023-08-23
Introduce new project code standards

JacobCallahan:
This change adds Ruff as the new linter and loosens Black's formatting as well.
Due to this change, we can drop flake8 and old pre-commit hooks.

### DIFF
--- a/airgun/entities/virtwho_configure.py
+++ b/airgun/entities/virtwho_configure.py
@@ -1,13 +1,14 @@
 from navmazing import NavigateToSibling
 
 from airgun.entities.base import BaseEntity
-from airgun.navigation import NavigateStep
-from airgun.navigation import navigator
+from airgun.navigation import NavigateStep, navigator
 from airgun.utils import retry_navigation
-from airgun.views.virtwho_configure import VirtwhoConfigureCreateView
-from airgun.views.virtwho_configure import VirtwhoConfigureDetailsView
-from airgun.views.virtwho_configure import VirtwhoConfigureEditView
-from airgun.views.virtwho_configure import VirtwhoConfiguresView
+from airgun.views.virtwho_configure import (
+    VirtwhoConfigureCreateView,
+    VirtwhoConfigureDetailsView,
+    VirtwhoConfigureEditView,
+    VirtwhoConfiguresView,
+)
 
 
 class VirtwhoConfigureEntity(BaseEntity):
@@ -33,7 +34,7 @@ class VirtwhoConfigureEntity(BaseEntity):
         """Check if the config can be viewed/created"""
         try:
             view = self.navigate_to(self, 'All')
-        except Exception:
+        except Exception:  # noqa: BLE001 - TODO: determine proper exception
             return {"can_view": False, "can_create": False}
         return {"can_view": True, "can_create": view.new.is_displayed}
 
--- a/airgun/views/common.py
+++ b/airgun/views/common.py
@@ -1,36 +1,36 @@
-from widgetastic.widget import Checkbox
-from widgetastic.widget import ConditionalSwitchableView
-from widgetastic.widget import do_not_read_this_widget
-from widgetastic.widget import ParametrizedLocator
-from widgetastic.widget import ParametrizedView
-from widgetastic.widget import Text
-from widgetastic.widget import TextInput
-from widgetastic.widget import View
-from widgetastic.widget import WTMixin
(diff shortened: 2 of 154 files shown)


JacobCallahan:
The pre-commit check will fail on the below error until we have a determination on what the actual contents of the method should be.
```
airgun/entities/host_new.py:770:9: T201 `print` found
```

airgun/entities/virtwho_configure.py
```diff
@@ -33,7 +34,7 @@ def check_create_permission(self):
         """Check if the config can be viewed/created"""
         try:
             view = self.navigate_to(self, 'All')
-        except Exception:
+        except Exception:  # noqa: BLE001 - TODO: determine proper exception
```
JacobCallahan, line 37: Perhaps @yanpliu may know the correct exception class here.

airgun/views/common.py
```diff
@@ -285,7 +285,9 @@ def remove(self, value):
     def fill(self, values):
         """Remove associated resource(s)."""
         if not isinstance(values, list):
-            values = list((values,))
+            values = [
+                values,
+            ]
```
omkarkhatavkar, line 290: extra line not needed.

airgun/views/common.py
```diff
@@ -316,7 +318,9 @@ def add(self, value):
     def fill(self, values):
         """Associate resource(s)"""
         if not isinstance(values, list):
-            values = list((values,))
+            values = [
+                values,
+            ]
```
omkarkhatavkar, line 323: extra line

airgun/views/contenthost.py
```diff
@@ -1,35 +1,42 @@
...
+    EditableEntry,
+    EditableEntryCheckbox,
+    EditableEntryMultiCheckbox,
+    EditableEntrySelect,
+    Pagination,
+    ReadOnlyEntry,
+    Search,
+)
+
+BREADCRUMB_LENGTH = 2
```
omkarkhatavkar, line 39: this could be on the class attribute

airgun/views/hostcollection.py
```diff
@@ -185,7 +183,9 @@ def get_action_button(self, name):
         """Return an action button by it's name"""
         action_button = getattr(self, name)
         if not isinstance(action_button, ActionsDropdown):
-            raise ValueError(f'Action with name: "{name}" does not exists')
+            raise TypeError(
```
omkarkhatavkar, line 186: I think `ValueError` is more correct it seems, as TypeError would come if we are try to type something

airgun/views/product.py
```diff
@@ -38,7 +46,9 @@ class CreateDiscoveredReposView(View):
     def fill(self, values):
         """Select necessary repo/repos to be added to new or existing product"""
         if not isinstance(values, list):
-            values = list((values,))
+            values = [
```
omkarkhatavkar, line 49: extra line not needed

omkarkhatavkar (approved):

omkarkhatavkar:
trigger: test-robottelo
pytest: --component Puppet tests/foreman/ui

airgun/views/hostcollection.py
JacobCallahan: TypeError is more accurate since we're checking the type of the action_button as the error condition. See: https://beta.ruff.rs/docs/rules/type-check-without-type-error

airgun/views/contenthost.py
JacobCallahan: So this is something that I've been debating with myself. Let's discuss this more in chat

airgun/views/hostcollection.py
omkarkhatavkar: is that sense make sense

omkarkhatavkar:
trigger: test-robottelo
pytest: --component Puppet tests/foreman/ui

airgun/views/contenthost.py
JacobCallahan: ok, so I did ultimately decide to just make these class-level attributes instead of mixing module with class.

pyproject.toml
ogajduse, line 1: There are more rules that I would like to enable in our projects. Are you open to selecting extra rules on top of what you selected?

ogajduse (approved):

JacobCallahan:
trigger: test-robottelo
pytest: --component Puppet tests/foreman/ui

sambible (approved):

PR #964, opened 2023-09-11
adding the test name as metadata for searching in grid sessions

omkarkhatavkar:
image

### DIFF
--- a/airgun/browser.py
+++ b/airgun/browser.py
@@ -170,6 +170,10 @@ class SeleniumBrowserFactory:
 
         Note: should not be called directly, use :meth:`get_browser` instead.
         """
+        if self.test_name:
+            self.web_kaifuku.webdriver_options.desired_capabilities.update(
+                {'se:test_name': self.test_name}
+            )
         manager = BrowserManager.from_conf(self.web_kaifuku)
         self._webdriver = manager.start()
         self._set_session_cookie()


airgun/browser.py
```diff
@@ -170,6 +170,10 @@ def _get_remote_browser(self):
 
         Note: should not be called directly, use :meth:`get_browser` instead.
         """
+        if self.test_name:
+            self.web_kaifuku.get('webdriver_options').get('desired_capabilities').update(
+                {'se:test_name': self.test_name}
+            )
```
Gauravtalreja1, line 176: I think self.web_kaifuku should be an Box object, then can we try to update it something like?
```suggestion
            self.web_kaifuku.webdriver_options.desired_capabilities.update({'se:test_name': self.test_name})
```

Gauravtalreja1 (approved):
Ack, with a suggestion

airgun/browser.py
adarshdubey-star: It is a Box object I believe - https://github.com/SatelliteQE/airgun/blob/99d7b537479a3b20d0cd325fd17f9a1fa4c7b778/airgun/browser.py#L75

Gauravtalreja1 (approved):

adarshdubey-star (approved):
Ack

PR #966, opened 2023-09-11
add Host statues modal support

mstrlc:
Adds support for reading the `Manage host statuses` modal in the new Host page.
Calling `session.host_new.get_host_statuses(entity_name)` returns a dictionary of the values:

```
{
'Execution': {'Status': 'Last execution failed', 'Reported at': '8 minutes ago', 3: ''},
'Configuration': {'Status': 'N/A', 'Reported at': 'N/A', 3: ''}, 
'Build': {'Status': 'N/A', 'Reported at': 'N/A', 3: ''}, 
...
}
```

I will update https://github.com/SatelliteQE/robottelo/blob/master/tests/foreman/ui/test_rhcloud_insights.py to use this instead of the old page, then PRT can be ran, see here https://github.com/SatelliteQE/robottelo/pull/12582

### DIFF
--- a/airgun/views/host_new.py
+++ b/airgun/views/host_new.py
@@ -800,13 +800,13 @@ class ManageHostStatusesView(View):
 
     ROOT = './/div[@data-ouia-component-id="statuses-modal"]'
 
-    host_statuses_table = Table(
-        locator='.//table[contains(@class, "pf-c-table")]',
+    host_statuses_table = PatternflyTable(
+        component_id='statuses-table',
         column_widgets={
-            'name': Text('.//td[1]'),
-            'status': Text('.//td[2]'),
-            'reported_at': Text('.//td[3]'),
-            '3': Button(locator='.//td[4]'),
+            'Name': Text('.//td[1]'),
+            'Status': Text('.//td[2]'),
+            'Reported at': Text('.//td[3]'),
+            3: Button(locator='.//td[4]'),
         },
     )
 
--- a/airgun/views/host_new.py
+++ b/airgun/views/host_new.py
@@ -795,6 +795,26 @@ class EditSystemPurposeView(View):
     cancel = OUIAButton('cancel-syspurpose')
 
 
+class ManageHostStatusesView(View):
+    """Manage host statuses modal"""
+
+    ROOT = './/div[@data-ouia-component-id="statuses-modal"]'
+
+    host_statuses_table = Table(
+        locator='.//table[contains(@class, "pf-c-table")]',
+        column_widgets={
+            'name': Text('.//td[1]'),
+            'status': Text('.//td[2]'),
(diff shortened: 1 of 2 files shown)


airgun/views/host_new.py
```diff
@@ -795,6 +795,26 @@ class EditSystemPurposeView(View):
     cancel = OUIAButton('cancel-syspurpose')
 
 
+class ManageHostStatusesView(View):
+    """Manage host statuses modal"""
+
+    ROOT = './/div[@data-ouia-component-id="statuses-modal"]'
+
+    host_statuses_table = Table(
+        locator='.//table[contains(@class, "pf-c-table")]',
```
jameerpathan111, line 804: Is it possible to use ouia-id(`data-ouia-component-id="statuses-table"`) here as a locator? And https://github.com/RedHatQE/widgetastic.patternfly4/blob/master/src/widgetastic_patternfly4/ouia.py#L97C7-L97C22

airgun/views/host_new.py
```diff
@@ -795,6 +795,26 @@ class EditSystemPurposeView(View):
...
+
+    ROOT = './/div[@data-ouia-component-id="statuses-modal"]'
+
+    host_statuses_table = PatternflyTable(
+        component_id='statuses-table',
+        column_widgets={
+            'Name': Text('.//td[1]'),
+            'Status': Text('.//td[2]'),
+            'Reported at': Text('.//td[3]'),
+            3: Button(locator='.//td[4]'),
```
adarshdubey-star, line 809: I would suggest using something like this, instead of indexing it if possible :)
```suggestion
'Name': Text('.//td[contains(@data-label, "Name")]')
```

mstrlc: Should be possible, I just wanted to keep the code consistent through the file -- the same indexing is used in other views in the file, e.g. https://github.com/SatelliteQE/airgun/blob/master/airgun/views/host_new.py#L693C2-L701C15. Depends on what's more important

mstrlc:
https://github.com/SatelliteQE/robottelo/pull/12582/ Insights UI tests are updated to the new host page with the new modal. The PRT there should test this PR.

airgun/views/host_new.py
jameerpathan111: yeah, I agree. It's fine if it's not consistent with the rest of the code in the file.

mstrlc: Updated PR to address comment. Looks better now!

mstrlc:
Updated to fix the issue of the modal not closing after reading, thus blocking the UI.

mstrlc:
PRT in parent PR passes.

Griffin-Sullivan (approved):
LGTM

jameerpathan111 (approved):

PR #977, opened 2023-09-19
Enabling RedHat repositories enhanced

LadislavVasina1:
In this small PR, I have added an option to search for RedHat repository via custom query which was not available in the current implementation of the `enable` function.


vsedmik (approved):
LGTM

pondrejk (approved):

PR #979, opened 2023-09-21
Add PF4 versions of MultiSelect and FilteredDropdown. Use them in Filters.

lhellebr:


lhellebr:
```
$ pytest tests/foreman/ui/test_role.py::test_positive_end_to_end
[...]
=============================================== 1 passed in 373.15s (0:06:13) ================================================
```

airgun/widgets.py
```diff
@@ -1006,6 +1058,30 @@ def fill(self, value):
...
 
+class PF4FilteredDropdown(GenericLocatorWidget):
+    """Drop-down element with filtering functionality - PatternFly 4 version"""
+
+    filter_criteria = TextInput(locator=".//input[@aria-label='Select a resource type']")
+    filter_content = ItemsList(".//ul")
+
+    def clear(self):
+        """Clear currently selected value for drop-down"""
+        self.browser.clear(self.filter_criteria)
```
omkarkhatavkar, line 1069: There is no need for this method as `TextInput` has a fill method which clears the content.

lhellebr: There is, `PF4FilteredDropdown` is a subclass of `GenericLocatorWidget`. It needs to know it is supposed to call `clear` on that field. Otherwise, when I remove the method and call `clear` on the widget: `AttributeError: 'PF4FilteredDropdown' object has no attribute 'clear'`.

OTOH, you could say explicit `self.clear()` on line 1076 is not necessary. If I didn't remove it, I'd still get that field cleared. This way it is explicit + I have a feeling there was some corner case with empty string being filled and the field not cleared, but I'm not sure there.

In conclusion, I'd keep it as it is.

sambible (approved):
Looks good. ACK - always nice to see more older widgets get solid PF4 implementations

sambible:
@omkarkhatavkar lgtm!!

PR #991, opened 2023-09-28
No blank cv/french view 

sambible:
Airgun PR for a test automating a recent BZ. 

This contains all the publish changes, which should be merged soon. The portions here that should be given attention to will be marked up with a comment by me. Thanks :)


airgun/entities/contentview_new.py
```diff
@@ -23,25 +24,91 @@ def search(self, value):
...
 
-    VIEW = NewContentViewTableView
+    VIEW = ContentViewTableView
 
     @retry_navigation
     def step(self, *args, **kwargs):
         self.view.menu.select('Content', 'Lifecycle', 'Content Views')
 
 
+@navigator.register(NewContentViewEntity, 'French')
```
sambible, line 57: Here's the actual view introduced by this PR. This simply uses some different language to allow navigation to the CV page. ( Contenu instead of Content )

airgun/views/common.py
```diff
@@ -383,6 +386,71 @@ class add_tab(AddTab):
...
+class NewAddRemoveResourcesView(View):
+    searchbox = PF4Search()
+    type = Dropdown(
+        locator='.//div[contains(@class, "All repositories") or'
+        ' contains(@aria-haspopup="listbox")]'
+    )
+    Status = Dropdown(
+        locator='.//div[contains(@class, "All") or contains(@aria-haspopup="listbox")]'
+    )
+    add_repo = PF4Button('OUIA-Generated-Button-secondary-2')
```
omkarkhatavkar, line 398: looking at this code this is not the candidate here in the common.py. Please move this to specific view entity.

airgun/entities/contentview_new.py
```diff
@@ -23,25 +24,91 @@ def search(self, value):
...
+        view = self.navigate_to(self, 'Publish', entity_name=entity_name)
+        if values:
+            view.fill(values)
+        view.next.click()
+        view.finish.click()
+        view.progressbar.wait_for_result()
+        view = self.navigate_to(self, 'Edit', entity_name=entity_name)
+        return view.versions.table.read()
+
+    def check_if_blank_in_french(self):
```
omkarkhatavkar, line 42: I would rather name this differently `read_french_lang()`, here you are not checking this could be part of the test.

airgun/entities/contentview_new.py
```diff
@@ -23,25 +24,91 @@ def search(self, value):
...
 
+@navigator.register(NewContentViewEntity, 'French')
+class ShowAllContentViewsScreenFrench(NavigateStep):
+    """Navigate to All Content Views screen ( in French )"""
+
+    VIEW = ContentViewTableView
+
+    @retry_navigation
+    def step(self, *args, **kwargs):
+        self.view.menu.select('Contenu', 'Lifecycle', 'Content Views')
```
omkarkhatavkar, line 65: how will you change the language of the browser to French is that part you are handling in the test? or any configuration? @jyejare are you okay with adding such scenarios in Robottelo?

airgun/views/common.py
```diff
@@ -383,6 +386,71 @@ class add_tab(AddTab):
...
+        """Search for specific available resource and return the results"""
+        self.searchbox.search(value)
+        # Tried following ways to wait for table to be displayed, only sleep worked
+        # Might need a before/after fill
+        wait_for(
+            lambda: self.table.is_displayed is True,
+            timeout=60,
+            delay=1,
+        )
+        time.sleep(3)
```
omkarkhatavkar, line 423: avoid the hard sleep

airgun/entities/contentview_new.py
lhellebr: Yes. This is just a read in french without any assertion.

airgun/entities/contentview_new.py
lhellebr: This part handles navigation AFTER the language has already been changed to french. Nothing more, nothing less. Right, Sam?

airgun/views/contentview_new.py
```diff
@@ -1,10 +1,42 @@
...
+from airgun.views.common import NewAddRemoveResourcesView
+from airgun.views.common import SearchableViewMixinPF4
+from airgun.widgets import ActionsDropdown
+from airgun.widgets import ConfirmationDialog
+from airgun.widgets import EditableEntry
+from airgun.widgets import PF4Search
+from airgun.widgets import PF4ProgressBar
+from airgun.widgets import ReadOnlyEntry
+from airgun.views.common import BaseLoggedInView
+from airgun.views.common import NewAddRemoveResourcesView
```
lhellebr, line 27: Double import?

airgun/views/common.py
lhellebr: After all, it's only used in one view. Do you perhaps have more plans with it?

lhellebr, line 398: These "generated" locators may not be very stable, is there any alternative, like label?

airgun/views/common.py
```diff
@@ -383,6 +386,71 @@ class add_tab(AddTab):
...
+        locator='.//div[contains(@class, "All repositories") or'
+        ' contains(@aria-haspopup="listbox")]'
+    )
+    Status = Dropdown(
+        locator='.//div[contains(@class, "All") or contains(@aria-haspopup="listbox")]'
+    )
+    add_repo = PF4Button('OUIA-Generated-Button-secondary-2')
+    # Need to add kebab menu
+    table = PatternflyTable(
+        component_id='OUIA-Generated-Table-4',
```
lhellebr, line 401: The same, can you use some better locator?

airgun/views/common.py
lhellebr: That's what the comment on line 416 is about. He tried. Sam, what have you tried? Ensure_page_safe, wait_displayed?

airgun/views/contentview_new.py
```diff
@@ -56,3 +88,148 @@ def is_displayed(self):
...
         """Ensure 'Create content view' button is enabled after filling out the required fields"""
         self.submit.wait_displayed()
+
+
+class ContentViewEditView(BaseLoggedInView):
+    breadcrumb = BreadCrumb()
+    search = PF4Search()
+    title = Text("//h2[contains(., 'Publish) or contains(@id, 'pf-wizard-title-0')]")
+    actions = ActionsDropdown(
+        "//div[contains(@data-ouia-component-id, 'OUIA-Generated-Dropdown-2')]"
```
lhellebr, line 98: These 2 locators, can they be improved?

airgun/views/contentview_new.py
```diff
@@ -56,3 +88,148 @@ def is_displayed(self):
...
+    )
+    publish = PF4Button('cv-details-publish-button')
+    # not sure if this is needed
```
lhellebr, line 101: Can you try?

airgun/views/contentview_new.py
```diff
@@ -56,3 +88,148 @@ def is_displayed(self):
...
+            },
+        )
+        publishButton = PF4Button('cv-details-publish-button')
+
+        def search(self, version_name):
+            """Searches for content view version.
+            Searchbox can't search by version name, only by id, that's why in
+            case version name was passed, it's transformed into recognizable
+            value before filling, for example::
+                'Version 1.0' -> 'version = 1'
```
lhellebr, line 150: Is this really the correct mapping? Can't version 3.1 have id 16?

airgun/entities/contentview_new.py
sambible: ah yeah, that does make more sense. good suggestion!

airgun/entities/contentview_new.py
sambible: Correct. This essentially models navigating to cv_ in a french language satellite _

In the test, I create a user and change their language to french.

airgun/views/common.py
sambible: I can investigate. A lot of this stuff is from an older PR I salvaged to begin work on CV UI component eval, and I didn't clean it up properly.

airgun/entities/contentview_new.py
jyejare: @sambible @omkarkhatavkar  As per the discussion with people leads / Sudhir , we as a functional QE are not supposed to either test/automate globalization BZs. 

@sambible Any reason why we have automated the BZ ?

lhellebr: The BZ was about a whole empty page, right? That doesn't sound like a "globalization BZ".

airgun/entities/contentview_new.py
vijaysawant: @sambible Good to add doc string which would describe functionality

airgun/views/common.py
```diff
@@ -383,6 +386,71 @@ class add_tab(AddTab):
         )
 
 
+class NewAddRemoveResourcesView(View):
```
vijaysawant, line 389: Please add doc string which would help to understand added functionality.

airgun/views/contentview_new.py
```diff
@@ -56,3 +88,148 @@ def is_displayed(self):
     def after_fill(self, value):
         """Ensure 'Create content view' button is enabled after filling out the required fields"""
         self.submit.wait_displayed()
+
+
+class ContentViewEditView(BaseLoggedInView):
```
vijaysawant, line 93: Please add doc string for a class which would help to understand added functionality.

airgun/views/contentview_new.py
```diff
@@ -56,3 +88,148 @@ def is_displayed(self):
...
+        name = EditableEntry(name='Name')
+        label = ReadOnlyEntry(name='Label')
+        type = ReadOnlyEntry(name='Composite?')
+        description = EditableEntry(name='Description')
+        # depSolv is maybe a conditionalswitch
+        solve_dependencies = Switch(name='solve_dependencies switch')
+        import_only = Switch(name='import_only_switch')
+
+    @View.nested
+    class versions(Tab):
```
vijaysawant, line 127: Can we add short doc string here?

airgun/views/contentview_new.py
```diff
@@ -56,3 +88,148 @@ def is_displayed(self):
...
+        self.promote.click()
+        wait_for(
+            lambda: self.lce.is_displayed is True,
+            timeout=30,
+            delay=1,
+            logger=self.logger,
+        )
+
+
+class NewContentViewVersionDetailsView(BaseLoggedInView):
```
vijaysawant, line 223: Please add doc string for a class which would help to understand added functionality.

airgun/views/contentview_new.py
```diff
@@ -1,10 +1,42 @@
...
+from airgun.widgets import ReadOnlyEntry
+from airgun.views.common import BaseLoggedInView
+from airgun.views.common import NewAddRemoveResourcesView
+from airgun.views.common import SearchableViewMixinPF4
+from airgun.widgets import ActionsDropdown
+from airgun.widgets import ConfirmationDialog
+from airgun.widgets import EditableEntry
+from airgun.widgets import PF4Search
+from airgun.widgets import PF4ProgressBar
+from airgun.widgets import ReadOnlyEntry
```
vijaysawant, line 34: Looks like import twice, need to remove

vijaysawant (changes requested):

airgun/entities/contentview_new.py
sambible: IT's not a globalization BZ, it's testing a specific customer Issue, where when we display certain characters in the UI, the pages will fail to display. This could happen in non-localization scenarios.

airgun/views/contentview_new.py
sambible: This is just talking about mapping "Version 22" to search term '22", since you can't actually search "Version 22"

sambible: Updated language

airgun/entities/contentview_new.py
jyejare: @sambible @omkarkhatavkar Not sure whether such bugs comes into the category of Localization but I just wanted to ensure that we dont divert airgun to adapt something new for such a localization-specific case. Airgun is a pure functional testing tool. 

Also I really dont think we should cover such cases in regression testing thats less likely to reproduce in the future to catch the `regression` of it!

Gauravtalreja1:
@sambible Could you update this PR?

Gauravtalreja1:
@sambible Did you get a chance to revisit this? If not, can it be closed if this PR isn't worked on?

Issue #989, opened 2023-09-26
Date dependent path added for selenium screenshots


dosas:
Currently for [selenium screenshots](https://github.com/SatelliteQE/airgun/blob/master/airgun/session.py#L290) a folder consisting of the current date is created, IMHO this is should be removed for the following reasons:

* The screenshot path is provided by the setting `settings.selenium.screenshots_path`, adding a folder in the code makes it impossible to reference this setting elsewhere (CI, pytest, ...)
* Everything that is date dependent makes problems in a CI system: Imagine running the tests around midnight then you get two folders for one testrun
* The date is already added in the screenshot filename, so this information is duplicated

Further more a setting to not have the datetime in the screenshot file name could be added.

PR #1002, opened 2023-10-11
Remove date dependent screenshot path

dosas:
* Do not implicitly overwrite `settings.selenium.screenshots_path`
* Date is already set in the filename
* Reduce datetime dependency

See also https://github.com/SatelliteQE/airgun/issues/989

### DIFF
--- a/airgun/session.py
+++ b/airgun/session.py
@@ -301,10 +301,8 @@ class Session:
         session happens.
         """
         now = datetime.now()
-        path = os.path.join(
-            settings.selenium.screenshots_path,
-            now.strftime('%Y-%m-%d'),
-        )
+        path = settings.selenium.screenshots_path
+
         if not os.path.exists(path):
             os.makedirs(path)
         path = os.path.join(


omkarkhatavkar:

sambible (approved):
Seems reasonable to me. No real reason to have the date both places, I'd agree.

Gauravtalreja1 (approved):

dosas:
Are there any actions left required from me?

ogajduse:
> Are there any actions left required from me?
I can not see any. Let me run our PR tester, that should show us if something broke on our end.

ogajduse:
trigger: test-robottelo
pytest: tests/foreman/ui/test_architecture.py tests/foreman/ui/test_bookmarks.py tests/foreman/ui/test_hardwaremodel.py tests/foreman/ui/test_host.py::test_positive_read_from_details_page

ogajduse:
PRT did not trigger a job build in our internal infra. @omkarkhatavkar Can you please have a look at what might be wrong?

airgun/session.py
```diff
@@ -301,10 +301,8 @@ def take_screenshot(self):
         session happens.
         """
         now = datetime.now()
-        path = os.path.join(
-            settings.selenium.screenshots_path,
-            now.strftime('%Y-%m-%d'),
-        )
+        path = settings.selenium.screenshots_path
+
```
omkarkhatavkar, line 305: @dosas Thanks for creating the PR, I appreciate the work you added. Looking at the function doc string this was added long back, looking at this statement

 `All directories will be created if they don't exist. Make sure that the user running Robottelo has the right permissions to create files and directories.` this could be the reason this was added. I see some value in that. 

- It dynamically creates directories based on the current date, ensuring that the screenshots or files are organized in a structured manner. Imagine if someone specifies the '/tmp' path here; all the screenshots would be jumbled together in a single directory. It's much better to have them organized by date for better isolation.
- 

@jyejare, @ogajduse what are your thoughts?

dosas: There is an easy fix for that don't use `/tmp` but use `/tmp/screenshots`.
Don't take away all responsibility from the user.

omkarkhatavkar: > I would even go as far and not put the date information into the filename, since this is duplicate information that is already provided by the filesystem.

I have a different perspective on this. If we don't include the date and time in the screenshot filenames using (DateTime) f'{self.name}-screenshot-YYYY-mm-dd_HH_MM_SS.png', there's a risk of overwriting the same screenshots when running the test again.

Don't fix something that isn't broken.

dosas: Good point I did not think of that

omkarkhatavkar (changes requested):
I have reviewed the proposed changes in this pull request in light of the current requirements within Airgun and Robottelo, as indicated in my previous review comments https://github.com/SatelliteQE/airgun/pull/1002/files#r1371209526. After careful consideration, I find that these changes do not align with the current project requirements and standards.

Therefore, I regret to inform you that I am disapproving of this PR at this time. If you have any questions or require further clarification regarding the decision, please don't hesitate to reach out, and we can discuss how to best proceed.

PR #1013, opened 2023-10-23
Added possibility to search for debian errata

dosas:
in the form DSA-1234-1

### DIFF
--- a/airgun/views/errata.py
+++ b/airgun/views/errata.py
@@ -1,8 +1,7 @@
-import re
-
 from widgetastic.widget import Checkbox, Text, View
 from widgetastic_patternfly import BreadCrumb
 
+from airgun import ERRATA_REGEXP
 from airgun.views.common import BaseLoggedInView, SatTab, TaskDetailsView
 from airgun.widgets import ItemsList, ReadOnlyEntry, SatSelect, SatTable, Search
 
@@ -40,7 +39,7 @@ class ErratumView(BaseLoggedInView):
         if repo is not None:
             self.repo_filter.fill(repo)
 
-        if re.search(r'\w{3,4}[:-]\d{4}[-:]\d{4}', query):
+        if ERRATA_REGEXP.search(query):
             query = f'id = {query}'
         self.searchbox.search(query)
 
--- a/airgun/__init__.py
+++ b/airgun/__init__.py
@@ -1,3 +1,7 @@
+import re
+
 from airgun.settings import Settings
 
 settings = Settings()
+
+ERRATA_REGEXP = re.compile(r'\w{3,4}[:-]\d{4}[-:]\d{1,4}')
--- a/airgun/entities/errata.py
+++ b/airgun/entities/errata.py
@@ -1,5 +1,4 @@
-import re
-
+from airgun import ERRATA_REGEXP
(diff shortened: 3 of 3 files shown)


omkarkhatavkar:

airgun/views/errata.py
```diff
@@ -40,7 +40,7 @@ def search(self, query, applicable=True, installable=False, repo=None):
         if repo is not None:
             self.repo_filter.fill(repo)
 
-        if re.search(r'\w{3,4}[:-]\d{4}[-:]\d{4}', query):
+        if re.search(r'\w{3,4}[:-]\d{4}[-:]\d{1,4}', query):
```
omkarkhatavkar, line 43: @sambible  @vsedmik @vijaysawant  Do you see any value for this regex? As per I know we never use the Debian errata, mostly test RHEL errata.

sambible: We don't test any Debian errata that I'm aware of. This will change this particular regex to allow for 
DSA-1234-1 where in the past it would not have matched, since RHEL errata are in the form
RHSA-1234-1234 ( With explicitly 4 characters in the last third) 

These will still get captured by this adjusted regex, although it won't match them as strictly.  If there's value to @dosas here, I'd be curious to know the use case

dosas: My use case is to read and install debian errata with robottelo which currently is not possible since the [search](https://github.com/SatelliteQE/airgun/blob/master/airgun/entities/errata.py#L138) is a requirement for the ErrataDetailsView

dosas: Upon writing the above comment I realized that the same regexp is also used for this [view ](https://github.com/SatelliteQE/airgun/blob/master/airgun/entities/errata.py#L140).

So if there is consens on accepting this PR I would maybe put this information in one place?

JacobCallahan:
@omkarkhatavkar I don't think PRT will help us here, aside from basic regression testing, since debian isn't supported by Satellite. This would need to be tested in an upstream build.

omkarkhatavkar:
@JacobCallahan As you said rightly, running PRT to ensure regression is passing and we will have to ask the anyone from the Phonix team to test this with an upstream build. I guess @sambible @vsedmik @ColeHiggins2 could help.

dosas:
any news on this?

vsedmik:
trigger: test-robottelo
pytest: tests/foreman/ui/test_errata.py -k test_end_to_end

dosas:
I see that some tests are failing. Let me know if this is related to the change and I should fix something.

dosas:
@vsedmik I saw that you removed cherry pick labels for 6.14 and 6.13 is there any strong reason for this. Would it be possible to cherry pick it into those branches?

vsedmik:
@dosas oh no, it was a mid-air collision. I just added `6.15` and Sam added `6.15, 6.14, 6.13` at the same time, which rendered as me removing the diff :(

dosas:
@vsedmik wjhat would we have to do to get these changes to 6.13.z and 6.14.z

vsedmik:
@dosas I've opened manual cherry-pick PRs with your changes and triggered the PRT.
@sambible please review and merge if all ok.

PR #1039, opened 2023-11-16
New All Hosts Page UI work

sambible:
First entities and views for the new All Hosts page. Everything in there is modeled currently, and this will be an ongoing effort as dev continues to flesh out this new page. 

Note: This is in a new, seperate file from the current All Hosts page, as both of these pages will exist for some time on Satellite. In the future, effort will need to be made to bridge these two, as this becomes the default page.


damoore044 (approved):
Looks good!

airgun/entities/all_hosts.py
```diff
@@ -0,0 +1,62 @@
...
+
+@navigator.register(AllHostsEntity, 'All')
+class ShowAllHostsScreen(NavigateStep):
+    """Navigate to All Hosts screen."""
+
+    VIEW = AllHostsTableView
+
+    @retry_navigation
+    def step(self, *args, **kwargs):
+        self.view.menu.select('Hosts', 'All Hosts')
```
omkarkhatavkar, line 62: we already have an entry for this using https://github.com/SatelliteQE/airgun/blob/92926768d3dfeb16bbb8bcc3c61ca5fce111bf0c/airgun/entities/host.py#L311 how this is a different entity then?

airgun/entities/all_hosts.py
```diff
@@ -0,0 +1,62 @@
...
+        view.table[0][2].widget.item_select('Delete')
+        delete_modal = HostDeleteDialog(self.browser)
+        if delete_modal.is_displayed:
+            delete_modal.confirm_delete.click()
+        view = self.navigate_to(self, 'All')
+        self.browser.plugin.ensure_page_safe(timeout='5s')
+        view.wait_displayed()
+        view.search(host_name)
+        return view.no_results
+
```
omkarkhatavkar, line 36: This extra step of searching for host deleting after is unnecessary.

airgun/entities/all_hosts.py
```diff
@@ -0,0 +1,62 @@
...
+    def delete(self, host_name):
+        """Delete host through table dropdown"""
+        view = self.navigate_to(self, 'All')
+        self.browser.plugin.ensure_page_safe(timeout='5s')
+        view.wait_displayed()
+        view.search(host_name)
+        view.table[0][2].widget.item_select('Delete')
+        delete_modal = HostDeleteDialog(self.browser)
+        if delete_modal.is_displayed:
+            delete_modal.confirm_delete.click()
```
omkarkhatavkar, line 30: I think an exception can be raised in case the confimration_delete modal can not be displayed.

sambible: I can do that

airgun/entities/all_hosts.py
adarshdubey-star: @sambible Anyway we do a teardown/assertion at the test level for deletion so I believe @omkarkhatavkar is correct here.

sambible: This is just to confirm that the host was actually deleted, and not errored out by satellite.

adarshdubey-star (approved):
Ack

adarshdubey-star:
@sambible can you resolve the conflicts on the exercising robottelo PR.

PR #1049, opened 2023-11-28
eol banner

pondrejk:


omkarkhatavkar (approved):

airgun/entities/eol_banner.py
```diff
@@ -0,0 +1,32 @@
...
+    def is_danger(self):
+        view = self.navigate_to(self, 'NavigateToEOLBanner')
+        return view.danger
+
+
+@navigator.register(EOLBannerEntity)
+class NavigateToEOLBanner(NavigateStep):
+    VIEW = EOLBannerView
+
+    def step(self, *args, **kwargs):
```
omkarkhatavkar, line 28: Is this information intended for display on the dashboard? I couldn't locate any actionable steps or a URL path to navigate to the corresponding entity. If this is the case, how do we define this entity? Shouldn't this be handled within the dashboard page entity itself?

pondrejk: Hm, it doesn't feel right to me, because it would seem that the banner is something tied to the dashboard, but it really is a thing of its own. It's just not big enough to have a component of its own, therefore it is tracked under dashboard component. Hope this makes sense

lhellebr:
This is cherrypicked. Why is the robottelo PR not?

PR #1050, opened 2023-11-30
Added Satellite Capsule page coverage

LadislavVasina1:
In this PR I have added views and entities for Satellite page coverage.


pondrejk (approved):
one non-blocking comment, otherwise nice work

PR #1059, opened 2023-12-05
Add check to verify supported provisioning templates

Gauravtalreja1:
**Description:**
1. Add check to verify supported provisioning templates with check to verify presence of RH logo.
2. Update existing is_locked method to handle if template is not locked


omkarkhatavkar (approved):
ACK

shubhamsg199 (approved):

PR #1069, opened 2023-12-07
Remove LCE field from HostRegisterView

Gauravtalreja1:
Related Katello PR https://github.com/Katello/katello/pull/10764


sambible (approved):
Do we have any tests that are using those locators? If not, this seems fine. If so, we need to cleanup those areas as well. ACK if not.

Gauravtalreja1:
@sambible yes, we have test in robottelo (PR is linked), updating the test as well by removing this field

omkarkhatavkar (approved):

PR #1087, opened 2023-12-18
Update get_register_command to reflect new mandatory fields

Gauravtalreja1:
**Description**
- Update get_register_command to reflect new mandatory fields i.e ActivationKey, which was previously operating systems.
- Make get_register_command work when no values are passed, required for default values validation and few negative tests.
- Add locator for new_activation_key_link, which appears when no AK available in organization, and when there's single AK in organization, AK will be auto-selected in 6.15+

### DIFF
--- a/airgun/entities/host.py
+++ b/airgun/entities/host.py
@@ -46,13 +46,19 @@ class HostEntity(BaseEntity):
         host_view.flash.assert_no_error()
         host_view.flash.dismiss()
 
-    def get_register_command(self, values, full_read=None):
+    def get_register_command(self, values=None, full_read=None):
         """Get curl command generated on Register Host page"""
         view = self.navigate_to(self, 'Register')
-        view.fill(values)
-        self.browser.click(view.generate_command)
-        self.browser.plugin.ensure_page_safe()
-        view.registration_command.wait_displayed()
+        if values is not None:
+            view.fill(values)
+        if view.general.activation_keys.read():
+            self.browser.click(view.generate_command)
+            self.browser.plugin.ensure_page_safe()
+            view.registration_command.wait_displayed()
+        else:
+            view.general.new_activation_key_link.wait_displayed()
+            if view.generate_command.disabled:
+                raise DisabledWidgetError('Generate registration command button is disabled')
         if full_read:
             return view.read()
         return view.registration_command.read()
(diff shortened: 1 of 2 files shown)


airgun/entities/host.py
```diff
@@ -46,13 +46,19 @@ def create(self, values):
...
 
-    def get_register_command(self, values, full_read=None):
+    def get_register_command(self, values=None, full_read=None):
         """Get curl command generated on Register Host page"""
         view = self.navigate_to(self, 'Register')
-        view.fill(values)
-        self.browser.click(view.generate_command)
-        self.browser.plugin.ensure_page_safe()
-        view.registration_command.wait_displayed()
+        if values is not None:
```
omkarkhatavkar, line 52:
```suggestion
        if not values
```

airgun/entities/host.py
```diff
@@ -46,13 +46,19 @@ def create(self, values):
...
+            view.fill(values)
+            if view.general.activation_keys:
```
omkarkhatavkar, line 54: There is no significance of this check if you are not planning to raise anything.

omkarkhatavkar (approved):
ACK, 1 minor change

LadislavVasina1 (approved):

airgun/entities/host.py
Gauravtalreja1: `not None => True` will be wrong condition here, we need to fill when there are values

omkarkhatavkar: I mean `if values`

airgun/entities/host.py
```diff
@@ -46,13 +46,19 @@ def create(self, values):
...
+    def get_register_command(self, values=None, full_read=None):
         """Get curl command generated on Register Host page"""
         view = self.navigate_to(self, 'Register')
-        view.fill(values)
-        self.browser.click(view.generate_command)
-        self.browser.plugin.ensure_page_safe()
-        view.registration_command.wait_displayed()
+        if values is not None:
+            view.fill(values)
+        if view.general.activation_keys.read():
```
omkarkhatavkar, line 54: have you tested this when there is no data? I mean if nothing was selected. It should not show exception or getting empty space string

adarshdubey-star (approved):

airgun/entities/host.py
adarshdubey-star: Yeah worked perfectly with my PR

adarshdubey-star:
Good work @Gauravtalreja1

PR #1152, opened 2024-01-12
host_new>content>errata tab pagination

damoore044:
### Purpose
- supports [robottelo #13767](https://github.com/SatelliteQE/robottelo/pull/13767)
- Add new host UI support for reading errata tab pagination.
- method could be extended to any content>{tab with pagination} in New Host UI page once working.

### Solution
-Added new method `clear()` to `SearchInput` class (aigun/widgets.py)
-usage in the airgun view like so`view.content.errata.searchbar.clear()`
-Now using this in `apply_erratas()` method (entities/host_new.py), to clear the searchbar of any possible value present, when no search passed to` apply_erratas()`, meaning show and apply all errata with a blank searchbar.
-We also `clear` the searchbar in `apply_erratas()` after installing, that way any search value passed does not persist in the searchbar.
### ~Issue~, Resolved
- ~When reading info from PF4Pagination object in [views/host_new.py@362](https://github.com/SatelliteQE/airgun/blob/06ea1937089659ad1e27007f73b18cd297c6bf88/airgun/views/host_new.py#L362C7-L362C7), we get an empty dictionary with no information. _**Solved**_~
- ~**When the method `apply_erratas(entity, search)` is run,** it will leave the searchbar filled, and keep it filled until the view is reset.~
-  ~Workaround: Before accessing `view.content.errata.pagination`, we pass ' ' to searchbar, to display all available errata unfiltered. In case the searchbar was left occupied by a method run previously. I have been looking for a better way or built in method to clear the searchbar /unfilter, before geting pagination.~

### DIFF
--- a/airgun/entities/host_new.py
+++ b/airgun/entities/host_new.py
@@ -299,17 +299,41 @@ class NewHostEntity(HostEntity):
         view.content.errata.table.wait_displayed()
         return view.read(widget_names="content.errata.table")
 
-    def apply_erratas(self, entity_name, search):
-        """Apply errata on selected host based on errata_id"""
+    def apply_erratas(self, entity_name, search=None):
+        """Apply available errata on selected host based on searchbar result.
+
+        param: search (string): search value to filter results.
+        example: search="errata_id == {ERRATA_ID}"
+        default: None; all available errata are returned and installed.
+        """
         view = self.navigate_to(self, 'NewDetails', entity_name=entity_name)
         view.wait_displayed()
-        view.content.errata.searchbar.fill(search)
+        if search is None:
+            # Return all errata, clear the searchbar
+            view.content.errata.searchbar.clear()
+        else:
+            # Return only errata by search
+            view.content.errata.searchbar.fill(search)
         # wait for filter to apply
+        view.content.errata.wait_displayed()
         self.browser.plugin.ensure_page_safe()
         view.content.errata.select_all.click()
         view.content.errata.apply.fill('Apply')
         view.flash.assert_no_error()
         view.flash.dismiss()
+        # clear the searchbar so any input will not persist
+        view.content.errata.searchbar.clear()
+        view.content.errata.wait_displayed()
+        self.browser.plugin.ensure_page_safe()
+
+    def get_errata_pagination(self, entity_name):
+        """Get pagination info from Errata tab on selected host."""
+        view = self.navigate_to(self, 'NewDetails', entity_name=entity_name)
+        view.wait_displayed()
+        view.content.errata.select()
+        view.content.errata.wait_displayed()
+        self.browser.plugin.ensure_page_safe()
+        return view.content.errata.pagination
 
     def get_module_streams(self, entity_name, search):
         """Filter module streams"""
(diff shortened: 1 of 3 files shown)


airgun/entities/host_new.py
```diff
@@ -311,6 +311,14 @@ def apply_erratas(self, entity_name, search):
         view.flash.assert_no_error()
         view.flash.dismiss()
 
+    def get_errata_pagination(self, entity_name):
+        """Get pagination info from Errata tab on selected host"""
+        view = self.navigate_to(self, 'NewDetails', entity_name=entity_name)
+        view.content.errata.select()
+        view.wait_displayed()
+        view.content.errata.wait_displayed()
+        return view.read(widget_names="content.errata.pagination")
```
omkarkhatavkar, line 320: I think the further issue is related to making a right locators for pagination. at this location https://github.com/SatelliteQE/airgun/blob/593616d3f3cbe06cd88aaa1f433c2aff0802f1b5/airgun/views/host_new.py#L362. I had chat with @vijaysawant which know the details about it.

damoore044: Yes he follow up with the new locator workaround i've committed here. 
Since there are two pagination objects on `newhost>content>errata` tab, 
we only read the top one, correct? Seeing success with this locally

airgun/entities/host_new.py
```diff
@@ -311,6 +311,18 @@ def apply_erratas(self, entity_name, search):
...
         view.flash.dismiss()
 
+    def get_errata_pagination(self, entity_name):
+        """Get pagination info from Errata tab on selected host"""
+        view = self.navigate_to(self, 'NewDetails', entity_name=entity_name)
+        view.wait_displayed()
+        view.content.errata.select()
+        # pass empty search to show any and all,
+        # if a previous method left the searchbar filled
+        view.content.errata.searchbar.fill(' ')
```
omkarkhatavkar, line 321: Please correct this. We won't be applying the `'  '` string

sambible (approved):
Pending @omkarkhatavkar comments

airgun/entities/host_new.py
damoore044: Yes, we made a new method to `clear` the SearchInput searchbar (widgets.py), which was unintentionally left occupied by `search` value passed to `apply_erratas` method (host_new.py).
Now, we clear the searchbar after selecting and applying erratas by search, so that the passed `search` does not persist.

omkarkhatavkar:
@damoore044 looks good can you take a look at pre-commit failures?

airgun/widgets.py
```diff
@@ -2645,7 +2645,25 @@ def read(self):
 
 
 class SearchInput(TextInput):
+    search_field = TextInput(locator=('//input[@aria-label="Search input"]'))
```
vsedmik, line 2648: Could you add some docstring here? I wonder if the docs build fails coz of this.

vsedmik (approved):
Looks good, just fix the CI/CQ

ColeHiggins2 (approved):

vsedmik (approved):

PR #1231, opened 2024-02-08
Fix hardware model delete functionality (6.15.z)

dosas:
For foreman 3.9 (6.15.z) the 'Actions' header string was removed (see picture):

hardware_models

The delete action is now accessible via an svg.

This should fix https://github.com/SatelliteQE/robottelo/blob/master/tests/foreman/ui/test_hardwaremodel.py

### DIFF
--- a/airgun/views/hardware_model.py
+++ b/airgun/views/hardware_model.py
@@ -1,11 +1,12 @@
 from widgetastic.widget import Text, TextInput
-from widgetastic_patternfly import BreadCrumb, Button
+from widgetastic_patternfly import BreadCrumb
+from widgetastic_patternfly4 import Dropdown
 
 from airgun.views.common import BaseLoggedInView, SearchableViewMixinPF4
-from airgun.widgets import ConfirmationDialog, SatTable
+from airgun.widgets import Pf4ConfirmationDialog, SatTable
 
 
-class DeleteHardwareModelDialog(ConfirmationDialog):
+class DeleteHardwareModelDialog(Pf4ConfirmationDialog):
     confirm_dialog = Text(".//button[contains(normalize-space(.),'Delete')]")
     cancel_dialog = Text(".//button[normalize-space(.)='Cancel']")
 
@@ -18,7 +19,7 @@ class HardwareModelsView(BaseLoggedInView, SearchableViewMixinPF4):
         './/table',
         column_widgets={
             'Name': Text('.//a'),
-            'Actions': Button('Delete'),
+            4: Dropdown(),
         },
     )
 
--- a/airgun/entities/hardware_model.py
+++ b/airgun/entities/hardware_model.py
@@ -46,7 +46,7 @@ class HardwareModelEntity(BaseEntity):
         """
         view = self.navigate_to(self, 'All')
         view.search(entity_name)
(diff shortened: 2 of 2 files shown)


airgun/views/hardware_model.py
```diff
@@ -18,7 +18,7 @@ class HardwareModelsView(BaseLoggedInView, SearchableViewMixinPF4):
         './/table',
         column_widgets={
             'Name': Text('.//a'),
-            'Actions': Button('Delete'),
+            4: Pf4SvgActionsDropdown(".//div[contains(@class, 'pf-c-dropdown')]"),
```
omkarkhatavkar, line 21: Can you please try out this widget rather writing new widget https://github.com/SatelliteQE/airgun/blob/master/airgun/views/acs.py#L362

airgun/widgets.py
```diff
@@ -743,6 +743,35 @@ def select(self, item):
...
+          </ul>
+        </div>
+
+    """
+
+    dropdown = Text('.//button[contains(@class,"pf-c-dropdown__toggle")]')
+    ITEM_LOCATOR = ".//ul/li/button[@role='menuitem' and contains(normalize-space(.), '{}')]"
+
+    def fill(self, item):
+        raise NotImplementedError
```
omkarkhatavkar, line 772: for now I dont feel it might not be needed to write new widget, can you please try this https://github.com/SatelliteQE/airgun/blob/master/airgun/views/acs.py#L362 ?

airgun/views/hardware_model.py
dosas: Thank you for the hint. This was the abstraction I was looking for but couldn't not find.

omkarkhatavkar (approved):

omkarkhatavkar:
@lhellebr @jameerpathan111 @LadislavVasina1 @pondrejk Can anyone from your team run these PRT tests for this PR?

lhellebr:
Have you verified this change is intended?
Where is this used, what tests should we run PRT for?

lhellebr:
trigger: test-robottelo
pytest: tests/foreman/ui/test_hardwaremodel.py::test_positive_end_to_end

dosas:
The mentioned test will still fail since there is a follow up error:
```
AssertionError: assert_message: failed to find matching notifications. Available notifications: []
```

The error fixed within this PR is about not being able to open the delete pop-up

omkarkhatavkar:
@dosas Thanks for giving the fix for this. I verified the test. It seems there there is failure in case of deleting the already used model with wrong error message. While I was still able to record the test execution see that delete it working okay.

omkarkhatavkar (approved):

omkarkhatavkar:
@dosas will wait for any other opinion, else we will merge this PR.

lhellebr:
The test fails with error `Request failed with status code 422` because the hardware model is already in use. Actually, a different error is expected:
```
        if err_message:
            view.flash.assert_message(f"Danger alert: {err_message}")
```
That's an issue unrelated to the PR that should be fixed:
1) A correct error message should be expected
2) The message should say what the problem was, not just `422`

lhellebr (approved):

dosas:
Despite the label this was never cherry-picked into 6.15.z. Does somebody know what went wrong?

PR #1281, opened 2024-03-12
Check that breadcrumb is displayed

dosas:
error occurred in 6.15.z

### DIFF
--- a/airgun/entities/contenthost.py
+++ b/airgun/entities/contenthost.py
@@ -39,6 +39,7 @@ class ContentHostEntity(BaseEntity):
     def read(self, entity_name, widget_names=None):
         """Read content host details, optionally read only the widgets in widget_names."""
         view = self.navigate_to(self, 'Edit', entity_name=entity_name)
+        view.is_displayed
         return view.read(widget_names=widget_names)
 
     def read_legacy_ui(self, entity_name, widget_names=None):


jyejare (approved):

LadislavVasina1:
@jyejare @shweta83 Hi, the change that @dosas did does totally nothing in my opinion.
 `view.is_displayed` returns just bool and the value is not even checked afterward.
Correct me if I am wrong about this, but I don't understand the ACKS on this PR.

jyejare:
@LadislavVasina1 Here he is waiting for the view to be displayed which was not happening earlier, though he is not checking the return value, he is just adding a dynamic wait before proceeding further.

LadislavVasina1:
Ok, but I still think that using just `self.browser.wait_for_element(self.breadcrumb, exception=False)` is still more readable.

LadislavVasina1:
Sure it works too, but you are using `ContentHostDetailsView` not `ContentHostsView`. This view is loaded by navigating to 'Edit' https://github.com/SatelliteQE/airgun/blob/7995b52325eac8696bb49d2618458f82839ca88e/airgun/entities/contenthost.py#L228-L236.
This means that `is_displayed` is implemented like this https://github.com/SatelliteQE/airgun/blob/7995b52325eac8696bb49d2618458f82839ca88e/airgun/views/contenthost.py#L122-L134

dosas:
> Sure it works too, but you are using `ContentHostDetailsView` not `ContentHostsView`. This view is loaded by navigating to 'Edit'
> 
> https://github.com/SatelliteQE/airgun/blob/7995b52325eac8696bb49d2618458f82839ca88e/airgun/entities/contenthost.py#L228-L236
> .
> This means that `is_displayed` is implemented like this
> 
> https://github.com/SatelliteQE/airgun/blob/7995b52325eac8696bb49d2618458f82839ca88e/airgun/views/contenthost.py#L122-L134

Yes, exactly even better. I get everything from the abstraction for free.

LadislavVasina1:
What I primarily wanted to say is that `is_displayed` here returns bool, which you are not using. That is all. That's why I suggested just putting `wait_for_element` by yourself there.

dosas:
But if the issue is with the return value check shouldn't the fix be:
```
if not view.is_displayed():
   raise RuntimeError(:Element not displayed")
```
?

LadislavVasina1:
yes, but I think that `self.browser.wait_for_element(self.breadcrumb)` should be sufficient, that is why I suggested that.

omkarkhatavkar:
@dosas, I don't believe that adding the view.is_displayed() within the read method is the optimal approach. Since it pertains to the view, it would be more appropriate to place it within the navigation steps, particularly post_navigate.

- The rationale behind this decision is to avoid cluttering the read function with a specific view check. This way, if the same view is utilized in other functions, we won't need to duplicate the code in each location, which could lead to maintenance issues.
- I concur with @LadislavVasina1's observation that calling view.is_displayed() may not be the most logical choice. Instead, employing a proper wait_for mechanism would ensure that the script waits until the view is displayed, as I understand you intend to verify the presence of the page.

For reference, you can find a similar example below: 
https://github.com/SatelliteQE/airgun/blob/5b7aa9d30c9ff0bdb5dd02d2b8c280b799398f91/airgun/entities/modulestream.py#L65

dosas:
@omkarkhatavkar Thank you for the hint. 

With this additional information I agree that adding `view.is_displayed()` where it is currently added is not a good solution.
So I will move it (as you suggested) to the `EditContentHost` class.

I disagree with the statement that `is_displayed()` is not a 'proper' wait mechanism. `is_displayed()` calls [self.browser.wait_for_element](https://github.com/SatelliteQE/airgun/blob/master/airgun/views/contenthost.py#L118) which again calls [wait_for](https://github.com/RedHatQE/widgetastic.core/blob/master/src/widgetastic/browser.py#L432)

So in your example `wait_for` calls `is_displayed` which again calls `wait_for` so the first call of wait_for is absolutely not required.

omkarkhatavkar:
> @omkarkhatavkar Thank you for the hint.

**Appreciated** 

> With this additional information I agree that adding `view.is_displayed()` where it is currently added is not a good solution. So I will move it (as you suggested) to the `EditContentHost` class.

**Yeah for now that is a better solution.** 

> I disagree with the statement that `is_displayed()` is not a 'proper' wait mechanism. `is_displayed()` calls [self.browser.wait_for_element](https://github.com/SatelliteQE/airgun/blob/master/airgun/views/contenthost.py#L118) which again calls [wait_for](https://github.com/RedHatQE/widgetastic.core/blob/master/src/widgetastic/browser.py#L432)
> So in your example `wait_for` calls `is_displayed` which again calls `wait_for` so the first call of wait_for is absolutely not required.

**If you look at the code in the is_displayed() method currently does not have any way to include increased timeout unless you change it or add the wrapper. It is taking 5 sec as a timeout. This is fine in most cases but certainly covers all the cases which in your case I'm not aware. The main point was to have that part more abstract rather than in the main read function. I do agree that it will use the wait which will work in your case. if applicable**   

> Furthermore the pattern mentioned above seems to be quite repetitive and could be abstracted in a single place?

**I have not seen this being an issue in many places (I see 1 or 2 exceptions), if that is the case I would think of adding an abstract layer to cover this. This probably again slows down the test execution as it will execute all the entity displayed calls which we do not want to have.**

PR #1286, opened 2024-03-14
[6.15.z] Check that breadcrumb is displayed

Satellite-QE:
Cherrypick of PR: https://github.com/SatelliteQE/airgun/pull/1281

error occurred in 6.15.z

### DIFF
--- a/airgun/entities/contenthost.py
+++ b/airgun/entities/contenthost.py
@@ -39,6 +39,7 @@ class ContentHostEntity(BaseEntity):
     def read(self, entity_name, widget_names=None):
         """Read content host details, optionally read only the widgets in widget_names."""
         view = self.navigate_to(self, 'Edit', entity_name=entity_name)
+        view.is_displayed
         return view.read(widget_names=widget_names)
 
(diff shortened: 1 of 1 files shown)


LadislavVasina1 (changes requested):
In my opinion, this does totally nothing. 
Correct me if I am wrong.
It just returns bool and nothing is done with it.

dosas:
@LadislavVasina1 That is not correct it does not do nothing, it waits for the element to be displayed.

No, it makes much more sense to use the abstraction. That is why it is there.

LadislavVasina1:
@dosas exception false is just from this implementation, that your implementation actual uses https://github.com/SatelliteQE/airgun/blob/7995b52325eac8696bb49d2618458f82839ca88e/airgun/views/contenthost.py#L122-L134 
See: https://github.com/SatelliteQE/airgun/pull/1281#issuecomment-2006539110

LadislavVasina1:
exception false is set because you want to get True or False from this property and do with this information whatever you want

dosas:
Closed in favor of https://github.com/SatelliteQE/airgun/pull/1299

PR #1299, opened 2024-03-28
Rework is_displayed logic for contenthost

dosas:
Labels: master, 6.15.z  

See also: https://github.com/SatelliteQE/airgun/pull/1281#issuecomment-2019926059

relevant test:
tests/foreman/ui/test_repository.py::test_positive_create_as_non_admin_user

### DIFF
--- a/airgun/entities/contenthost.py
+++ b/airgun/entities/contenthost.py
@@ -39,7 +39,8 @@ class ContentHostEntity(BaseEntity):
     def read(self, entity_name, widget_names=None):
         """Read content host details, optionally read only the widgets in widget_names."""
         view = self.navigate_to(self, 'Edit', entity_name=entity_name)
-        view.is_displayed
+        view.wait_displayed()
+        self.browser.plugin.ensure_page_safe()
         return view.read(widget_names=widget_names)
 
     def read_legacy_ui(self, entity_name, widget_names=None):


airgun/entities/contenthost.py
```diff
@@ -39,7 +41,6 @@ def read_all(self, widget_names=None):
     def read(self, entity_name, widget_names=None):
         """Read content host details, optionally read only the widgets in widget_names."""
         view = self.navigate_to(self, 'Edit', entity_name=entity_name)
-        view.is_displayed
```
omkarkhatavkar, line 42: have you tried using below code earlier. we might be not needed to so that whole stuff.   

```python
       view.wait_displayed()
```

dosas: In this [comment](d9f2c1bedad625bfe8a7839b205a21de9918f150) you said that you do not want to have it in the read method and it should be moved to the navigation steps and now it is fine to have it in the read method? I am confused.

omkarkhatavkar:
```python 
       view.wait_displayed()
```
you could have multiple approaches to solve the problem, If the above code does not work for you then I'm fine with the current approach.

dosas:
@omkarkhatavkar You removed the labels but I would require it in master and 6.15.z.

airgun/entities/contenthost.py
omkarkhatavkar: I should have mentioned earlier, come to know the later part. That's fine, thanks for doing those changes.

PR #1337, opened 2024-04-15
Added entity that returns counts of the errata types

LadislavVasina1:
I have extended host_new entity for a function that returns a dictionary with counts of the errata types 
['Security', 'Bugfix', 'Enhancement'].

### DIFF
--- a/airgun/entities/host_new.py
+++ b/airgun/entities/host_new.py
@@ -300,6 +300,30 @@ class NewHostEntity(HostEntity):
         view.content.errata.table.wait_displayed()
         return view.read(widget_names="content.errata.table")
 
+    def get_errata_type_counts(self, entity_name):
+        """
+        Get errata counts for each type of errata on selected host.
+
+        Args:
+            entity_name: Name of the host.
+        Returns:
+            errata_counts (dict): Dictionary with counts of each type of errata.
+        """
+
+        errata_types = ['Security', 'Bugfix', 'Enhancement']
+        errata_counts = {type: 0 for type in errata_types}
+        view = self.navigate_to(self, 'NewDetails', entity_name=entity_name)
+        view.wait_displayed()
+        view.content.errata.select()
+        for type in errata_types:
+            view.content.errata.wait_displayed()
+            view.content.errata.pagination.set_per_page(50)
+            view.content.errata.type_filter.fill(type)
+            self.browser.plugin.ensure_page_safe()
+            view.content.errata.table.wait_displayed()
+            errata_counts[type] = view.content.errata.table.row_count
+        return errata_counts
+
     def apply_erratas(self, entity_name, search=None):
         """Apply available errata on selected host based on searchbar result.
 


vsedmik (approved):
LGTM

airgun/entities/host_new.py
```diff
@@ -300,6 +300,30 @@ def get_errata_by_type(self, entity_name, type):
...
+        Args:
+            entity_name: Name of the host.
+        Returns:
+            errata_counts (dict): Dictionary with counts of each type of errata.
+        """
+
+        errata_types = ['Security', 'Bugfix', 'Enhancement']
+        errata_counts = {type: 0 for type in errata_types}
+        view = self.navigate_to(self, 'NewDetails', entity_name=entity_name)
+        view.wait_displayed()
```
vijaysawant, line 316: - Can we use `waif_for_element()` method (or `wait_for()`) just to avoid use of `time.sleep(sec)` [here test module](https://github.com/SatelliteQE/robottelo/pull/14773/files#diff-bd6863972772cca482646a04c83fba2589bdb45b1fcfdb0c4d544cd9fd17c920R1324), 
- Can take reference from same file.
```suggestion
        view.wait_displayed()
        self.browser.wait_for_element(<locator>, exception=False)
        OR
        wait_for(lambda: view.<web_element difined in VIEW>.is_displayed, timeout=5)
```

vijaysawant (changes requested):
Suggested one change, which ultimately helps in robottelo test module. this [time.sleep(2)](https://github.com/SatelliteQE/robottelo/pull/14773/files#diff-bd6863972772cca482646a04c83fba2589bdb45b1fcfdb0c4d544cd9fd17c920R1324) could be remove.

airgun/entities/host_new.py
omkarkhatavkar: @vijaysawant I dont think it is valid to hold this PR for the robottelo change. He is using the wait_disaplyed() internally uses the wait_for.

omkarkhatavkar (approved):

airgun/entities/host_new.py
vijaysawant: - No no @omkarkhatavkar, I am not holding Ack on PR, as in robottelo PR we are refreshing browser and immediatly calling this function, so I am thinking what if we use `wait_for_element` method.
- As in `wait_displayed()` we could pass timeout explicitly, may be this could help
`view.wait_displayed(timeout=5, delay=2)` @LadislavVasina1

damoore044 (approved):
ack nice addition!

airgun/entities/host_new.py
LadislavVasina1: But I need to wait before calling this function and mainly before the refresh in the test.

vijaysawant: as suggested changes doesn't work (I saw your response in rebottelo PR), I will leave upto you how to use `wait_for` / `wait_displayed` in this airgun PR.

PR #1384, opened 2024-05-07
Basic host statuses page

lhellebr:
For PR https://github.com/SatelliteQE/robottelo/pull/14986

### DIFF
--- a/airgun/views/host.py
+++ b/airgun/views/host.py
@@ -200,6 +200,20 @@ class HostInterface(View):
         return self.browser.wait_for_element(self.title, visible=True, exception=False) is not None
 
 
+class HostStatusesView(BaseLoggedInView):
+    title = Text("//h5[normalize-space(.)='Host Status Overview']")
+    status_green_total = Text("//div[contains(@class, 'status-count')][1]/a[1]")
+    status_green_owned = Text("//div[contains(@class, 'status-count')][1]/a[2]")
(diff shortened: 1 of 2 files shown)


airgun/views/host.py
```diff
@@ -200,6 +200,20 @@ def is_displayed(self):
         return self.browser.wait_for_element(self.title, visible=True, exception=False) is not None
 
 
+class HostStatusesView(BaseLoggedInView):
+    title = Text("//h5[normalize-space(.)='Host Status Overview']")
+    status_green_total = Text("//div[contains(@class, 'status-count')][1]/a[1]")
```
shweta83, line 205: Can we rename these to something more relatable? Like pass, fail and I don't know what does `yellow` mean.

lhellebr: I thought about it as well. I asked Adam about interpretation of these. And it is different for each status type (currently there is only Build without some plugins), they are not described in any way in the WebUI, one color can mean multiple things (installation running OR installation successful for green)... and to top it off, Build status type doesn't even use yellow - meaning there should never be any host in yellow status.

So I concluded these are not really pass, warn and fail, they are just... well... green, yellow and red :)

shweta83: Thanks for explaining. We can keep it as it is.

omkarkhatavkar (approved):

PR #1385, opened 2024-05-08
added new support for feature  manifest expires soon

vijaysawant:
### Problem Statement
REF - Notification when manifest is expiring soon
https://bugzilla.redhat.com/show_bug.cgi?id=2075163#c23

### Solution
Add functionality in entities and view to read mafifest expiring soon message.

### Related Issues
Robottelo PR-https://github.com/SatelliteQE/robottelo/pull/15003

### DIFF
--- a/airgun/entities/subscription.py
+++ b/airgun/entities/subscription.py
@@ -103,14 +103,18 @@ class SubscriptionEntity(BaseEntity):
             manage_view.close_button.click()
         return delete_message
 
-    def read_expire_manifest_message(self):  # new changes
+    def read_expire_manifest_message(self):
         """Read message displayed about 'manifest expiration' at Subscription Manifest section"""
+        expire_manifest_header = None
+        expire_manifest_message = None
         view = self.navigate_to(self, 'Manage Manifest')
         view.wait_animation_end()
-        if not view.manifest_alert.is_displayed:
-            return None
-        expire_manifest_header = view.manifest_expire_header.read()
-        expire_manifest_message = view.manifest_expire_message.read()
+        if view.manifest_alert.is_displayed:
+            expire_manifest_header = view.manifest_expire_header.read()
+            expire_manifest_message = view.manifest_expire_message.read()
+        else:
+            # Subscription Manifest header & message is not present
+            expire_manifest_message = 'Manifest expire alert not found'
         # close opened modal dialogs views
         manage_view = ManageManifestView(self.browser)
         if manage_view.is_displayed:
--- a/airgun/entities/subscription.py
+++ b/airgun/entities/subscription.py
@@ -118,7 +118,6 @@ class SubscriptionEntity(BaseEntity):
         if view.manifest.expire_date.is_displayed:
             expire_manifest_date = view.manifest.expire_date.read()
         else:
-            # Subscription Manifest expire date is not present
             raise Exception('Manifest expire date not found')
         # close opened modal dialogs views
         manage_view = ManageManifestView(self.browser)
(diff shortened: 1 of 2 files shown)


damoore044 (approved):
looks good, pending prt pass of robottelo MR

airgun/entities/subscription.py
```diff
@@ -103,6 +103,24 @@ def read_delete_manifest_message(self):
...
+        expire_manifest_header = None
+        expire_manifest_message = None
+        view = self.navigate_to(self, 'Manage Manifest')
+        view.wait_animation_end()
+        if view.manifest_alert.is_displayed:
+            expire_manifest_header = view.manifest_expire_header.read()
+            expire_manifest_message = view.manifest_expire_message.read()
+        else:
+            # Subscription Manifest header & message is not present
+            expire_manifest_message = 'Manifest expire alert not found'
```
jyejare, line 117: Should you raise as exception here ?

jyejare (commented):
ACK pending question

airgun/entities/subscription.py
vijaysawant: I think it's not needed, becuase inside test case we are updating settings for `expire soon days` and re-calling same function to get expected message (that time if block would executes)

airgun/entities/subscription.py
```diff
@@ -103,6 +103,24 @@ def read_delete_manifest_message(self):
...
+        # close opened modal dialogs views
+        manage_view = ManageManifestView(self.browser)
+        if manage_view.is_displayed:
+            manage_view.close_button.click()
+        return expire_manifest_header, expire_manifest_message
```
vsedmik, line 122: A small suggestion - wouldn't it better to return a dict here
```suggestion
        return {'header'=expire_manifest_header, 'message'=expire_manifest_message}
```
and reference it via keywords instead of indexes in the test? It would improve the test readability IMHO.

vsedmik (commented):
One suggestion plus I would lean towards @jyejare 's proposal for an exception.

airgun/entities/subscription.py
vijaysawant: sure, nice suggestion.

vsedmik (approved):
Looks good to me

airgun/views/subscription.py
```diff
@@ -158,6 +158,12 @@ class ManageManifestView(BaseLoggedInView):
 
     @View.nested
     class manifest(SatTab):
+        alert_message = Text(
+            '//*[@id="manifest-history-tabs-pane-1"]/div/h3//following-sibling::div[@aria-label="Warning Alert" or @aria-label="Danger Alert"]'
+        )
+        expire_header = Text('//*[@id="manifest-history-tabs-pane-1"]//following::div/h4')
+        expire_message = Text('//*[@id="manifest-history-tabs-pane-1"]//following::div/div[2]/span')
+        expire_date = Text('//*[@id="manifest-history-tabs-pane-1"]/div[1]/div[3]/div[2]')
```
omkarkhatavkar, line 166: Please specify the correct tag name instead of using '*' in the list of tag names.

airgun/entities/subscription.py
```diff
@@ -103,6 +103,43 @@ def read_delete_manifest_message(self):
...
+            expire_manifest_header = view.manifest.expire_header.read()
+            expire_manifest_message = view.manifest.expire_message.read()
+        else:
+            # Subscription Manifest header & message is not present
+            raise Exception('Manifest expire alert not found')
+        # Read manifest expiration date
+        if view.manifest.expire_date.is_displayed:
+            expire_manifest_date = view.manifest.expire_date.read()
+        else:
+            # Subscription Manifest expire date is not present
```
omkarkhatavkar, line 121: No need to provide comments for errors that are self-explanatory

PR #1390, opened 2024-05-21
browser teardown fix at session exit

pnovotny:
### Problem Statement

When nesting `airgun.session.Session` in multiple context managers (like robottelo `session` fixture does),
the `__exit__` method, which handles browser teardown, is called mutliple times as well.

This causes that the second and further attempt to call `webdriver.quit()` will fail,
because the browser was already closed.

### Solution

Execute the browser teardown only once.

### Related Issues

Needed for https://github.com/SatelliteQE/robottelo/pull/15050

### DIFF
--- a/airgun/session.py
+++ b/airgun/session.py
@@ -232,7 +232,7 @@ class Session:
             not risen not to shadow real session result.
         """
         if self.browser is None:
-            # browser was never started, don't do anything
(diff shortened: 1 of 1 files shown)


omkarkhatavkar (approved):

omkarkhatavkar:
This needs to be merged as it is highly dependent on the Robottelo PR.

PR #1405, opened 2024-05-30
6.15.z-fix subscription search box xpath

vijaysawant:
**Problem**
Content - Subscription page search box xpath unable to locate due to wrong xpath

**Solution**
This updated xpath will work as expected.

**Note**
Robottelo changes from the PR [[Link]](https://github.com/SatelliteQE/robottelo/pull/15237) are not dependant these changes as I have updated some part of test case ( `test_positive_access_with_non_admin_user_with_manifest`) which won't search for any Subscription

### DIFF
--- a/airgun/entities/subscription.py
+++ b/airgun/entities/subscription.py
@@ -176,6 +176,7 @@ class SubscriptionEntity(BaseEntity):
     def read_subscriptions(self):
         """Return subscriptions table"""
         view = self.navigate_to(self, 'All')
+        view.wait_displayed(timeout=10, delay=1)
         return view.table.read()
 
     def sca_alert(self):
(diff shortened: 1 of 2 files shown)


sambible (approved):
@omkarkhatavkar can you review?

airgun/entities/subscription.py
```diff
@@ -176,6 +176,11 @@ def delete(self, entity_name):
     def read_subscriptions(self):
         """Return subscriptions table"""
         view = self.navigate_to(self, 'All')
+        wait_for(
```
omkarkhatavkar, line 179: we are trying to read the table 2 times here. which is not a good way. Better way would be just add this line 

`view.wait_displayed()`

omkarkhatavkar (changes requested):
minor change

airgun/entities/subscription.py
sambible: ah yeah, this is definitely the better way to do this. fuzzy late night brain :) 

If that ends up not being specific enough, you can always use wait_displayed on a specific element, as well.

I.e. `view.table.wait_displayed`

vijaysawant: okay

omkarkhatavkar (approved):

vsedmik (approved):

PR #1425, opened 2024-06-12
Removing outdating views for hostcontent sca tests

ColeHiggins2:
Subscription info and System Purpose Status have been removed in stream. Removing them in views so they no longer cause issues with invalid locators 

Also removed "show_all" no longer an option... and updated the logic in the .read() method


LadislavVasina1 (approved):
ACK

omkarkhatavkar (approved):

PR #1427, opened 2024-06-18
Add Facts View and add function to read host facts

shweta83:
1. Added module for Facts page and locator for the facts table.
2. Added a function in host entities to read host facts.

### DIFF
--- a/airgun/entities/host_new.py
+++ b/airgun/entities/host_new.py
@@ -5,6 +5,7 @@ from wait_for import wait_for
 
 from airgun.entities.host import HostEntity
 from airgun.navigation import NavigateStep, navigator
+from airgun.views.fact import HostFactView
 from airgun.views.host_new import (
     AllAssignedRolesView,
     EditAnsibleRolesView,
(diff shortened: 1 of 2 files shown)


airgun/entities/host_new.py
```diff
@@ -949,6 +950,16 @@ def remediate_with_insights(
         view = RemediationView(self.browser)
         view.remediate.click()
 
+    def get_host_facts(self, entity_name):
+        view = self.navigate_to(self, 'NewDetails', entity_name=entity_name)
+        self.browser.plugin.ensure_page_safe()
+        view.wait_displayed()
+        self.browser.wait_for_element(view.dropdown, exception=False)
+        assert view.dropdown.has_item('Facts')
```
omkarkhatavkar, line 958: These are not required, anyways it will fail if it is not able to see the option of  `Facts`

airgun/views/fact.py
```diff
@@ -0,0 +1,20 @@
...
+        './/table',
+        column_widgets={
+            'Name': Text("./a"),
+        },
+    )
+
+    @property
+    def is_displayed(self):
+        breadcrumb_loaded = self.browser.wait_for_element(self.breadcrumb, exception=False)
+        return breadcrumb_loaded and self.breadcrumb.read().startswith('Facts Values ')
```
omkarkhatavkar, line 20: can you please avoid the space here `'Facts Values '`

omkarkhatavkar (changes requested):
Can you please attach the PRT result here ?

omkarkhatavkar:
I see the PR attached for the PRT, Thanks

omkarkhatavkar (approved):
ACK, 1 minor comment

airgun/views/fact.py
shweta83: Updated.

PR #1443, opened 2024-06-25
Manifest expiration date appearance of header and message

vijaysawant:
### Problem 
- Considering newly implemented 6.16.0 feature/RFE "Notification when manifest is going to expire" and reported bug "Manifest expiration date (year) shows older for all manifest" it is neccessary to check manifest expiration message is disappearing when manifest upload very first time into organization.
- It was left behind because related https://github.com/SatelliteQE/airgun/pull/1385 merged very early phase of testing.
- During phoenix qe team discussion, we have decided to open new robottelo PR and airgun PR to validate manifest expiration header message functionality before updating settings value (expire_soon_days)
### Solution
Implemented function will return True or False base on availability of header message on manage manifest modal box.
### Related PR
Robottelo https://github.com/SatelliteQE/robottelo/pull/15500

### DIFF
--- a/airgun/entities/subscription.py
+++ b/airgun/entities/subscription.py
@@ -140,12 +140,10 @@ class SubscriptionEntity(BaseEntity):
         return manifest_expiration_date
 
     def is_subscription_manifest_header_message_display(self):
-        """Is header & message present in manage manifest modal box or not"""
-        result = False
+        """Checks header and massage present in the manage manifest modal"""
         view = self.navigate_to(self, 'Manage Manifest')
         view.wait_animation_end()
-        if view.manifest.alert_message.is_displayed:
-            result = True
+        result = view.manifest.alert_message.is_displayed
         # close opened modal dialogs views
         manage_view = ManageManifestView(self.browser)
         if manage_view.is_displayed:
--- a/airgun/entities/subscription.py
+++ b/airgun/entities/subscription.py
@@ -139,6 +139,19 @@ class SubscriptionEntity(BaseEntity):
             manage_view.close_button.click()
         return manifest_expiration_date
 
+    def is_subscription_manifest_header_message_display(self):
+        """Is header & message present in manage manifest modal box or not"""
+        result = False
+        view = self.navigate_to(self, 'Manage Manifest')
+        view.wait_animation_end()
+        if view.manifest.alert_message.is_displayed:
+            result = True
+        # close opened modal dialogs views
+        manage_view = ManageManifestView(self.browser)
+        if manage_view.is_displayed:
+            manage_view.close_button.click()
+        return result
+
     def add(self, entity_name, quantity=1):
         """Attach new subscriptions
         :param entity_name: Name of subscription to attach


airgun/entities/subscription.py
```diff
@@ -139,6 +139,19 @@ def read_subscription_manifest_expiration_date_only(self):
             manage_view.close_button.click()
         return manifest_expiration_date
 
+    def is_subscription_manifest_header_message_display(self):
```
damoore044, line 142:
```suggestion
    def manifest_header_message_displayed(self):
```
Propose a shorter method name like this, or something similar?
Non-blocking

airgun/entities/subscription.py
```diff
@@ -139,6 +139,19 @@ def read_subscription_manifest_expiration_date_only(self):
...
+        result = False
+        view = self.navigate_to(self, 'Manage Manifest')
+        view.wait_animation_end()
+        if view.manifest.alert_message.is_displayed:
+            result = True
+        # close opened modal dialogs views
+        manage_view = ManageManifestView(self.browser)
+        if manage_view.is_displayed:
+            manage_view.close_button.click()
+        return result
```
damoore044, line 153: Is it possible to return the actual message if displayed, False if absent? 
Or do you think it is better as a strictly boolean return type?

airgun/entities/subscription.py
```diff
@@ -139,6 +139,19 @@ def read_subscription_manifest_expiration_date_only(self):
             manage_view.close_button.click()
         return manifest_expiration_date
 
+    def is_subscription_manifest_header_message_display(self):
+        """Is header & message present in manage manifest modal box or not"""
+        result = False
+        view = self.navigate_to(self, 'Manage Manifest')
+        view.wait_animation_end()
+        if view.manifest.alert_message.is_displayed:
+            result = True
```
LadislavVasina1, line 148:
```suggestion
        result = view.manifest.alert_message.is_displayed:
```
Can't you just assign it like this? (also removing result = False on line 144) `is_displayed` returns bool

airgun/entities/subscription.py
```diff
@@ -139,6 +139,19 @@ def read_subscription_manifest_expiration_date_only(self):
             manage_view.close_button.click()
         return manifest_expiration_date
 
+    def is_subscription_manifest_header_message_display(self):
+        """Is header & message present in manage manifest modal box or not"""
```
ColeHiggins2, line 143: "Checks header and massage present in the manage manifest modal"
Non - blocking

ColeHiggins2 (approved):
Ack, pending comments

airgun/entities/subscription.py
vijaysawant: This function is being used/called during testing subscription expiration date scenario, I think it makes more sense to use well descriptive name instead of shorter. This will return True or False based of condition "is header message currently displaying on page".

airgun/entities/subscription.py
vijaysawant: there is another function which is returning 3 things (header, message & date) if you check `read_subscription_manifest_header_message_and_date`. This is just to check presence/absence of header message, we want only boolean value out of it. I would like to keep the behaviour as it is.

airgun/entities/subscription.py
vijaysawant: Suggested changes should work, will test it and update PR.

damoore044 (approved):

sambible (approved):

PR #1447, opened 2024-06-26
Fix race condition in PF4 progressbar

dosas:
when the progress bar disappears between the
`is_displayed` query and the `current_progress` query
a NoSuchElement exception is raised


omkarkhatavkar:

dosas:
Closing in favor of https://github.com/SatelliteQE/airgun/pull/1448

PR #1453, opened 2024-07-01
hosts table: Name locator fix

pnovotny:
### Problem Statement

The `HostsView` `Name` table column XPath was missing leading '.' character,
which caused that all host rows that were returned, had the same Name value,
equal to the first table row host Name.


LadislavVasina1 (approved):

pnovotny:
PRT of the parent PR https://github.com/SatelliteQE/robottelo/pull/15539 passed.

pondrejk (approved):

PR #1476, opened 2024-07-24
making auto-merge should be same as master

omkarkhatavkar:
Automerge is failing because of wait for status looking wrong Runner Context, This PR will fix that


omkarkhatavkar:
self-merging this because of priority

PR #1487, opened 2024-08-08
[6.16] Adjust all hosts to use menu, add support for Build Management

sambible:
Adds support for Build Management in All Hosts page, as well as updates the dropdown to a Menu, matching recent changes from the dev side.

Note: Not all of these changes appear to have made it into stream, so there will likely need to be more work with this component.


LadislavVasina1 (approved):
Looks good to me, I have also implemented the change of the kebab button from the Dropdown to Menu locally myself, your implementation seems to work too, I will try to adapt it into my work (so I will postpone my PR for Manage Packages feature for a bit till this is merged)

pondrejk (approved):

PR #1500, opened 2024-08-16
Fix ruff deprecation warnings

ogajduse:
Fixing:
```
$ ruff check .
warning: The top-level linter settings are deprecated in favour of their counterparts in the `lint` section. Please update the following options in `pyproject.toml`:
  - 'fixable' -> 'lint.fixable'
  - 'ignore' -> 'lint.ignore'
  - 'select' -> 'lint.select'
  - 'flake8-pytest-style' -> 'lint.flake8-pytest-style'
  - 'isort' -> 'lint.isort'
  - 'mccabe' -> 'lint.mccabe'
  - 'per-file-ignores' -> 'lint.per-file-ignores'
warning: `PGH001` has been remapped to `S307`.
warning: `TRY200` has been remapped to `B904`.
```

I removed `PGH001` as `S307` is in the `select` section already

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,6 +19,8 @@ exclude = '''
 
 [tool.ruff]
 target-version = "py311"
+
+[tool.ruff.lint]
 fixable = ["ALL"]
 
 select = [
@@ -31,6 +33,7 @@ select = [
     "B014", # Exception handler with duplicate exception
     "B023", # Function definition does not bind loop variable {name}
     "B026", # Star-arg unpacking after a keyword argument is strongly discouraged
+    "B904", # Use raise from to specify exception cause
     "BLE001", # Using bare except clauses is prohibited
     "C", # complexity
     "C4", # flake8-comprehensions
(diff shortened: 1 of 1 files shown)


Gauravtalreja1 (approved):

pyproject.toml
```diff
@@ -19,6 +19,8 @@ exclude = '''
 
 [tool.ruff]
 target-version = "py311"
```
LadislavVasina1, line 21: @ogajduse just a side comment not regarding these changes, shouldn't we change the ruff target version to `py312` when we already use 3.12 in our frameworks, same on line 8 in tool.black target-version

ogajduse: We should. That should be done, however, every time we bump the Python version. AFAIK there are no guidelines or documentation, making it easy to forget.

ogajduse: Looking at https://docs.astral.sh/ruff/configuration, `target-version` is `The minimum Python version that should be supported`. Therefore it should be equal to the minimal version that we support. We still support py310 - defined by the GH Actions matrix.

PR #1515, opened 2024-08-20
New job invocation details page

pnovotny:
New view and navigation was added for the first part of the new job invocation details page.

This feature is still an experimantal feature/tech preview and it is not yet fully completed.
Other content will be added in later phases.

### DIFF
--- a/airgun/entities/job_invocation.py
+++ b/airgun/entities/job_invocation.py
@@ -11,6 +11,7 @@ from airgun.views.job_invocation import (
     JobInvocationCreateView,
     JobInvocationStatusView,
     JobInvocationsView,
+    NewJobInvocationStatusView,
 )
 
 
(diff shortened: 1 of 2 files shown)


pondrejk (approved):

lhellebr (approved):
ACK once PRT passes repeatedly (to avoid race conditions)

pnovotny:
I am trying to solve a problem, where the new details page doesn't load completely when running the PRT in our Selenium Grid w/ Chrome. 
The page is only partially loaded like this:
job-inv-not-fully-loaded

But it loads fine when running the test locally w/ Firefox or Chrome:
job-inv-fully-loaded

pnovotny:
PRT on https://github.com/SatelliteQE/robottelo/pull/15980 passed (twice).
@pondrejk @lhellebr Can we pls merge?

LadislavVasina1 (approved):
ACK

PR #1589, opened 2024-10-07
Add entity that reads filled searchbox on new and old hosts pages

LadislavVasina1:
Add an entity `read_filled_searchbox` to the `host.py` (old hosts UI) and `all_hosts.py` (new hosts UI) so that it can be used in the automation.

Usage:

- Old UI

```python
with target_sat.ui_session() as session:
    session.host.search('some_search_string')
    read_searchbox_value = session.host.read_filled_searchbox()
    print(read_searchbox_value)
```

- New UI

```python
with target_sat.ui_session() as session:
    session.all_hosts.search('some_search_string')
    read_searchbox_value = session.all_hosts.read_filled_searchbox()
    print(read_searchbox_value)
```

PR #1591, opened 2024-10-08
check facts for host

amolpati30:
On the host, it selects the facts and returns the value to verify whether the 'insights_client' is present.

Dependent PR: https://github.com/SatelliteQE/robottelo/pull/16606

### DIFF
--- a/airgun/entities/host.py
+++ b/airgun/entities/host.py
@@ -426,6 +426,19 @@ class HostEntity(BaseEntity):
         view.wait_displayed()
         return view.displayed_table_header_names
 
+    def build_mode(self, entity_name, value=None):
+        """On the host, it checks the facts and returns the value.
+
+        :steps: select host - click 3 dot - facts
+        """
+        view = self.navigate_to(self, 'All')
+        view.table.row(name=entity_name)['Name'].widget.click()
+        host_view = NewHostDetailsView(self.browser)
+        host_view.wait_displayed()
+        host_view.dropdown.wait_displayed()
+        host_view.dropdown.item_select(value)
+        return host_view.read()
+
 
 @navigator.register(HostEntity, 'All')
 class ShowAllHosts(NavigateStep):


airgun/entities/host.py
```diff
@@ -419,6 +419,15 @@ def get_displayed_table_headers(self):
         view.wait_displayed()
         return view.displayed_table_header_names
 
+    def build_mode(self, entity_name, value=None):
```
LadislavVasina1, line 422: `Added a method to verify the facts for the host.` Can you please more elaborate on the PR description and add some docstring to your entity and maybe alter the entity name so it more reflects what the entity is doing. Thanks!

LadislavVasina1 (changes requested):
Added a comment to get more clarity about the addition.

airgun/entities/host.py
amolpati30: Updated.

LadislavVasina1:
@amolpati30 I can't see you using the airgun entity I coded for you https://github.com/SatelliteQE/airgun/pull/1589, why didn't you use it? Also leaving commit history would be good review of quarter-old PR.

amolpati30:
@LadislavVasina1 
Will your PR verify whether the insights_client is present in the facts? I believe it is providing some other information, which is not exactly what I need for the test case.

LadislavVasina1:
Didn't you want to read the filled searchbox when you created the PR? I suppose you have edited the description in the meanwhile so it is probable that my addition might be no use to you as it was implement for former PR description and former commit.

amolpati30:
@LadislavVasina1 
As per the bug, I updated the test to verify that insights are present in the facts and that registration works without any issues. I believe I have addressed all the required aspects accordingly.

shweta83:
@amolpati30 There is a method present in [airgun](https://github.com/SatelliteQE/airgun/blob/ca67a487b5aa0523867cfa631494660294cc48ed/airgun/entities/host_new.py#L959) to read host facts and you can pass the fact name in robottelo which you want to check.

amolpati30:
@shweta83 
It is working, This PR is not necessary. I have updated the Robottelo PR; please take a look.

shweta83:
> @shweta83 Existing method is working, This PR is not necessary. I have updated the Robottelo PR; please take a look.

Great! So can we close this PR?

PR #1611, opened 2024-10-29
Add support for getting documentation related urls from Satellite page

jameerpathan111:
Historically, after a Satellite release, we have found that the documentation links on Satellite pages were broken, causing bugs and customer cases to be filed, resulting in a poor user experience. 
This PR adds support to get the Documentation related links from a Satellite page. It also adds basic code for missing entities/pages in Airgun/

See SAT-27552

### DIFF
--- a/airgun/views/common.py
+++ b/airgun/views/common.py
@@ -1,3 +1,4 @@
+from selenium.common.exceptions import ElementNotInteractableException
 from widgetastic.widget import (
     Checkbox,
     ConditionalSwitchableView,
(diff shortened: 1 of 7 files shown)


jameerpathan111:
Tested the code locally with Satellite 6.15

airgun/views/common.py
```diff
@@ -80,6 +81,35 @@ def read(self, widget_names=None, limit=None):
                 values[widget_name] = widget.read()
         return normalize_dict_values(values)
 
+    def documentation_links(self):
+        """Return Documentation links present on the given page if any.
+        Note: This is not a full-proof helper. For example, it can't get links hidden behind a dropdown button.
```
LadislavVasina1, line 86: I would say, let's add that feature of getting the link from the dropdowns later. Right? :D

LadislavVasina1 (approved):
Seems like a great addition. GJ!

airgun/views/common.py
jameerpathan111: yes, I'd leave it for later too. One problem at a time:)

LadislavVasina1: Agree

pondrejk (approved):

PR #1615, opened 2024-10-30
Add reclaim_space_warning node to the capsule view

vsedmik:
Just adding the reclaim_space_warning node to the capsule view so that we can test its hiding from robottelo.


chris1984 (approved):
Looks good on the test and us checking to make sure the text is there.

LadislavVasina1 (approved):

chris1984:
@vsedmik 2 green acks :) thanks for explaining how it works in Slack.

PR #1617, opened 2024-10-31
[6.16.z] Add reclaim_space_warning node to the capsule view

Satellite-QE:
Cherrypick of PR: https://github.com/SatelliteQE/airgun/pull/1615

Just adding the reclaim_space_warning node to the capsule view so that we can test its hiding from robottelo.


vsedmik (approved):
Looks CPed correctly.

PR #1685, opened 2024-12-19
Switch Airgun CI/CQ to uv

ogajduse:
This PR switches `pip` to `uv` as @JacobCallahan requested in the chat.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -7,6 +7,7 @@ on:
 
 env:
     PYCURL_SSL_LIBRARY: openssl
+    UV_SYSTEM_PYTHON: 1
 
 jobs:
   codechecks:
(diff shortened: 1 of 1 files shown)


.github/workflows/pull_request.yml
```diff
@@ -50,6 +67,18 @@ jobs:
         with:
           python-version: '3.12'
 
+      - name: Set up uv
+        run: curl -LsSf https://astral.sh/uv/install.sh | sh
```
JacobCallahan, line 71: we could also move to using this: https://github.com/astral-sh/setup-uv

not a blocker though, since this works fine as well

JacobCallahan (approved):

.github/workflows/pull_request.yml
ekohl: https://docs.astral.sh/uv/guides/integration/github/ has plenty of recommendations on how to do that, including caching.

ogajduse: haha, I saw your suggestion just now. I figured it out independently. I have just pushed the change. Let's check the results...

ekohl (approved):
Looks much cleaner now with this action

LadislavVasina1 (approved):

PR #1708, opened 2025-01-22
Extend sync_status page with active_only support

vsedmik:
There is the "Active only" check-box on the sync status page which adjusts the product/repos table to hide the lines where no sync happens. This is done through the `tr`'s `style` set to `display: none`, which means the table remains still complete, only the style change hides what needs to be hidden.

image

Unfortunately Airgun does not care about styles and returns the complete table since it's still there in the page.

This PR changes this behaviour to let Airgun return the table just like the user can see it based on the style (via `is_displayed` property).
Second change allows user to set the Active only check-box before read.
Third and last change allows user to sync a repo asynchronously without waiting for result.

Test PR: https://github.com/SatelliteQE/robottelo/pull/17413

### DIFF
--- a/airgun/entities/sync_status.py
+++ b/airgun/entities/sync_status.py
@@ -9,16 +9,18 @@ from airgun.views.sync_status import SyncStatusView
 class SyncStatusEntity(BaseEntity):
     endpoint_path = '/katello/sync_management'
 
-    def read(self, widget_names=None):
+    def read(self, widget_names=None, active_only=False):
         """Read all widgets at Sync status entity"""
         view = self.navigate_to(self, 'All')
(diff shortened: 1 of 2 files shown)


vijaysawant (approved):

LadislavVasina1 (approved):
PR looks good to me, let's wait for some Robotello PRs testing this addition before merging. Good addition @vsedmik!

vsedmik:
> PR looks good to me, let's wait for some Robotello PRs testing this addition before merging. Good addition @vsedmik!

Ah, I forgot to update the PR description. The robo PR is up here (with green PRT): https://github.com/SatelliteQE/robottelo/pull/17413

PR #1800, opened 2025-04-17
Add helper entites for RH inventory

LadislavVasina1:
This PR adds some helper entities used in https://github.com/SatelliteQE/robottelo/pull/18259 and updates some views.

PRT runs in the robottelo PR.

### DIFF
--- a/airgun/views/cloud_inventory.py
+++ b/airgun/views/cloud_inventory.py
@@ -3,8 +3,9 @@ from wait_for import wait_for
 from widgetastic.utils import ParametrizedLocator
 from widgetastic.widget import Text, View
 from widgetastic_patternfly import Button, Tab
-from widgetastic_patternfly4.button import Button as Pf4Button
 from widgetastic_patternfly4.switch import Switch
+from widgetastic_patternfly5 import Button as PF5Button
+from widgetastic_patternfly5.ouia import Text as PF5OUIAText
 
 from airgun.exceptions import ReadOnlyWidgetError
 from airgun.views.common import BaseLoggedInView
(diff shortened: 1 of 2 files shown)


vsedmik (approved):
Codewisely looks good to me, PRT at robo PR passed -> ACK

airgun/views/cloud_inventory.py
```diff
@@ -110,11 +111,18 @@ class CloudInventoryListView(BaseLoggedInView):
...
     obfuscate_ips = Switch('.//label[@for="rh-cloud-switcher-obfuscate_inventory_ips"]')
     exclude_packages = Switch('.//label[@for="rh-cloud-switcher-exclude_installed_packages"]')
-    cloud_connector = Pf4Button(locator='//button[normalize-space(.)="Configure cloud connector"]')
-    reconfigure_cloud_connector = Pf4Button(
+    auto_mismatch_deletion = Switch(
+        './/label[@for="rh-cloud-switcher-allow_auto_insights_mismatch_delete"]'
+    )
+
+    auto_upload_desc = PF5OUIAText('OUIA-Generated-Text-6')
+    manual_upload_desc = PF5OUIAText('OUIA-Generated-Text-7')
```
pnovotny, line 119: I see that we use some these auto-generated IDs also elsewhere, but they can change over time, when some widgets get added or removed on the page.
If there is not a better locator, I'm fine with it, but we need to be aware that they don't change and then point to a different element.

ColeHiggins2 (approved):

ColeHiggins2:
Have enough Acks. Merging

PR #1801, opened 2025-04-17
wait_to_success_job_task

amolpati30:
The page is not loading in time, and that is why unable to click `Dynflow console` button.
Because the button doesn't appear until the page reloads properly.

To avoid this, it will wait until the task result is successful.


Gauravtalreja1 (approved):
ACK, PRT passed on https://github.com/SatelliteQE/robottelo/pull/18253

LadislavVasina1 (approved):

PR #1842, opened 2025-05-07
Fix Login page locators for Stream

jameerpathan111:
Locators for `login_text` and `logo` have changed in the latest Stream snap.


lpramuk (approved):

rmynar (approved):
It seems that the logo is useless. Maybe some historical reasons, so i I'm OK if we keep it.

LadislavVasina1 (approved):

PR #1873, opened 2025-06-10
fix: broken techdocs annotations

yashoswalyo:
Removed the techdocs annotation, as it only supports building docs from md files


Satellite-QE:

ogajduse (approved):

PR #1883, opened 2025-06-17
Add view and entities for Hosts->CVEs tab and CVE list page

Gauravtalreja1:
Adding UI changes from https://github.com/theforeman/foreman_rh_cloud/pull/988 and https://github.com/theforeman/foreman_rh_cloud/pull/999

### DIFF
--- a/airgun/entities/host_new.py
+++ b/airgun/entities/host_new.py
@@ -908,6 +908,12 @@ class NewHostEntity(HostEntity):
         view.wait_displayed()
         self.browser.plugin.ensure_page_safe()
         return view.insights.read()
+    
+    def get_cves(self, entity_name):
+        view = self.navigate_to(self, 'NewDetails', entity_name=entity_name)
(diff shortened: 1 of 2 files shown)


airgun/entities/host_new.py
```diff
@@ -908,6 +908,12 @@ def get_insights(self, entity_name):
         view.wait_displayed()
         self.browser.plugin.ensure_page_safe()
         return view.insights.read()
+    
+    def get_cves(self, entity_name):
```
sambible, line 912: This is an unnecessary method, you can use https://github.com/SatelliteQE/airgun/blob/da69b696cd14095420d947348fa044fd5ac9179a/airgun/entities/host_new.py#L56 to read this widget instead

sambible (changes requested):

Gauravtalreja1:
Thanks @Lennonka , I had noticed this change earlier and have now incorporated it into this PR

PR #1909, opened 2025-07-01
Disassociate Hosts Coverage [SAT-31029]

LadislavVasina1:
This PR adds airgun coverage for Disassociate hosts function coming to stream on All hosts page.
See SAT-31029.
Needed by: https://github.com/SatelliteQE/robottelo/pull/18859

## Summary by Sourcery

Add Airgun test coverage for the Disassociate Hosts function on the All hosts page, including navigation, selection logic, and modal interactions.

New Features:
- Add disassociate_hosts action to automate disassociating hosts from the All hosts page
- Introduce DisassociateHostsModal view to represent and interact with the disassociate hosts modal

Enhancements:
- Validate mutually exclusive select_all_hosts and host_names parameters
- Support both individual host selection and select-all flow in the disassociate_hosts method

PR #1927, opened 2025-07-14
all host change org and location airgun support

vijaysawant:
Feature story https://issues.redhat.com/browse/SAT-31030 Change the location and organization of multiple hosts
Robottelo PR: https://github.com/SatelliteQE/robottelo/pull/18980


airgun/views/all_hosts.py
```diff
@@ -605,3 +607,61 @@ class DisassociateHostsModal(PF5Modal):
...
+            else:
+                raise ValueError(f"Invalid option: {option}. Must be 'Fix on mismatch' or 'Fail on mismatch'")
+
+    def save(self):
+        """Click the Save button to apply changes."""
+        self.save_button.click()
+
+    def cancel(self):
+        """Click the Cancel button to close modal without saving."""
+        self.cancel_button.click()
```
LadislavVasina1, line 667: @vijaysawant I would like to keep the logic of the added views out of the "views" files and keep it just to "entities" files as we do elsewhere.
You can get inspired i.e. here https://github.com/SatelliteQE/airgun/pull/1909.
There you can see clearly differentiated view definition in the "view" file and the logic using that view in the "entities" file.

airgun/views/all_hosts.py
```diff
@@ -605,3 +607,61 @@ class DisassociateHostsModal(PF5Modal):
...
+
+class ChangeOrganizationLocationModal(PF5Modal):
+    """
+        This class represents 'Change organization/location' modal that is used to change org and location
+        for one or more hosts.
+    """
+    OUIA_ID = 'bulk-assign-taxonomy-modal'
+
+    # Organization section
+    organization_select = PF5Select(locator='.//div[@data-ouia-component-id="select-organization"]')
```
LadislavVasina1, line 619: do, at the top of the file:
`from widgetastic_patternfly5.ouia import Select as PF5OUIASelect`
and use it like this
```suggestion
    organization_select = PF5Select('select-organization')
```

airgun/views/all_hosts.py
```diff
@@ -605,3 +607,61 @@ class DisassociateHostsModal(PF5Modal):
...
+    organization_fix_on_mismatch = PF5Radio(id='radio-org-fix-on-mismatch')
+    organization_fail_on_mismatch = PF5Radio(id='radio-org-fail-on-mismatch')
+
+    # Location section
+    location_select = PF5Select(locator='.//div[@data-ouia-component-id="select-location"]')
```
LadislavVasina1, line 624: same as above
```suggestion
    location_select = PF5OUIASelect('select-location')
```

airgun/entities/all_hosts.py
```diff
@@ -547,6 +548,72 @@ def disassociate_hosts(self, host_names, select_all_hosts=False):
...
+
+        view = self.select_any_number_of_hosts(host_names, select_all_hosts)
+
+        # Open change organization and location modal
+        view.bulk_actions_kebab.click()
+
+        self.browser.move_to_element(view.bulk_actions_menu.item_element('Change associations'))
+        view.bulk_actions_manage_content_menu.item_select('Organization/location')
+
+        view = ChangeOrganizationLocationModal(self.browser)
```
LadislavVasina1, line 614: I don't see the usage of the view here, please move filling logic here

airgun/entities/all_hosts.py
```diff
@@ -547,6 +548,72 @@ def disassociate_hosts(self, host_names, select_all_hosts=False):
         view = DisassociateHostsModal(self.browser)
         view.confirm_btn.click()
 
+    def select_any_number_of_hosts(
```
LadislavVasina1, line 551: Good idea with this helper 👍🏻 , I have the similar approach implemented right now locally, it would be also nice to use it everywhere in the file where it is possible to use it, but you don't have to do that in this PR as I have this change prepared locally as all_hosts.py files refactor I am planning to push in near future.

PR #1928, opened 2025-07-15
Update insights to lightspeed select and locators

ColeHiggins2:
SAT-34946

### DIFF
--- a/airgun/entities/cloud_insights.py
+++ b/airgun/entities/cloud_insights.py
@@ -51,21 +51,21 @@ class CloudInsightsEntity(BaseEntity):
 
 @navigator.register(CloudInsightsEntity, 'Token')
 class SaveCloudTokenView(NavigateStep):
-    """Navigate to main Red Hat Insights page"""
+    """Navigate to main Red Hat Lightspeed page"""
 
     VIEW = CloudTokenView
 
     @retry_navigation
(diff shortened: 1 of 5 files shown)


airgun/entities/cloud_insights.py
```diff
@@ -51,21 +51,21 @@ def update(self, values):
 
 @navigator.register(CloudInsightsEntity, 'Token')
 class SaveCloudTokenView(NavigateStep):
-    """Navigate to main Red Hat Insights page"""
+    """Navigate to main Red Hat Lightspeed page"""
```
Gauravtalreja1, line 54: This file needs more updates for `s/Insights/Lightspeed`, could you update those too?

synkd (approved):

chris1984 (approved):

LadislavVasina1 (approved):

airgun/entities/cloud_insights.py
ColeHiggins2: I am only updated the doc string for any class or method that actually needed a Insights to light speed change. Changed every doc string and var name across robottelo will be another task once we finally update all the light speed changes in satellite. This feature was just a quick branding change

Gauravtalreja1 (approved):

PR #1983, opened 2025-08-20
[RFE][IOP] Add recommendations views for iop

ColeHiggins2:
Draft state of IOP Recommendations view/entities

### DIFF
--- a/airgun/entities/cloud_insights.py
+++ b/airgun/entities/cloud_insights.py
@@ -1,7 +1,7 @@
 from airgun.entities.base import BaseEntity
 from airgun.navigation import NavigateStep, navigator
 from airgun.utils import retry_navigation
-from airgun.views.cloud_insights import CloudInsightsView, CloudTokenView
+from airgun.views.cloud_insights import CloudInsightsView, CloudTokenView, RecommendationsTabView, RecommendationsTableExpandedRowView
 
 
 class CloudInsightsEntity(BaseEntity):
@@ -48,6 +48,35 @@ class CloudInsightsEntity(BaseEntity):
         view = self.navigate_to(self, 'All')
         view.fill(values)
 
+class RecommendationsTabEntity(BaseEntity):
+    endpoint_path = '/foreman_rh_cloud/insights_cloud'
+
+    def search(self, value):
+        """Search for 'query' and return matched hostnames/recommendations.
+
+        :param value: text to filter (default: no filter)
+        """
+        view = self.navigate_to(self, 'All')
+        view.clear_button.click()
+        view.search_field.fill(value)
+        return self.table.read()
+
+    def affected_systems(self):
+        """"""
+        view = self.navigate_to(self, 'All')
+        view.table.click()
+        #view = self.edit_helper(acs_name_to_edit)
+        # Close side panel view
+        view.capsules.affected_systems_url.click()
+        view = RecommendationsTableExpandedRowView(self.browser)
+        #view.wait_for(lambda: view.available_options_search.is_displayed, timeout=10, delay=1)
+
+    def read(self, widget_names=None):
+        """Read all values."""
+        view = self.navigate_to(self, 'All')
+        self.browser.plugin.ensure_page_safe(timeout='10s')
+        view.wait_displayed()
+        return view.read(widget_names=widget_names)
 
 @navigator.register(CloudInsightsEntity, 'Token')
 class SaveCloudTokenView(NavigateStep):
@@ -65,6 +94,17 @@ class ShowCloudInsightsView(NavigateStep):
     """Navigate to main Red Hat Lightspeed page"""
 
     VIEW = CloudInsightsView
+
+    @retry_navigation
+    def step(self, *args, **kwargs):
+        self.view.menu.select('Red Hat Lightspeed', 'Recommendations')
+
+
+@navigator.register(RecommendationsTabEntity, 'All')
+class ShowRecommendationsView(NavigateStep):
+    """Navigate to main Red Hat Lightspeed page"""
+
+    VIEW = RecommendationsTabView
 
     @retry_navigation
     def step(self, *args, **kwargs):
--- a/airgun/views/cloud_insights.py
+++ b/airgun/views/cloud_insights.py
@@ -2,16 +2,18 @@ from widgetastic.widget import Checkbox, Text, TextInput, View
 from widgetastic_patternfly5 import (
     Button as PF5Button,
     Pagination as PF5Pagination,
+    Title as PF5Title,
 )
 from widgetastic_patternfly5.ouia import (
     Dropdown as PF5OUIADropdown,
     Modal as PF5OUIAModal,
     PatternflyTable as PF5OUIAPatternflyTable,
     Switch as PF5OUIASwitch,
+    Text as PF5Text,
+    ExpandableTable as PF5ExpandableTable,
 )
 
 from airgun.views.common import BaseLoggedInView, SearchableViewMixinPF4
-
 
 class CloudTokenView(BaseLoggedInView):
     """Red Hat Lightspeed Landing page for adding RH Cloud Token."""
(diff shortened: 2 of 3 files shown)


airgun/entities/cloud_insights.py
```diff
@@ -48,6 +48,40 @@ def update(self, values):
...
+        """Search for 'query' and return matched hostnames/recommendations.
+
+        :param value: text to filter (default: no filter)
+        """
+        view = self.navigate_to(self, 'All')
+        view.clear_button.click()
+        view.search_field.fill(value)
+        return self.table.read()
+
+    def affected_systems(self, recommendation_name: str):
```
LadislavVasina1, line 64:
```suggestion
    def open_affected_systems(self, recommendation_name: str):
```
I would name the entity like this ^^^ so it is clear from the name what it does.

airgun/entities/cloud_insights.py
```diff
@@ -49,6 +60,93 @@ def update(self, values):
...
+        # Use navigator to open the Affected Systems details view
+        view = self.navigate_to(self, 'Affected Systems', recommendation_name=recommendation_name)
+        # view.search_field.wait_displayed()
+        wait_for(lambda: view.table.row(), handle_exception=True, timeout=20)
+        time.sleep(5)
+        view.bulk_select.select_all()
+        view.remediate.click()
+        self.browser.plugin.ensure_page_safe(timeout='30s')
+        modal = RemediateSummary(self.browser)
+        if modal.is_displayed:
```
LadislavVasina1, line 114: Here I am thinking about the case where this if clause might be False (some timing issue for example with opening the modal), that's why I would maybe rather use `wait_for` here.
```suggestion
    wait_for(lambda: modal.is_displayed, handle_exception=True, timeout=20)
```
What do you think @ColeHiggins2 ?

airgun/entities/cloud_insights.py
```diff
@@ -49,6 +60,93 @@ def update(self, values):
...
+
+@navigator.register(RecommendationsTabEntity, 'Affected Systems')
+class NavigateToAffectedSystems(NavigateStep):
+    """Navigate from Recommendations tab to the Affected Systems details view."""
+
+    VIEW = RecommendationsDetailsView
+
+    def prerequisite(self, *args, **kwargs):
+        # Ensure we are on the Recommendations tab first
+        return self.navigate_to(self.obj, 'All')
```
LadislavVasina1, line 136:
```suggestion
        return self.navigate_to(self.obj, 'Recommendations')
```
Change linked to [this review comment](https://github.com/SatelliteQE/airgun/pull/1983/files#r2374593279).

airgun/views/cloud_insights.py
```diff
@@ -71,3 +76,128 @@ class CloudInsightsView(BaseLoggedInView, SearchableViewMixinPF4):
...
+        """Check that the row is displayed."""
+        return self.affected_systems_url.is_displayed
+
+
+class RecommendationsTabView(BaseLoggedInView):
+    """View representing the Recommendations Tab."""
+
+    title = PF5Title('Recommendations')
+    search_field = TextInput(locator=(".//input[@aria-label='text input']"))
+    clear_button = PF5Button("Reset filters")
```
LadislavVasina1, line 175:
```suggestion
    clear_button = PF5Button('Reset filters')
```

airgun/views/cloud_insights.py
```diff
@@ -71,3 +76,128 @@ class CloudInsightsView(BaseLoggedInView, SearchableViewMixinPF4):
...
+    incidents = Text(locator=".//a[@data-testid='Incidents']")
```
LadislavVasina1, line 176:
```suggestion
    incidents = Text(locator='.//a[@data-testid="Incidents"]')
```

airgun/views/cloud_insights.py
```diff
@@ -71,3 +76,128 @@ class CloudInsightsView(BaseLoggedInView, SearchableViewMixinPF4):
...
+
+
+class RecommendationsDetailsView(BaseLoggedInView):
+    """Models everything in the recommendations details views execpt the affected system link"""
+
+    title = PF5Title('Affected Systems')
+    clear_button = PF5Button("Reset filters")
+    remediate = PF5Button('Remediate')
+    download_playbook = PF5Button('Download playbook')
+    search_field = TextInput(locator=(".//input[@aria-label='text input']"))
```
LadislavVasina1, line 134:
```suggestion
    search_field = TextInput(locator=('.//input[@aria-label="text input"]'))
```

airgun/views/cloud_insights.py
```diff
@@ -71,3 +76,128 @@ class CloudInsightsView(BaseLoggedInView, SearchableViewMixinPF4):
...
+    bulk_select = BulkSelectMenuToggle()
+    table = PF5Table(
+        locator='.//table[contains(@aria-label, "Host inventory")]',
+        column_widgets={
+            0: Checkbox(locator='.//input[@type="checkbox"]'),
+            "Name": Text(".//a"),
+            "OS": Text(".//span"),
+            "Last seen": Text(".//span"),
+            "First impacted": Text(".//span"),
```
LadislavVasina1, line 143:
```suggestion
            'Name': Text('.//a'),
            'OS': Text('.//span'),
            'Last seen': Text('.//span'),
            'First impacted': Text('.//span'),
```

airgun/views/cloud_insights.py
```diff
@@ -71,3 +76,128 @@ class CloudInsightsView(BaseLoggedInView, SearchableViewMixinPF4):
...
+    @property
+    def is_displayed(self):
+        return self.browser.wait_for_element(self.table, exception=False) is not None
+
+
+class RecommendationsTableExpandedRowView(RecommendationsDetailsView, View):
+    """View that models the recommendation expandable row content"""
+
+    affected_systems_url = Text(
+        locator=".//*[contains(@class, 'ins-c-rule-details__view-affected')]//a"
```
LadislavVasina1, line 161:
```suggestion
        locator='.//*[contains(@class, "ins-c-rule-details__view-affected")]//a'
```

airgun/views/cloud_insights.py
```diff
@@ -71,3 +76,128 @@ class CloudInsightsView(BaseLoggedInView, SearchableViewMixinPF4):
...
+        content_view=RecommendationsTableExpandedRowView,
+        column_widgets={
+            "Name": Text(".//a"),
+            "Modified": Text(".//span"),
+            "Category": Text(".//span"),
+            "Total risk": Text(".//span"),
+            "Systems": Text(".//div"),
+            "Remediation type": Text(".//span"),
+        },
+    )
```
LadislavVasina1, line 191:
```suggestion
    critical_recommendations = Text(locator='.//a[@data-testid="Critical recommendations"]')
    important_recommendations = Text(locator='.//a[@data-testid="Important recommendations"]')
    conditional_filter_dropdown = PF5OUIATextInput('ConditionalFilter')
    table = PF5ExpandableTable(
        locator='.//table[contains(@aria-label, "rule-table")]',
        content_view=RecommendationsTableExpandedRowView,
        column_widgets={
            'Name': Text('.//a'),
            'Modified': Text('.//span'),
            'Category': Text('.//span'),
            'Total risk': Text('.//span'),
            'Systems': Text('.//div'),
            'Remediation type': Text('.//span'),
        },
    )
```

airgun/entities/cloud_insights.py
```diff
@@ -49,6 +60,93 @@ def update(self, values):
...
+class RecommendationsTabEntity(BaseEntity):
+    endpoint_path = '/foreman_rh_cloud/insights_cloud'
+
+    def search(self, value):
+        """Search for 'query' and return matched hostnames/recommendations.
+
+        :param value: text to filter (default: no filter)
+        """
+        view = self.navigate_to(self, 'All')
+        time.sleep(5)
```
sambible, line 72: It would be preferable to use a wait_for construct on something you expect to only be present when the page is loaded instead of the sleeps here, I wouldn't block on this but it will likely save you time in the future.

airgun/views/cloud_insights.py
```diff
@@ -71,3 +76,128 @@ class CloudInsightsView(BaseLoggedInView, SearchableViewMixinPF4):
...
+            'Total risk': Text('.//span'),
+            'Systems': Text('.//div'),
+            'Remediation type': Text('.//span'),
+        },
+    )
+
+    @property
+    def is_empty(self):
+        """Check whether the table is empty."""
+        return self.table.is_displayed
```
sambible, line 196: This doesn't actually check whether the table is empty or not. If we are going to include this, it should do what it says.

airgun/views/cloud_insights.py
```diff
@@ -71,3 +76,128 @@ class CloudInsightsView(BaseLoggedInView, SearchableViewMixinPF4):
...
+            'Name': Text('.//a'),
+            'OS': Text('.//span'),
+            'Last seen': Text('.//span'),
+            'First impacted': Text('.//span'),
+        },
+    )
+
+    @property
+    def is_empty(self):
+        """Check whether the table is empty."""
```
sambible, line 149: Same as below.

sambible (changes requested):

airgun/entities/cloud_insights.py
ColeHiggins2: I actually tried used a wait_for construct here but the elements inconsistency on the page made time.sleep() the only solution that worked consistently

airgun/entities/cloud_insights.py
```diff
@@ -69,3 +167,14 @@ class ShowCloudInsightsView(NavigateStep):
     @retry_navigation
     def step(self, *args, **kwargs):
         self.view.menu.select('Red Hat Lightspeed', 'Recommendations')
+
+
+@navigator.register(RecommendationsTabEntity, 'All')
```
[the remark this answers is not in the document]
ColeHiggins2: im confused what you mean here... ALL does not navigate to the all hosts page?. It navigates to the recommendations page and returns the table view... ALL in this case is just referring to the main recommendations page...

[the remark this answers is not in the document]
sambible: I'd recommend 'All Recommendations' just for clarity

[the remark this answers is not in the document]
LadislavVasina1: @sambible The second parameter in the `@navigator.register()` here is just a name by which you can refer to this particular navigation, (It can be whatever, that's why I suggested to use `Recommendations` so it properly describes the executed navigation step) and in this case the the `ShowRecommendationsView` goes to the `self.view.menu.select('Red Hat Lightspeed', 'Recommendations')`. It does not go through the All hosts page.
@sambible correct me if I am wrong.

airgun/entities/cloud_insights.py
jeremylenz: Is that because this is trying to work with both IoP and non-IoP recommendations page? I would recommend against that if possible, since they're totally different

airgun/entities/cloud_insights.py
```diff
@@ -49,6 +60,93 @@ def update(self, values):
...
+        return view.read()
+
+    def bulk_remediate_affected_systems(self, recommendation_name):
+        """Open Affected systems, bulk select affected systems, and click Remediate.
+
+        Returns the details view contents after remediation click.
+        """
+        # Use navigator to open the Affected Systems details view
+        view = self.navigate_to(self, 'Affected Systems', recommendation_name=recommendation_name)
+        # view.search_field.wait_displayed()
```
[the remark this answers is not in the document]
LadislavVasina1: @ColeHiggins2 Please remove this comment

airgun/entities/cloud_insights.py
```diff
@@ -49,6 +60,93 @@ def update(self, values):
...
+        wait_for(lambda: view.table.row(), handle_exception=True, timeout=20)
+        time.sleep(5)
```
LadislavVasina1, line 109: Can't we conceal these waits in the `Affected Systems` navigator `step` (line 138) function, make it more rigid there, so we don't have to deal with it in the entities where this navigator is used.

LadislavVasina1:
@ColeHiggins2 Also docs are failing, look into that too please.

airgun/views/cloud_insights.py
```diff
@@ -71,3 +76,118 @@ class CloudInsightsView(BaseLoggedInView, SearchableViewMixinPF4):
...
+    """
+    A menu toggle component that combines a checkbox with a dropdown menu.
+    Used for bulk selection operations with additional menu options.
+    Usage example (view.bulk_select = BulkSelectMenuToggle()):
+        view.bulk_select.select_all() -> to select all items using the checkbox
+        view.bulk_select.deselect_all() -> to deselect all items using the checkbox
+        view.bulk_select.is_all_selected -> to check if all items are selected
+        view.bulk_select.items -> to access the menu items
+            (['Select none', 'Select page (1 items)', 'Select all (1 items)'])
+        view.bulk_select.item_select('Select none') -> to select a specific item by name
```
LadislavVasina1, line 91: @ColeHiggins2 I think that this is the indent that is not liked by docs.

PR #2000, opened 2025-09-03
Added refresh as job status for perticular hosts is not getting updated by itself

jnagare-redhat:
### Problem statement 
Job status for specific host is not getting updated once job got completed successfully.

### Solution 
We need to refresh page before reading job status of host otherwise it return existing status i.e **Pending**

Note: This is dependent on https://github.com/SatelliteQE/robottelo/pull/19348


LadislavVasina1:
Yay 🥳 Airgun's 2000th PR

LadislavVasina1 (approved):

PR #2164, opened 2025-10-20
PF5 button updates

pondrejk:
- changes to airgun to adjust to https://github.com/theforeman/foreman/pull/10716 
- experimental: fixes proposed by claude, testset for verification as well


sambible (approved):
The airgun changes here look sound. 

The failiures here are all the same test, with the same signature.. If that failure is expected/being addressed in a separate PR, then I'm good.  ACK pending that.

LadislavVasina1 (approved):
ACK

PR #2181, opened 2025-10-24
Add RH Flatpak info alert to the Create modal

vsedmik:
"Add Red Hat Flatpak remote" has been added to the Create Flatpak Remote modal to ease the creation.
image

This PR adds it to the view and entities.

Robottelo PR will follow.

### DIFF
--- a/airgun/entities/flatpak.py
+++ b/airgun/entities/flatpak.py
@@ -83,6 +83,52 @@ class FlatpakRemotesEntity(BaseEntity):
         view = FlatpakRemoteDetailsView(self.browser)
         view.wait_displayed(delay=3)
 
+    def create_redhat_remote(self, values):
+        """Create a Red Hat Flatpak remote using the info alert action.
+
+        This method clicks the 'Add Red Hat flatpak remote' button from the
+        info alert instead of manually filling the URL.
+
+        :param dict values: values to create the remote with
+        """
+        view = self.navigate_to(self, 'All')
+        view.wait_displayed()
+        view.create_new_btn.click()
+        create_modal = CreateFlatpakRemoteModal(self.browser)
+        create_modal.wait_displayed()
+        create_modal.fill(values)
+
+        if create_modal.info_alert.is_displayed:
+            create_modal.add_rh_fr.click()
+        else:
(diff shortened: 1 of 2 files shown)


airgun/entities/flatpak.py
```diff
@@ -83,6 +83,52 @@ def create(self, values):
...
+        This method clicks the 'Add Red Hat flatpak remote' button from the
+        info alert instead of manually filling the URL.
+
+        :param dict values: values to create the remote with
+        """
+        view = self.navigate_to(self, 'All')
+        view.wait_displayed()
+        view.create_new_btn.click()
+        create_modal = CreateFlatpakRemoteModal(self.browser)
+        create_modal.wait_displayed()
```
sambible: I would suggest moving all this logic to a 'Create' Navigator, and replacing existing usage of this construct with that instead. Should make it more readable, and reusable.
```suggestion
    view = self.navigate_to(self, 'Create')
```

airgun/entities/flatpak.py
```diff
@@ -83,6 +83,52 @@ def create(self, values):
...
+        create_modal.fill(values)
+
+        if create_modal.info_alert.is_displayed:
+            create_modal.add_rh_fr.click()
+        else:
+            raise ValueError('Red Hat Flatpak remote info alert is not displayed')
+
+        create_modal.create_btn.click()
+        view = FlatpakRemoteDetailsView(self.browser)
```
sambible: Same here, I'd suggest a 'Details' Navigator, which could replace both the flat navigation to the `FlatpakRemoteDetailsView` and the version of it that needs to look at a specific entity, like in the above `create` method. If you want inspiration, you can look at this method in host_new: https://github.com/SatelliteQE/airgun/blob/f2a5e47274ed1c66ee7bd6148ae07e85adc86667/airgun/entities/host_new.py#L1117

sambible (approved):
Additions are fine, but I've given some suggestions to improve the readibility/usability/maintainability of the file overall. 

I'm fine with merging, but I'd like to see at least a github issue for the suggestions created, maybe a JIRA for when we have some time to impelement :)

vsedmik:
rebased

PR #2214, opened 2025-11-21
New All Hosts Features - Manage SystemPurpose, Manage Traces

LadislavVasina1:
Provide coverage for new 6.19 features that are part of EPIC `[Part 4] - New All Hosts page - Complete the top actions and eliminate some AngularJS pages`
  - https://issues.redhat.com/browse/SAT-35464 (`As a user I want to be able to change the system purpose of multiple hosts`)
- https://issues.redhat.com/browse/SAT-35465 (`As a user I want to be able to manage host tracer of multiple hosts`)

This PR is needed by https://github.com/SatelliteQE/robottelo/pull/20299, PRT PASSED there.

### DIFF
--- a/airgun/entities/all_hosts.py
+++ b/airgun/entities/all_hosts.py
@@ -24,6 +24,8 @@ from airgun.views.all_hosts import (
     ManageErrataModal,
     ManagePackagesModal,
     ManageRepositorySetsModal,
+    ManageSystemPurposeModal,
+    ManageTracesModal,
 )
 from airgun.views.job_invocation import JobInvocationCreateView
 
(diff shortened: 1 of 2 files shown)


airgun/entities/all_hosts.py
```diff
@@ -750,6 +752,118 @@ def read_host_status_icon(self, host_name):
...
+        role=None,
+        usage=None,
+        service_level=None,
+        release_version=None,
+    ):
+        """
+        This function allows changing system purpose attributes for selected hosts
+        through the All Hosts page.
+
+        :param host_names: str with one host or list of hosts to select
```
ColeHiggins2: Im not requesting a change here but do we usually define each param in our doc strings? looks like AI to me...

PR #2238, opened 2025-12-11
fix update function for repository airgun entities

vijaysawant:
Stream and 6.18.2 fix 
Robottelo PR https://github.com/SatelliteQE/robottelo/pull/20470
### Problem
- The test test_positive_end_to_end_custom_yum_crud failed when updating a repository. The flash message widget tried to read a second toast notification that didn't exist, causing:
This happened because:
- EditableEntry widgets auto-save individually when filled
- Multiple flash messages can appear
- The widget attempted to read all messages, including a non-existent second one

### Solution
Updated the update method in airgun/airgun/entities/repository.py to:
- Dismiss existing flash messages before filling (using contextlib.suppress)
- Wait for the UI to stabilize after auto-saves
- Handle flash message reading errors gracefully by catching specific exceptions (NoSuchElementException, AttributeError) instead of bare Exception 
- Used contextlib.suppress() instead of try-except-pass
- Caught specific exceptions instead of bare Exception
- Added logging when flash message reading fails
- Test continues even if flash messages can't be read

### DIFF
--- a/airgun/entities/repository.py
+++ b/airgun/entities/repository.py
@@ -1,5 +1,7 @@
+from contextlib import suppress
 import re
 
+from selenium.common.exceptions import NoSuchElementException
 from wait_for import wait_for
 
 from airgun.entities.base import BaseEntity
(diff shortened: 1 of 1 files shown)


airgun/entities/repository.py
```diff
@@ -61,9 +63,34 @@ def update(self, product_name, entity_name, values):
         if values.get('repo_content.http_proxy_policy') == 'Global Default':
             values['repo_content.http_proxy_policy'] = self.global_default_http_proxy
         view = self.navigate_to(self, 'Edit', product_name=product_name, entity_name=entity_name)
+        # Dismiss any existing flash messages before filling to avoid confusion
+        with suppress(NoSuchElementException, AttributeError):
```
sambible: Is there actually any case where this is occuring? I'd prefer to not over error-proof this, it increases test run time and lowers code readibility.

airgun/entities/repository.py
```diff
@@ -61,9 +63,34 @@ def update(self, product_name, entity_name, values):
...
+        # Wait for all EditableEntry auto-saves to complete and flash messages to stabilize
+        # EditableEntry widgets auto-save individually, so we need to wait for UI to settle
+        # before checking flash messages to avoid reading non-existent toast messages
+        wait_for(
+            lambda: view.browser.plugin.ensure_page_safe(timeout='1s') is None,
+            handle_exception=True,
+            timeout=30,
+            delay=1,
+            logger=view.flash.logger,
+        )
```
sambible:
```suggestion
self.browser.plugin.ensure_page_safe(timeout='30s')
```
Can simply do this, should make it quite a bit more readable.

PR #2248, opened 2025-12-17
Synced container pullable path

sambible:
Adds support for the Pullable Paths table to both the Synced Containers table via a new modal, and the Manifest Details page via a dropdown. In both cases, the table is the same. 

Supports: https://github.com/SatelliteQE/robottelo/pull/20542


LadislavVasina1 (approved):
ACK, looks good from the Airgun side

vijaysawant (approved):

PR #2263, opened 2026-01-14
Improve get_register_command entity

LadislavVasina1:
In the current implementation of the entity, it was not properly handling the state of the form when there were more or none activation keys on SAT.


LadislavVasina1:
```
trigger: test-robottelo
pytest: tests/foreman/ui/test_registration.py 
```

PR #2297, opened 2026-02-03
Draft initial multi-CV view and entity components

synkd:
This PR adds the necessary bits to support the ability to select multiple content view environments, which will be used in several places in the UI.

The view classes added here will likely need to end up in `airgun/views/common.py`, but I am keeping them in `host_new.py` for debugging purposes until I get them functioning properly.


LadislavVasina1 (approved):
ACK, Good addition @synkd!

PR #2299, opened 2026-02-04
Add functionality for bulk vulnerability analysis Disable/Enable

ColeHiggins2:
Views and entities for Bulk Vulnerability Analysis Disable/Enable


LadislavVasina1 (approved):
ACK, let's wait for PRT to pass in the robottelo

PR #2300, opened 2026-02-05
Updated views and added helper method to support multi-cv activation key

jnagare-redhat:
Updated views and added helper method to support multi-cv activation key
New Modal was added in recent snap to select LCE and CV as a part of Multi-CV functionality
Screenshot From 2026-02-05 15-37-53


ColeHiggins2 (approved):
Nice work! I have no problems with this

synkd (approved):
One non-blocking request, otherwise this looks good.

PR #2308, opened 2026-02-10
Updated multiCV modal locator to support bulk update as well

jnagare-redhat:
MultiCV bulk update `Manage content->Content view environments` modal/View has some different locator. To reuse same Modal/View that we used for activation key and host overview I have updated locators.
Robotello PR:https://github.com/SatelliteQE/robottelo/pull/20755

### DIFF
--- a/airgun/entities/all_hosts.py
+++ b/airgun/entities/all_hosts.py
@@ -110,8 +110,8 @@ class AllHostsEntity(BaseEntity):
     def manage_cve(self, lce_name=None, cv_name=None):
         """Bulk reassign Content View Environments through the All Hosts page
         args:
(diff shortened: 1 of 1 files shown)


airgun/entities/all_hosts.py
vijaysawant: please update the doc string as well according to updated parameter list.
```suggestion
    def manage_cve(self, lce_name=None, cv_name=None):
"""Bulk reassign Content View Environments through the All Hosts page
       args:
           lce_name (str): Lifecycle Environment to swap the hosts to.
           cv_name (str): CV within that LCE to assign the hosts to.
       """
```

PR #2313, opened 2026-02-11
Fix some search bugs with CV UI

sambible:
Just fixes some small problems that have crept into CV UI over the last year or so.

Removing the 
```search = PF4Search()``` 
is fine, since it uses SearchableViewMixin, like most other pages.

### CODE REVIEW

**Comment by sambible** [comment]
```
trigger: test-robottelo
pytest: tests/foreman/ui/test_contentview.py
```

PR #2327, opened 2026-02-26
Add locator support for OpenShift Virtualization

amolpati30:
This change introduces locator support for OpenShift Virtualization. With this enhancement, during host creation, the automation framework will dynamically resolve the appropriate UI elements and navigation paths specific to OpenShift Virtualization, ensuring the correct workflow is executed.


Gauravtalreja1 (commented):
Great job @amolpati30 , added a few comments and could you also add locators for ComputeResource page so this PR covers everything related to OCP-V?

sambible (approved):
Looks good, but I'd like to see a test utilizing the ComputeResource page locators. If that's out of scope for the robottelo PR, can we create an issue to make sure we come back and do it?

PR #2332, opened 2026-03-09
Update entities for new content credentials UI list page

sjha4:
Testing out claude agent team work for new entities in Airgun for the new UI here: https://github.com/Katello/katello/pull/11671


sambible (changes requested):
We shouldn't move this into a "_new.py" file, since it's very similar to the old page, and the old one will be deprecated at the same time that the new one comes in.

sambible (changes requested):

sjha4:
This is not needed anymore..All entities got added in Vlad's PR. Closing this out.

PR #2494, opened 2026-07-29
Fix Create Host button locator for new UI empty state

nacoool:
The new Hosts page empty state uses OUIA ID
  'empty-state-secondary-action-button' instead of 'create-host-button'.
  Update the XPath to match either ID, with text filtering to distinguish
  Create Host from Register Host which shares the same OUIA ID.

### DIFF
--- a/airgun/views/host.py
+++ b/airgun/views/host.py
@@ -228,7 +228,13 @@ class HostsView(BaseLoggedInView, SearchableViewMixinPF4):
     manage_columns = PF5Button('manage-columns-button')
     searchbar_dropdown = PF5OUIADropdown('selection-checkbox')
     export = Text(".//a[contains(@class, 'btn')][contains(@href, 'hosts.csv')]")
-    new = Text(".//div[@id='foreman-page']//a[@data-ouia-component-id='create-host-button']")
+    new = Text(
+        ".//div[@id='foreman-page']//a["
+        "@data-ouia-component-id='create-host-button' or "
+        "(@data-ouia-component-id='empty-state-secondary-action-button' "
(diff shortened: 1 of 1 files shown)


LadislavVasina1:
This feels like a bug to me, I dont really like this particular ID rename, could you check that please @nacoool

airgun/views/host.py
```diff
@@ -228,7 +228,13 @@ class HostsView(BaseLoggedInView, SearchableViewMixinPF4):
...
     searchbar_dropdown = PF5OUIADropdown('selection-checkbox')
     export = Text(".//a[contains(@class, 'btn')][contains(@href, 'hosts.csv')]")
-    new = Text(".//div[@id='foreman-page']//a[@data-ouia-component-id='create-host-button']")
+    new = Text(
+        ".//div[@id='foreman-page']//a["
+        "@data-ouia-component-id='create-host-button' or "
+        "(@data-ouia-component-id='empty-state-secondary-action-button' "
+        "and normalize-space(.)='Create Host')"
+        "]"
+    )
```
stejskalleos, line 237: Why not just 
```suggestion
    new = Text(
        ".//a["
        "@data-ouia-component-id='create-host-button' or "
        "@data-ouia-component-id='empty-state-secondary-action-button' ""
        "]"
    )
```

and can we use better naming than `new`? Specially when there is a `new_ui_button` down there :D

stejskalleos (changes requested):

nacoool:
The PR was created since on local deployment; i was seeing the UI was clicking on old UI button first and then trying to find Create Button. Since now I it is not. Its safe to close the PR.

PR #2512, opened 2026-08-20
Make sure OUIA components are imported with OUIA prefix

LadislavVasina1:
## Problem
In some cases in Airgun, the custom of importing PF5 OUIA components with `PF5OUIA` was not honored and might cause confusion and inconsistencies.

## Solution
This PR fixes that, and now all the imported PF5 OUIA widgets have `PF5OUIA` prefix.
This change is just renaming so no test impact is expected.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -171,7 +171,10 @@ from datetime import datetime
 
 # Third-party
 from widgetastic.widget import Text, TextInput, View
-from widgetastic_patternfly5 import PF5Button, PF5Table
+from widgetastic_patternfly5 import (
+    Button as PF5Button
+    Table as PF5Table
+    )
 
 # Airgun
 from airgun.entities.base import BaseEntity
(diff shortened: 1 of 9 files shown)


vsedmik (approved):
LGTM

AGENTS.md
```diff
@@ -171,7 +171,10 @@ from datetime import datetime
 
 # Third-party
 from widgetastic.widget import Text, TextInput, View
-from widgetastic_patternfly5 import Button, Table
+from widgetastic_patternfly5 import (
+    Button as PF5Button
```
synkd, line 175: Non-blocking, but why not import these as `PF5OUIA` as well?

synkd (approved):
One non-blocking question, but looks good overall.

AGENTS.md
LadislavVasina1: Because they are just widgetastic_patternfly5 not widgetastic_patternfly5.ouia

PR #2522, opened 2026-08-31
Fix host content view environment widget for read-only fields

jnagare-redhat:
### Problem Statement
The host content view environment field in the UI has changed behavior:
- When creating/editing hosts: appears as a `<select>` dropdown (Select2)
- When viewing inherited values from hostgroup: appears as a disabled `<input type="text">`

The existing `FilteredDropdown` widget only handled the select dropdown case, causing `NoSuchElementException` when trying to read inherited content view environment values.

### Solution
Added a separate `read_content_view_environment` widget using `TextInput` to properly read the disabled input field that displays inherited values. This allows the same view to handle both cases:
- `content_view_environment`: for filling/selecting values (dropdown)
- `read_content_view_environment`: for reading inherited values (text input)

### Review Feedback Addressed
Removed hacky string matching for content source selection. Instead, callers should pass the actual smart proxy name obtained from API queries (see robottelo PR for helper method).

### Changes
- `airgun/views/host.py`: Added `read_content_view_environment` TextInput widget
- `airgun/entities/host.py`: Removed string prefix matching for content source

### Testing
- Verified widget can read inherited CV environment values from hostgroup
- Tested content source selection with actual smart proxy names
- Test: `test_manage_content_source_with_multi_cv` PASSED (9m 31s)

### Related
- Robottelo PR: https://github.com/SatelliteQE/robottelo/pull/22590

### DIFF
--- a/airgun/entities/host.py
+++ b/airgun/entities/host.py
@@ -256,14 +256,7 @@ class HostEntity(BaseEntity):
         # Select content source
         wait_for(lambda: view.content_source_select.is_displayed, timeout=10, delay=1)
         sleep(2)
-        # The rendered option may carry a Pulp content-proxy suffix (e.g. "<fqdn>-pulp").
-        # Match by prefix so callers can pass the plain Satellite/Capsule hostname.
-        available = view.content_source_select.items
-        match = next(
-            (i for i in available if i == content_source or i.startswith(f'{content_source}-')),
-            content_source,
-        )
-        view.content_source_select.fill(match)
+        view.content_source_select.fill(content_source)
 
         # Assign multiple CVEnvs
         for i, assignment in enumerate(cv_env_assignments):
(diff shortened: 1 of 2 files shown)


airgun/entities/host.py
```diff
@@ -256,7 +256,14 @@ def change_content_source(
         # Select content source
         wait_for(lambda: view.content_source_select.is_displayed, timeout=10, delay=1)
         sleep(2)
-        view.content_source_select.fill(content_source)
+        # The rendered option may carry a Pulp content-proxy suffix (e.g. "<fqdn>-pulp").
+        # Match by prefix so callers can pass the plain Satellite/Capsule hostname.
+        available = view.content_source_select.items
+        match = next(
+            (i for i in available if i == content_source or i.startswith(f'{content_source}-')),
```
ColeHiggins2: Same comment as the robottelo one. Can we remove the startswith method here?

ColeHiggins2 (changes requested):

airgun/entities/host.py
jnagare-redhat: Updated PR

LadislavVasina1 (approved):

ColeHiggins2 (approved):

PR #2529, opened 2026-09-10
Fix subscription search wait for PatternFly 5

jnagare-redhat:
### Problem Statement
With the PatternFly 5 UI migration, subscription search was failing because the wait logic returned too early. The PF5 UI shows an immediate "No Results" empty state while the search is still processing, which caused Airgun to read the empty table before actual results loaded.

This manifested as test failures where `session.subscription.search()` returned an empty list even though subscriptions existed and were visible in the UI.

### Root Cause
The `search_settled` property was checking for any `<tr>` element in tbody, but PF5 renders temporary rows with `data-ouia-component-id="table-empty"` during loading. These empty state rows appeared immediately, causing `search_settled` to return `True` before actual data rows loaded.

### Solution
Override `tbody_row` to only match actual data rows with component IDs starting with `subscriptions-table-row-`, excluding empty state rows. This ensures the wait logic only completes when real subscription data is present.

**Key changes:**
1. Added `tbody_row` override using XPath to match only data rows: `.//tbody/tr[starts-with(@data-ouia-component-id, "subscriptions-table-row-")]`
2. Updated `NO_RESULTS_MESSAGE` to match PF5 format: `'No Results No subscriptions match your search criteria.'`
3. Updated table locator to use PF5 `data-ouia-component-id` attribute
4. Increased search delay from 1s to 2s for more reliable waiting

### Testing
Verified with test `test_positive_access_with_non_admin_user_with_manifest` which was failing before this fix and now passes consistently.

### Related Issues
Fixes subscription search reliability on PatternFly 5 Satellite UI.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/airgun/views/subscription.py
+++ b/airgun/views/subscription.py
@@ -43,7 +43,9 @@ class SatSubscriptionsViewTable(SatTable):
 
     NO_RESULTS_MESSAGE = 'No Results No subscriptions match your search criteria.'
     # Override tbody_row to only match actual data rows, not empty state rows
-    tbody_row = Text('.//tbody/tr[starts-with(@data-ouia-component-id, "subscriptions-table-row-")]')
+    tbody_row = Text(
+        './/tbody/tr[starts-with(@data-ouia-component-id, "subscriptions-table-row-")]'
+    )
 
     @property
     def is_displayed(self):
--- a/airgun/views/subscription.py
+++ b/airgun/views/subscription.py
@@ -41,7 +41,9 @@ class SatSubscriptionsViewTable(SatTable):
     component instead of a table, so we need to check if the table exists first.
     """
 
-    NO_RESULTS_MESSAGE = 'No subscriptions match your search criteria.'
+    NO_RESULTS_MESSAGE = 'No Results No subscriptions match your search criteria.'
+    # Override tbody_row to only match actual data rows, not empty state rows
+    tbody_row = Text('.//tbody/tr[starts-with(@data-ouia-component-id, "subscriptions-table-row-")]')
 
     @property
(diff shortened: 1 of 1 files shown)


jnagare-redhat:
trigger: test-robottelo
pytest: tests/foreman/ui/test_subscription.py -k test_positive_access_with_non_admin_user_with_manifest
deployment_type: container

airgun/views/subscription.py
```diff
@@ -150,10 +154,11 @@ def search(self, query):
             return None
         self.searchbox.search(query)
         self.browser.plugin.ensure_page_safe(timeout=60)
+        # Wait longer for search to complete
         wait_for(
             lambda: self.table.search_settled,
             timeout=30,
-            delay=1,
+            delay=2,
```
chris1984, line 161: Should we use 3 or is 2 enough?

chris1984 (approved):
Looks fine, just had one question about the delay. I have seen this page sometimes take a moment to load everything.

Satellite-QE:
**PRT Result**
```
Build Number: 926
Build Status: SUCCESS
PRT Comment: pytest tests/foreman/ui/test_subscription.py -k test_positive_access_with_non_admin_user_with_manifest --external-logging
Test Result : =========== 1 passed, 6 deselected, 5 warnings in 805.44s (0:13:25) ============
```

synkd (approved):

airgun/views/subscription.py
jnagare-redhat: 2 seems enough to me

LadislavVasina1 (approved):

